PR #10, opened 2021-10-26
Add support for variant overrides

mportiz08:
Bug/issue #, if applicable: 82968068

## Summary

Adds renderer support for the newly introduced [`variantOverrides`][1] key in DocC Render JSON, which allows for using the same JSON file to represent documentation data for a symbol which is available in more than one variant—most commonly for a symbol available in both Swift and Objective-C languages.

For symbols available in both Swift/Objective-C, there will now be an affordance in the sidebar that can be used to toggle between the current language.

Example:

demo

### Details

As described in https://github.com/apple/swift-docc/pull/11, the `variantOverrides` data may contain overrides for a given variant expressed as a [JSON Patch][2]—this patch can be applied to the Render JSON to obtain the language specific version of any data that may be rendered on the page. When this overrides data is available for a Swift symbol that is also available in Objective-C, the language toggle will be presented and used to switch between the different page data.

[1]: https://github.com/apple/swift-docc/pull/11
[2]: https://datatracker.ietf.org/doc/html/rfc6902

## Dependencies

https://github.com/apple/swift-docc/pull/11

## Testing

Using fake example fixture data, verify that the language toggle is presented and can be used.

Steps:
1. Download/unzip [example.zip](https://github.com/apple/swift-docc-render/files/7420136/example.zip)
2. Configure `VUE_APP_DEV_SERVER_PROXY=/path/to/example` and run `npm run serve`.
3. Load the example page at http://localhost:8080/documentation/mykit/myclass and note that there is now a language toggle in the sidebar.
4. Note that "Swift" is not a link and click the "Objective-C" link.
5. Verify that the following changes have been made to the page:
  * "Objective-C" is no longer a link and "Swift" is
  * the page title changes from `init(fileURLWithPath:)` to `initFileURLWithPath:`
  * the declaration changes from `init(fileURLWithPath path: String)` to `- (instancetype)initFileURLWithPath:(NSString *)path;`
  * the last link in the first "Topics" group changed from "myFunction in Swift" to "myFunction in Objective-C"
  * the abstract for the same link also changes in a similar way
 6. Verify that you can toggle back to Swift and see the original data.
 7. Toggle to Objective-C and close the web browser—then open the same browser and load http://localhost:8080/documentation/mykit/myclass again—verify that it switches to Objective-C (not expected to work in private window/tab or across different browsers)

Also verify that an existing documentation bundle renders in the same way that it normally does—the language toggle will only be used going forward when `variantOverrides` data is provided and these changes should otherwise be backwards compatible with older Render JSON.

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -98,6 +98,7 @@ import { getSetting } from 'docc-render/utils/theme-settings';
 import Aside from 'docc-render/components/ContentNode/Aside.vue';
 import DocumentationNav from 'theme/components/DocumentationTopic/DocumentationNav.vue';
 import BetaLegalText from 'theme/components/DocumentationTopic/BetaLegalText.vue';
+import LanguageSwitcher from 'theme/components/DocumentationTopic/Summary/LanguageSwitcher.vue';
 import Abstract from './DocumentationTopic/Description/Abstract.vue';
 import ContentNode from './DocumentationTopic/ContentNode.vue';
 import CallToActionButton from './CallToActionButton.vue';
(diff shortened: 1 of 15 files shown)


mportiz08:
@swift-ci test

PR #20, opened 2021-11-09
Update URL in declaration to support external links

hqhhuang:
Update URL in declaration to support external links

Bug/issue #, if applicable: rdar://82575979

## Summary
In the Declaration Section of the documentation, we used to only be able to render internal, relative URLs and neglected the case where the Reference could be external. 
Before the fix, `swift-docc-render` would append the external URL as if it's a relative URL. 
With this PR, I updated the implementation in `/DeclarationToken/TypeIdentifierLink.vue` so this edge case is now supported. I updated the unit test with new expected behavior as well.

## Testing
1. Download the fixture:
[fixture.zip](https://github.com/apple/swift-docc-render/files/7508547/fixture.zip)
2. Run `VUE_APP_DEV_SERVER_PROXY=/path/to/fixture/ npm run serve`
3. Open Safari and go to: http://localhost:8080/documentation/test
4. Check that in the declaration section, there should be a link for "bool".
5. Clicking on that links should take you to https://www.swift.org/documentation/docc/, instead of http://localhost:8080/documentation/https:/www.swift.org/documentation/docc/

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationToken/TypeIdentifierLink.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationToken/TypeIdentifierLink.vue
@@ -10,7 +10,7 @@
 
 <script>
 // This component renders token text as a link to a given type.
-import { buildUrl } from 'docc-render/utils/url-helper';
+import Reference from 'docc-render/components/ContentNode/Reference.vue';
 
 export default {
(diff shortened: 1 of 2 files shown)


dobromir-hristov (approved):

marinaaisa (approved):

mportiz08:
@swift-ci test

PR #25, opened 2021-11-24
Stop inserting word-breaks inside type names

karwa:
Bug/issue #, if applicable: [SR-15519](https://bugs.swift.org/browse/SR-15519) (doesn't resolve, but improves)

## Summary

Currently, type names in declarations are subject to line-breaking, resulting in suboptimal rendering such as:

```
func frob<Foo>(..., ..., ..., ..., ..., ...) -> where Foo : Bidirectional
  Collection
```

Or

```
func frob<Foo>(..., ..., ..., ..., ..., ...) -> where Foo : String
  Protocol
```

This PR removes word-breaking from type names, meaning the above now render as:

```
func frob<Foo>(..., ..., ..., ..., ..., ...) -> where Foo : 
  BidirectionalCollection

func frob<Foo>(..., ..., ..., ..., ..., ...) -> where Foo : 
  StringProtocol
```

This is still not optimal, and we should be smarter and use non-breaking spaces in certain elements (e.g. generic constraints) to discourage them from being split across multiple lines, but it's an improvement.

## Checklist

- [X] Updated tests
- [ ] Updated documentation if necessary (no docs to update AFAICT)

I've tested it locally, and it has the desired effect.

Before:

After:

image

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationToken.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationToken.vue
@@ -9,7 +9,6 @@
 -->
 
 <script>
-import WordBreak from 'docc-render/components/WordBreak.vue';
 import ChangedToken from './DeclarationToken/ChangedToken.vue';
 import RawText from './DeclarationToken/RawText.vue';
 import SyntaxToken from './DeclarationToken/SyntaxToken.vue';
@@ -46,9 +45,7 @@ export default {
     }
     case TokenKind.typeIdentifier: {
       const props = { identifier: this.identifier };
-      return createElement(TypeIdentifierLink, { props }, [
-        createElement(WordBreak, text),
-      ]);
+      return createElement(TypeIdentifierLink, { props }, text);
     }
     case TokenKind.added:
     case TokenKind.removed:
--- a/tests/unit/components/DocumentationTopic/PrimaryContent/DeclarationToken.spec.js
+++ b/tests/unit/components/DocumentationTopic/PrimaryContent/DeclarationToken.spec.js
@@ -47,7 +47,7 @@ describe('DeclarationToken', () => {
     const link = wrapper.find(TypeIdentifierLink);
     expect(link.exists()).toBe(true);
     expect(link.props()).toEqual({ identifier: propsData.identifier });
-    expect(link.contains(WordBreak)).toBe(true);
+    expect(link.contains(WordBreak)).toBe(false);
     expect(link.text()).toBe(propsData.text);
   });
 


karwa:
Oh, I just encountered another issue related to this: those word-breaks are part of the text when you copy-and-paste the declaration name.

Firstly, that mans if you copy the name from the documentation and paste it in an IDE's search box, you likely won't find it.

Secondly, it means that if you paste the name in to code, you will have invisible unicode spaces. Recently, I was redesigning part of my API, and when I was done, I added some deprecated typealiases for names that no longer exist, to guide people towards the new API. I found those old names by copying and pasting from the documentation I had built previously (which is a reasonable thing to do, IMO), meaning my types secretly had invisible Unicode spaces in them. For example:

`Lazily​Percent​Decoded​UTF8Without​Substitutions`

Try copy-pasting that it in to Xcode's source editor, then delete it character-by-character. You'll see it appears to snag on word boundaries, because you're deleting the inserted invisible word-break character.

mportiz08:
Thanks for opening this PR @karwa! I don't see any issues with the code necessarily, but I do still have some minor concerns after doing some local testing and also thinking about the desired change here.

I made a quick Swift project to test your branch out, and it looks like the word breaks are still being inserted. The reason for this is most likely that the `TypeIdentifier` component is still injecting the word break elements at some point even after removing the code in this PR (I think it ultimately happens in the `CodeVoice` component). Here is a screenshot of what the example looks like using your branch:
Screen Shot 2021-11-29 at 2 04 53 PM

(Please let me know if I'm doing anything wrong when testing this out since it seems like you had already observed that it works.)

Ignoring that, I am still a little worried about the potential impact of removing the word break hints in this particular place in the UI. This is because it would change the way that we are displaying these declarations, resulting in places where the declaration will now scroll where we would prefer that it wraps. This might be a subjective design approach that we should maybe reconsider in the future, but we are currently wrapping code to the next line for (Swift-only) declarations since we don't yet have a good algorithm for properly breaking up parameters and other syntax nodes in a way that is formatted in a reliably "prettier" way which would look much improved in combination with scrolling when necessary.

> Try copy-pasting that it in to Xcode's source editor, then delete it character-by-character. You'll see it appears to snag on word boundaries, because you're deleting the inserted invisible word-break character.

Was the symbol you have here actually copied from the web browser showing a DocC-Render page? It looks like maybe you've used some invisible unicode characters in place of the `<wbr>` tag that we use, which _would_ have this copy/paste problem, but from my testing, this is not reproducible with the `<wbr>` implementation. I agree that copy/paste should 100% work, but I'm not aware of any issues related to that with the current word break approach used in the renderer.

Sorry for the super long message, but I really appreciate you taking the time to contribute here! Problems with long symbol names and responsive screen sizes give me a major headache 😅 haha

karwa:
Hey, sorry - I'm a little behind with other things at the moment. I'll investigate the issues you raised, but it might be a little while.

mportiz08:
Closing this for now since there haven't been any recent updates but please feel free to re-open at a later time if there are subsequent updates that you've followed up on. Thanks!

PR #26, opened 2021-11-30
Add support to render `References` with human readable name in plain text

hqhhuang:
Bug/issue #, if applicable: 61211397

## Summary
In the `TopicsSection`, we check if the `reference` has a human readable name before rendering. So, if a symbol reference has a human readable name, its title will be rendered in plain text instead of CodeVoice.
We did not consider this scenario anywhere else. This PR adds support to render symbols with human readable names in plain text. I added `ideTitle` and `titleStyle` to the props of the `Reference` component, so it can determine if a symbol has a human readable name and if it should be rendered as text or CodeVoice. 

## Testing
[fixture.zip](https://github.com/apple/swift-docc-render/files/7621933/fixture.zip)

Steps:
1. Run VUE_APP_DEV_SERVER_PROXY=/path/to/manual-fixtures/ npm run serve
2. Go to http://localhost:8080/documentation/example
3. Check that the first example is rendered in plain text, and the following two are rendered in CodeVoice.
The first example, is a `reference` with a human readable name, thus it should be rendered in plain text.
The second example also has a human readable name, but its `titleStyle` says to render as a symbol. So this should be rendered in CodeVoice.
The third is just a sanity check for a regular symbol reference that doesn't have a human readable name. 

Screen Shot 2021-11-29 at 2 57 16 PM

## Checklist

- [x] Added tests
- [x] Updated documentation if necessary - NA

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2117,17 +2117,6 @@
             "color-convert": "^2.0.1"
           }
         },
-        "chalk": {
-          "version": "4.1.2",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-4.1.2.tgz",
-          "integrity": "sha512-oKnbhFyRIXpUuez8iBMmyEa4nbj4IOQyuhc/wy9kY7/WVPcwIO9VA668Pu8RkO7+0G76SLROeyw9CpQ061i4mA==",
-          "dev": true,
-          "optional": true,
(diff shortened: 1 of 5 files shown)


dobromir-hristov (approved):

mportiz08:
@swift-ci test

PR #28, opened 2021-12-03
Change `Frameworks` to `Technologies` to support documenting more product types

hqhhuang:
Bug/issue #, if applicable: 83451964

## Summary
This PR changes the wording of the side bar from `Frameworks` to `Technologies` to improve the presentation of non-framework documentation. This is in support of some work that has already landed on `main` in `Swift-DocC` and has been discussed on the Swift Forums [here](https://forums.swift.org/t/improve-presentation-of-non-framework-documentation-sr-15434/53388).
I also changed variable names that reference `framework(s)` to `technology(ies)`.

| Before | After |
|--------|-------|
|   Screen Shot 2021-12-03 at 11 49 06 AM  |  Screen Shot 2021-12-03 at 11 48 48 AM |

## Testing
Steps:
Download this fixture file: [fixture.zip](https://github.com/apple/swift-docc-render/files/7652263/fixture.zip)
1. Run VUE_APP_DEV_SERVER_PROXY=/path/to/manual-fixtures/ npm run serve
2. Go to http://localhost:8080/documentation/example
3. Verify that the Summary side bar displays `Technologies` instead of `Frameworks`

## Checklist

- [x] Added tests
- [x] Updated documentation if necessary - NA


src/utils/strings.js
```diff
@@ -42,8 +42,14 @@ export function escapeHtml(str) {
 
 // Add the correct grammatical number (singular or plural)
 // to a noun depending on the number of items of an array
-export function pluralize(array) {
-  return array.length === 1 ? '' : 's';
+export function pluralize(str, array) {
```
mportiz08, line 45: What you've changed here is an improvement over what was there before, but I am a little concerned about the fact that we're hard-coding in some specialized logic for pluralization that only works for specific English words.

I have an idea about a way to make this more generic and will be easier to update for a future world when we need to support alternate localizations. Maybe I'll create a quick PR against this one and you can let me know what you think.

mportiz08 (commented):
Looks good, thanks!

I just have one minor concern that is more of a general problem with the existing code with regards to the pluralizing of words, and I might want to make a quick improvement there.

src/utils/strings.js
mportiz08: Maybe it's going a little overboard, but I started working on this today, because I think it would be really useful for the hopefully near future where we can provide localized translations of UI elements.

Here's the diff if you're curious about the direction: https://github.com/hqhhuang/swift-docc-render/compare/update-frameworks-to-technologies...mportiz08:generic-pluralize (haven't finished testing yet)

dobromir-hristov: New pluralisation looks good to me.

dobromir-hristov (commented):
Overall changes look OK to me. Waiting for Marcus to wrap up his pluralisation work.

src/utils/strings.js
mportiz08: https://github.com/hqhhuang/swift-docc-render/pull/1

PR #24, opened 2021-11-18
Add static-hostable support, by allowing changing the baseUrl at runtime

dobromir-hristov:
Bug #: 70509237

## Summary

Allow for the baseURL configuration, to be configurable at build-time, as well as runtime. This allows for the app, to be served from static file hostings that do not support URL rewrites, like Github Pages. 

This PR utilizes already existing configurations from Vue CLI and Webpack.

## Applied changes

1. All json based asset paths are now prefixed with a baseUrl, which defaults to root (`/`).
1. `process.env.BASE_URL` is a standard way to set the base url for Webpack and Vue CLI, so we use it to prefix our url.
1.  Adds a runtime configuration variable `var baseUrl`, allowing docc to inject a base url at runtime.
2. Generates a new `index-template.html` file, for usage inside docc

## Dependencies

https://github.com/apple/swift-docc/pull/44

## Testing with the convert command

Steps:
1. Build the app as normal `npm run build`
2. Run `DOCC_HTML_DIR=/path/to/dist docc convert ./path/to/MyFramework.docc --static-hosting-base-path /myname/myrepo/ --transform-for-static-hosting`
3. Move the output inside a `/myname/myrepo/` folder.
3. The final result can now be served with a server of your choice. If you visit `http://localhost:[port]/myname/myrepo/documentation/my-framework`, it should render your docs.

## Testing with the process-archive command

1. Build the app as normal `npm run build`
2. Run `DOCC_HTML_DIR=/path/to/dist docc process-archive transform-for-static-hosting ./path/to/MyFramework.doccarchive --static-hosting-base-path /myname/myrepo/`
3. Move the output inside a `/myname/myrepo/` folder.
3. The final result can now be served with a server of your choice. If you visit `http://localhost:[port]/myname/myrepo/documentation/my-framework`, it should render your docs.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Add tests
- [x] Add a release note in the radar
- [x] Assign the bug to the primary reviewer

PR #32, opened 2021-12-10
Allow the base URL for hosting to be configured post-build

mportiz08:
* __Rationale:__ Adds support to allow a base URL to be configured after the app has already been built. In this way, the `docc` compiler can generate output that can be hosted at any base URL instead of just the root path at `/`.
* __Risk:__ Medium.
* __Risk Detail:__ There is added complexity to the build process to create a template that DocC can easily manipulate.
* __Reward:__ High
* __Reward Details:__ Allows DocC users to host their app at any given subpath of their website instead of just the root (this is needed for most GitHub Pages hosting scenarios).
* __Original PR:__ #24 
* __Issue:__ rdar://70509237
* __Code Reviewed By:__ @mportiz08 
* __Testing Details:__ Unit tests were added and updated. The expected `docc` compiler workflows utilizing this new feature were manually tested along with normal existing usage.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -17,6 +17,7 @@
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <link rel="mask-icon" href="<%= BASE_URL %>favicon.svg" color="#333333">
     <title><%= process.env.VUE_APP_TITLE %></title>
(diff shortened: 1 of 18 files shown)


franklinsch (approved):

ethan-kusters:
@swift-ci please test

PR #36, opened 2021-12-17
Adjust `scrollBehavior` for smaller nav at small breakpoint

mportiz08:
Bug/issue #, if applicable: 78199719

## Summary

This fixes an issue where the scroll position is off by 4 pixels when jumping to a part of the page using a fragment URL _only_ for devices/windows with small widths.

Since the base nav has a height of 48px instead of 52px only at the small breakpoint, this smaller value should be used as an offset when calculating the y scroll position value.

Here are example before/after screenshots showing the scroll position with main compared to this branch (notice how the background from the previous section bleeds over by a very small amount on main).

**main**
Screen Shot 2021-12-16 at 2 01 01 PM

**this branch**
Screen Shot 2021-12-16 at 2 02 32 PM

## Testing

Steps:
1. Download/unzip [manual-fixtures.zip](https://github.com/apple/swift-docc-render/files/7731114/manual-fixtures.zip)
2. Configure `VUE_APP_DEV_SERVER_PROXY=/path/to/manual-fixtures npm run serve`
3. Enter "Responsive Design Mode" in Safari and choose a small device width.
4. Open the following URL: http://localhost:8080/tutorials/slothcreator/creating-custom-sloths#Create-a-new-project-and-add-SlothCreator
5. Verify that the background from the previous section is no longer visible like in the screenshot example from the main branch.

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/BreakpointEmitter.vue
+++ b/src/components/BreakpointEmitter.vue
@@ -9,44 +9,12 @@
 -->
 
 <script>
-import { BreakpointName } from 'docc-render/utils/breakpoints';
+import {
+  BreakpointAttributes,
+  BreakpointName,
+  BreakpointScopes,
+} from 'docc-render/utils/breakpoints';
 
-const BreakpointScopes = {
-  default: 'default',
-  nav: 'nav',
-};
-
-const BreakpointAttributes = {
-  [BreakpointScopes.default]: {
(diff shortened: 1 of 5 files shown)


mportiz08:
@swift-ci test

dobromir-hristov (approved):

PR #37, opened 2021-12-20
Prevent scrolling the page, when using the custom dropdown component

dobromir-hristov:
Bug/issue #, if applicable:  86641543

## Summary

In latest Safari Technology Preview if you click on the Tutorial pages navigation dropdowns, you get scrolled to the top.

This is a result of us focusing items inside the dropdown, and the page trying to scroll to those items.

## Dependencies

NA

## Testing

Steps:
1. Use any docc archive with a tutorial in it
2. Open the tutorial page and one of the two dropdowns in the Navigation bar
3. Assert that the page does not scroll to the top, upon opening the dropdowns.

## Checklist

- [ ] Added tests - no need

### DIFF
--- a/src/components/DropdownCustom.vue
+++ b/src/components/DropdownCustom.vue
@@ -117,7 +117,7 @@ export default {
     async closeAndFocusToggler() {
       this.closeDropdown();
       await this.$nextTick();
-      this.$refs.dropdownToggle.focus();
+      this.$refs.dropdownToggle.focus({ preventScroll: true });
     },
     closeDropdown() {
       this.isOpen = false;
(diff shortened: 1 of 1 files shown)


mportiz08 (commented):
Seems to work well, although I think this might indicate that the real problem is with the browser since the sticky nav is already in view and shouldn't need to be scrolled to the top of the page per-se.

I'm ok with taking this change as a workaround until the potential browser issue is resolved though.

Is there any reasonable update to the unit tests we can make to test this change do you think?

dobromir-hristov:
> 

Yeah... I think its an issue with the version of Safari, just wanted to find some fix quick :/ I am OK with leaving this out until they fix it too :))

marinaaisa (approved):
It seems to work fine in Safari Technology Preview.
I suggest that we merge this as a way to fix it until the Safari team fixes it.
Thanks, Dobri :)

mportiz08:
@swift-ci test

PR #39, opened 2022-01-03
 Limit full page selection to only code

marinaaisa:
Bug #50901512, if applicable: 

## Summary

Limit full page selection to only code when doing "Cmd + a" on CodeListing component

## Dependencies

NA

## Testing

[manual-fixtures.zip](https://github.com/apple/swift-docc-render/files/7803289/manual-fixtures.zip)

Steps:
1. Run `VUE_APP_DEV_SERVER_PROXY=/path/to/manual-fixtures/ npm run serve`
2. Go to http://localhost:8080/tutorials/slothcreator/creating-custom-sloths
3. Set a mobile view on your browser and scroll until a code listing
4. Click on the filename
5. A modal will open, click on the content
6. Select "cmd + a" on your keyboard
7. Select "cmd + c" on your keyboard
8. Go to any editor or notes app and paste the content
9. Assert that you have only copied and pasted the code instead of the whole page

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode/CodeListing.vue
+++ b/src/components/ContentNode/CodeListing.vue
@@ -10,9 +10,12 @@
 
 <template>
   <div
+    tabindex="-1"
     class="code-listing"
-    :data-syntax="syntaxNameNormalized"
     :class="{ 'single-line': syntaxHighlightedLines.length === 1 }"
+    :data-syntax="syntaxNameNormalized"
+    @keydown.meta.a.prevent="selectCode"
+    @keydown.ctrl.a.prevent="selectCode"
   >
     <Filename
       v-if="fileName"
@@ -21,7 +24,7 @@
       :fileType="fileType"
     >{{ fileName }}
     </Filename>
-    <div class="container-general">
+    <div class="container-general" ref="code">
       <!-- Do not add newlines in <pre>, as they'll appear in the rendered HTML. -->
       <pre><code><span
         v-for="(line, index) in syntaxHighlightedLines"
@@ -97,6 +100,11 @@ export default {
     },
   },
   methods: {
+    selectCode() {
+      window.getSelection().selectAllChildren(
+        this.$refs.code,
+      );
+    },
     isHighlighted(index) {
       return this.highlightedLineNumbers.has(this.lineNumberFor(index));
     },
--- a/src/components/ContentNode/CodeListing.vue
+++ b/src/components/ContentNode/CodeListing.vue
@@ -10,12 +10,9 @@
 
 <template>
   <div
-    tabindex="-1"
     class="code-listing"
-    :class="{ 'single-line': syntaxHighlightedLines.length === 1 }"
     :data-syntax="syntaxNameNormalized"
-    @keydown.meta.a.prevent="selectCode"
-    @keydown.ctrl.a.prevent="selectCode"
+    :class="{ 'single-line': syntaxHighlightedLines.length === 1 }"
   >
     <Filename
       v-if="fileName"
@@ -24,7 +21,7 @@
       :fileType="fileType"
     >{{ fileName }}
     </Filename>
-    <div class="container-general" ref="code">
+    <div class="container-general">
       <!-- Do not add newlines in <pre>, as they'll appear in the rendered HTML. -->
       <pre><code><span
         v-for="(line, index) in syntaxHighlightedLines"
@@ -100,11 +97,6 @@ export default {
     },
   },
   methods: {
-    selectCode() {
-      window.getSelection().selectAllChildren(
-        this.$refs.code,
-      );
-    },
     isHighlighted(index) {
       return this.highlightedLineNumbers.has(this.lineNumberFor(index));
     },
--- a/src/components/GenericModal.vue
+++ b/src/components/GenericModal.vue
@@ -137,7 +137,7 @@ export default {
   },
   mounted() {
     this.focusTrapInstance = new FocusTrap();
-    document.addEventListener('keydown', this.onClick);
+    document.addEventListener('keydown', this.onKeydown);
     // add listeners for dynamic themes
     if (this.isThemeDynamic) {
       const matchMedia = window.matchMedia('(prefers-color-scheme: dark)');
@@ -155,7 +155,7 @@ export default {
     if (this.isVisible) {
       scrollLock.unlockScroll(this.$refs.container);
     }
-    document.removeEventListener('keydown', this.onClick);
+    document.removeEventListener('keydown', this.onKeydown);
     this.focusTrapInstance.destroy();
   },
   methods: {
@@ -212,7 +212,7 @@ export default {
      * @param {KeyboardEvent} params
      * @param {String} params.key
      */
-    onClick(event) {
+    onKeydown(event) {
       const {
         metaKey = false, ctrlKey = false, key,
       } = event;
--- a/src/components/GenericModal.vue
+++ b/src/components/GenericModal.vue
@@ -28,7 +28,7 @@
         >
           <CloseIcon />
         </button>
-        <div class="modal-content">
+        <div class="modal-content" ref="content">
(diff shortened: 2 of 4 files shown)


marinaaisa:
@swift-ci please test

src/components/ContentNode/CodeListing.vue
```diff
@@ -10,9 +10,12 @@
 
 <template>
   <div
+    tabindex="-1"
     class="code-listing"
-    :data-syntax="syntaxNameNormalized"
     :class="{ 'single-line': syntaxHighlightedLines.length === 1 }"
+    :data-syntax="syntaxNameNormalized"
+    @keydown.meta.a.prevent="selectCode"
```
mportiz08, line 17: I don't think this approach will work by adding this listener to this specific element, because then it only works after you have clicked the code listing and it has focus as the target of the events from ctrl+a.

mportiz08 (changes requested):
I'm still able to reproduce the issue in Chrome with a small window size after clicking a code listing and using ctrl+a/ctrl+v after the modal has opened—it only works for me if I have first clicked the code listing.

In general, I think this is more of a higher level problem with the modal components and not code listings specifically. The main issue is that the background text is still being selected in certain browsers when using ctrl+a even when a modal element has the foreground focus. Applying listeners to the code listing element specifically may impact how selection works for these listings even when they aren't in a modal, which probably isn't ideal.

marinaaisa:
> I'm still able to reproduce the issue in Chrome with a small window size after clicking a code listing and using cmd+a/cmd+v after the modal has opened—it only works for me if I have first clicked the code listing (after opening the modal).
> 
> In general, I think this is more of a higher level problem with the modal components and not code listings specifically. The main issue is that the background text is still being selected in certain browsers when using ctrl+a even when a modal element has the foreground focus. Applying listeners to the code listing element specifically may impact how selection works for these listings even when they aren't in a modal, which probably isn't ideal. We may want to focus on the modal element so that this fix can work more generally even when other elements besides code listings are inside the modal since it will likely have the same problem.

Thanks for your review, @mportiz08 , I totally agree.
In my latest commit I took a totally different approach that allow us to select modal's content when it's visible/open. It should fix the previous issues :)

marinaaisa:
@swift-ci please test

src/components/GenericModal.vue
```diff
@@ -190,6 +190,14 @@ export default {
     closeModal() {
       this.isVisible = false;
     },
+    /**
+     * Select all modal's content
+     */
+    selectContent() {
```
mportiz08, line 196: I like this approach better since it is more generic about the modal and not specific to code listings.

However, I think it would be even better if we could also avoid having to listen for the keyboard event and re-implementing the built-in behavior for actually selecting the text.

While thinking about this a bit, I'm wondering if we can simplify this PR to just applying the [`user-select: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) CSS style to everything that isn't the modal (the page background) when the modal is open. That way, we wouldn't need any JavaScript for keyboard events or actually doing the selection itself—essentially, the page just adds a CSS style to prevent selection only when the modal is visible.

What do you think @marinaaisa? (\cc @dobromir-hristov)

dobromir-hristov: I think that Safari does not respect `user-select: none` when you try to copy stuff. Its deff worth a shot though.

dobromir-hristov:
@marinaaisa the change looks good imho. Try what Marcus is suggesting, if it doesnt work we can leave it at that.

src/components/GenericModal.vue
marinaaisa: As @dobromir-hristov mentioned, I tried `-webkit-user-select: none;` on `main` tag in Safari and it doesn't prevent it to copy the content inside.

@mportiz08, although I always prefer CSS solutions to JS ones, your proposal would break the principle of component isolation since it needs all the rest of html blocks in the page to have a certain CSS property.

In my opinion, this modal's feature should be scoped to the modal component, even if that means re-implementing the built-in behavior.
Thanks!

mportiz08: > As @dobromir-hristov mentioned, I tried -webkit-user-select: none; on main tag in Safari and it doesn't prevent it to copy the content inside.

It is working for me when I try it, but maybe I'm missing something important.

> In my opinion, this modal's feature should be scoped to the modal component, even if that means re-implementing the built-in behavior.

I think that's generally a good guideline, but I feel like this desired behavior is an app-level feature where we're restricting selection to the background of the modal, so it seems reasonable to me that we could apply a CSS rule to a global element for that. The modal is kind of tricky due to the use of the portal. The solution you have here also depends on how the text is selected (it only works if done with the right keyboard shortcut), so it feels a little suboptimal but maybe that will end up being the only reasonable workaround.

Maybe I could try a quick prototype of what I had in mind, and we can see which strategy we prefer?

mportiz08: Nevermind. I tried my little prototype with the below code, and it doesn't work completely like I was hoping (disabling selection works but re-enabling it just for the modal seems to re-enable it for the full page in Safari for some reason).

```diff
diff --git a/src/components/Tutorial.vue b/src/components/Tutorial.vue
index a028671..d504590 100644
--- a/src/components/Tutorial.vue
+++ b/src/components/Tutorial.vue
@@ -9,7 +9,7 @@
 -->
 
 <template>
-  <div class="tutorial">
+  <div class="tutorial" :class="dynamicClasses">
     <NavigationBar
       v-if="!isTargetIDE"
       :technology="metadata.category"
@@ -25,7 +25,7 @@
       />
       <BreakpointEmitter @change="handleBreakpointChange" />
     </main>
-    <PortalTarget name="modal-destination" multiple />
+    <PortalTarget name="modal-destination" multiple @change="onPortalChange" />
   </div>
 </template>
 
@@ -85,7 +85,11 @@ export default {
     'isTargetIDE',
     'store',
   ],
+  data: () => ({ hasPortalContent: false }),
   computed: {
+    dynamicClasses: ({ hasPortalContent }) => ({
+      unselectable: hasPortalContent,
+    }),
     heroSection() {
       return this.sections.find(({ kind }) => kind === 'hero');
     },
@@ -125,6 +129,9 @@ export default {
     handleCodeColorsChange(codeColors) {
       CodeThemeStore.updateCodeColors(codeColors);
     },
+    onPortalChange(hasContent) {
+      this.hasPortalContent = hasContent;
+    },
   },
   created() {
     this.store.reset();
@@ -150,5 +157,13 @@ export default {
 
 .tutorial {
   background-color: var(--colors-text-background, var(--color-tutorial-background));
+
+  &.unselectable {
+    -webkit-user-select: none;
+  }
+}
+
+/deep/ .modal-open {
+  -webkit-user-select: text;
 }
 </style>
```

Unless someone else spots something obviously wrong with what I tried, we will likely need to go with the workaround for overriding the default behavior for ctrl+a :cry:

mportiz08:
@swift-ci test

src/components/GenericModal.vue
```diff
@@ -137,7 +137,7 @@ export default {
   },
   mounted() {
     this.focusTrapInstance = new FocusTrap();
-    document.addEventListener('keydown', this.onEscapeClick);
+    document.addEventListener('keydown', this.onClick);
```
mportiz08, line 140: Should we rename `onClick` to something else since a "click" is never involved with these events? Maybe `onKeyDown` or something?

mportiz08 (approved):
Seems to work as expected for special casing the ctrl+a selection for visible modal content.

src/components/GenericModal.vue
marinaaisa: done! https://github.com/apple/swift-docc-render/pull/39/commits/b3e5bb8643c9e34efcd8c3350bde4a4d06b4522b

marinaaisa:
@swift-ci test

PR #43, opened 2022-01-24
Fix to make sure InitialLoadingPlaceholder gets removed, if something goes wrong on initial route

dobromir-hristov:
Bug/issue #, if applicable: 80924187 

## Summary

This PR fixes a potential problem, where the router might error out, before its ready. This can happen if a third party client navigates to a page, before the initial page load is ready.

## Dependencies

NA

## Checklist

- [x] Added tests


mportiz08 (approved):
Seems like a reasonable change of using the same callback for the error scenario to ensure that the loading placeholder is still removed then (although we should do some additional testing before merging).

mportiz08:
@swift-ci test

PR #46, opened 2022-01-25
Add metadata tags

marinaaisa:
Bug/issue #32549376, if applicable: 

## Summary

This PR adds client side metadata tags to all pages in Swift DocC Render

Screenshot 2022-01-25 at 20 42 01

## Dependencies

NA

## Testing

Example fixture folder:
[manual-fixtures.zip](https://github.com/apple/swift-docc-render/files/7803705/manual-fixtures.zip)

Steps:
1. Use any data or docc documentation, example attached above
2. Run VUE_APP_DEV_SERVER_PROXY=/path/to/manual-fixtures/ npm run serve
3. Go to any page, like http://localhost:8080/documentation/framework/sloth if using example data
4. Assert that social media metadata is correct

## Checklist

- [x] Added tests


marinaaisa:
@swift-ci please test

src/utils/assets.js
```diff
@@ -75,3 +75,12 @@ export function normalizeAssetUrl(url) {
   }
   return pathJoin([baseUrl, url]);
 }
+
+/**
+ * Creates absolute URLs from paths
+ * @param {String} path
+ * @return {String}
+ */
+export function absoluteURL(path) {
```
mportiz08, line 84: Would it be possible to avoid the need for this function if the images were included in the assets folder and we used the normal static asset handling that vue-cli [provides](https://cli.vuejs.org/guide/html-and-static-assets.html)?

Either way, we will likely want to ensure that we utilize the `normalizeAssetUrl` function so that these images still work as expected when a custom base path is used for a developer's website (we should probably manually test this branch with a GitHub Pages workflow before merging to double check that there are no problems there).

src/mixins/metadata.js
```diff
@@ -0,0 +1,37 @@
...
+
+export default {
+  methods: {
+    /**
+     * Extract and join all the values from key 'text' on every object inside an array of objects,
+     * if there is an inlineContent array, it will extract all its text recursively as well
+     * @param {Array} content - array of objects containing "text" key
+     * @returns {String}
+     */
+    extractText(content) {
```
mportiz08, line 21: We should add some unit tests for this.

src/mixins/metadata.js
```diff
@@ -0,0 +1,37 @@
...
+      return content.reduce((acc, { text, inlineContent = [] }) => (
```
mportiz08, line 22: Definitely not a blocker, but it might be nice to create a generic `ContentNode.reduce` function in the future that works on a content tree in the same way that we did for `map` and `forEach` so that we could write more of these kinds of functions without having to think about and repeat the recursive logic that's needed each time.

src/mixins/metadata.js
```diff
@@ -0,0 +1,37 @@
...
+        acc.concat(text || this.extractText(inlineContent)).filter(Boolean)
+      ), []).join('');
+    },
+  },
+  computed: {
+    pagePath: ({ $route: { path = '/' } = {} }) => path,
+  },
+  created() {
+    addMetadata({
```
mportiz08, line 31: Maybe `addOrUpdateMetadata` might be a better name? We should also make sure that we aren't accidentally duplicating meta tags for the scenario where a user is transitioning between pages—it would be good to manually test the scenario where we go from a doc page to a tutorial (or vice versa) and make sure that we are only updating the things that are expected to change dynamically from page to page (maybe only the description?).

src/utils/metadata.js
```diff
@@ -0,0 +1,93 @@
...
+    property: 'og:site_name',
+    content: themeTitle,
+  },
+  {
+    property: 'og:type',
+    content: 'website',
+  },
+  {
+    property: 'og:image',
+    content: absoluteURL('/developer-og.jpg'),
```
mportiz08, line 34: Definitely not a blocker, but it would be cool to work with the design team in the future to enhance these images to be a more visual representation of DocC pages instead of just the "Developer Documentation" text 🤩

src/utils/metadata.js
```diff
@@ -0,0 +1,93 @@
...
+    property: 'og:url',
+    content: absoluteURL(path),
+  },
+  {
+    name: 'twitter:card',
+    content: 'summary_large_image',
+  },
+  {
+    name: 'twitter:image',
+    content: absoluteURL('/developer-og-twitter.jpg'),
```
mportiz08, line 54: Is the twitter-specific one just a different padding and/or aspect ratio? Just curious.

src/utils/metadata.js
```diff
@@ -0,0 +1,93 @@
...
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+import { getSetting } from 'docc-render/utils/theme-settings';
+import { absoluteURL } from 'docc-render/utils/assets';
+
+const themeTitle = getSetting(['meta', 'title'], process.env.VUE_APP_TITLE);
+const DEFAULT_DESCRIPTION = 'There\'s never been a better time to develop for Apple platforms.';
```
mportiz08, line 14: I know it's just a default, but we should double check this copy with the team before merging. The message might possibly sound too Apple-specific as-is.

mportiz08 (commented):
Great work @marinaaisa! This will be a great enhancement 🤩

dobromir-hristov (commented):
Marcus pointed out all the things that I wanted to say too. Good work.

src/utils/metadata.js
marinaaisa: I deleted the default description and made sure description tags gets deleted if user navigates from a page with description to a page without it. I think it makes more sense like this since a page description should provide specific page value and if we don't have that information we shouldn't display a default value.

marinaaisa: 2ff3d0dd1bca1ba340382c95e4e8fb83c75c6309

src/utils/metadata.js
marinaaisa: Different size and different ratio, yes. That's why they are different pictures.

src/utils/metadata.js
marinaaisa: I totally agree. I would love to implement automated images for each page!

src/mixins/metadata.js
marinaaisa: I updated the name: https://github.com/apple/swift-docc-render/pull/46/commits/d1b2f4ccad01ca1e2f5dd3e3dbffec618dafb321

Because I moved the static images to the HTML, now we are only updating the data that changes from one page to another: title, description and path. These params should always be different from one page to another, so we are good. Also I made sure to delete descriptions if page doesn't have so we don't inherit from previous page.

src/mixins/metadata.js
marinaaisa: I moved `extractText` inside ContentNode and added couple of tests for it: https://github.com/apple/swift-docc-render/pull/46/commits/79661b0867292c50c3cd4a6b990f53ea1cfb4a48

src/mixins/metadata.js
marinaaisa: https://github.com/apple/swift-docc-render/pull/46/commits/79661b0867292c50c3cd4a6b990f53ea1cfb4a48

src/mixins/metadata.js
```diff
@@ -0,0 +1,37 @@
...
+    extractText(content) {
+      return content.reduce((acc, { text, inlineContent = [] }) => (
+        acc.concat(text || this.extractText(inlineContent)).filter(Boolean)
+      ), []).join('');
+    },
+  },
+  computed: {
+    pagePath: ({ $route: { path = '/' } = {} }) => path,
+  },
+  created() {
```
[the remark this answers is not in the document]
marinaaisa: Done

src/utils/assets.js
marinaaisa: I moved the images to the HTML as you suggested, thanks for that!

The reason why I was using this `absoluteURL` function is because we can only use absolute urls for metadata (instead of relative ones as in the rest of the app) but it's true that `normalizeAssetUrl` should render the proper absolute path when being built, so it should be fine.

marinaaisa:
@swift-ci please test

src/mixins/metadata.js
mportiz08: I don't think I communicated what I meant with this comment well enough, sorry.

I just opened a PR that better captures my intent with this feedback: https://github.com/marinaaisa/swift-docc-render/pull/1

Please let me know what you think—I think it will be slightly safer for varying types of content and will also provide some generic building blocks for doing other transformations in the future if needed.

src/mixins/metadata.js
mportiz08: Ah, I see that you changed the name, but we need to also change the logic to handle the update case. When I click a link to a tutorial page from a documentation page, I'm observing that there are duplicated `<meta>` tags for both pages on the second page.

We'll need to update the logic for creating these nodes to first check if they exist and update them instead of creating/adding them in that scenario.

src/utils/metadata.js
```diff
@@ -0,0 +1,93 @@
...
+ * It adds or updates client-side metadata tags to pages
+ * @param {Object} pageData
+ */
+// eslint-disable-next-line import/prefer-default-export
+export function addOrUpdateMetadata({ title, description, path }) {
+  const formattedTitle = formatTitle(title);
+  // add title
+  addTitle(formattedTitle);
+  // create and add metadata tags
+  createMetaTags({ title: formattedTitle, description, path }).forEach(
```
dobromir-hristov, line 90: We should probably try and remove the old meta tags, before we add the new ones.

src/mixins/metadata.js
marinaaisa: Done! https://github.com/apple/swift-docc-render/pull/46/commits/731200820b793e12030a9566b072973230bf09c7

marinaaisa:
@swift-ci please test

src/utils/metadata.js
```diff
@@ -0,0 +1,100 @@
...
+    property: 'og:title',
+    content: title,
+  },
+  {
+    property: 'og:description',
+    content: description,
+  },
+  {
+    property: 'og:url',
+    content: normalizeAssetUrl(path),
```
mportiz08, line 41: I think we're now correctly accounting for a base path when it has been configured (thanks!), but I noticed that we are now creating metadata with relative paths as opposed to absolute URLs, and it seems that the open graph protocol specifies that URLs should be full/absolute URLs with scheme/host/etc.

I went ahead and created a PR to turn these relative paths into full/absolute URLs: https://github.com/marinaaisa/swift-docc-render/pull/3

Let me know what you think!

src/utils/metadata.js
```diff
@@ -0,0 +1,100 @@
...
+    name: 'twitter:url',
+    content: normalizeAssetUrl(path),
+  },
+];
+const formatTitle = title => [title, themeTitle].filter(Boolean).join(' | ');
+const addMetaTags = (metadata) => {
+  const keys = Object.keys(metadata);
+  // check current metadata
+  const metaQuery = `meta[${keys[0]}="${metadata[keys[0]]}"]`;
+  if (document.querySelector(metaQuery)) {
```
mportiz08, line 65: Thanks for updating this to make sure we aren't duplicating metadata elements!

This is a very minor note, but I think we should save the `document.querySelector(metaQuery)` result as a variable so that we aren't repeating this selector query multiple times every place it's used.

mportiz08 (commented):
Almost there!

tests/unit/utils/metadata.spec.js
```diff
@@ -92,4 +93,11 @@ const assertMetadata = ({
 describe('Metadata', () => {
   assertMetadata(pageWithTitleAndDescription);
   assertMetadata(pageWithoutTitleOrDescription);
+
+  it('does not inherit metadata from previous pages', () => {
```
dobromir-hristov, line 97: You can also add a test case, whether it removes tags, that are not in the new page.

src/utils/metadata.js
```diff
@@ -0,0 +1,100 @@
...
+  {
+    name: 'twitter:url',
+    content: normalizeAssetUrl(path),
+  },
+];
+const formatTitle = title => [title, themeTitle].filter(Boolean).join(' | ');
+const addMetaTags = (metadata) => {
+  const keys = Object.keys(metadata);
+  // check current metadata
+  const metaQuery = `meta[${keys[0]}="${metadata[keys[0]]}"]`;
```
dobromir-hristov, line 64: 0 here is the first key, so `name` or `property` ?

mportiz08: https://github.com/apple/swift-docc-render/pull/46/commits/0546336e20a909fa61dbeb58eabb9432d2617600

mportiz08 (approved):

PR #48, opened 2022-02-07
Improve swift declaration formatting

mportiz08:
Bug/issue #, if applicable: 88594221

## Summary

This implements a new formatting strategy for Swift declarations whereby each parameter of a multi-parameter symbol will be broken onto its own line and indented.

This should be an improvement over the way Swift declarations are currently rendered where they simply wrap at arbitrary points with some indentation. It should be easier for viewers of the documentation to visually and mentally scan/parse functions/initializers/etc for their parameters when symbols are presented in this manner.

**Example:**

Before:
Screen Shot 2022-02-04 at 13 42 46
After:
Screen Shot 2022-02-04 at 13 42 57

## Testing

Steps:
1. Download/unzip [ExamplePackage.doccarchive.zip](https://github.com/apple/swift-docc-render/files/8019198/ExamplePackage.doccarchive.zip) and configure `VUE_DEV_SERVER_PROXY=/path/to/ExamplePackage.doccarchive` (or use an alternate fixture with Swift symbols)
2. Run the dev server with `npm run serve` and use the example docs at http://localhost:8080/documentation/examplepackage to verify that symbols are formatted as described above or compare with existing data.

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -88,7 +88,7 @@ export default {
 
         // loop through the token text to look for "(" and ")" characters
         // eslint-disable-next-line no-plusplus
-        for (let k = 0; k < token.text.length; k++) {
+        for (let k = 0; k < (token.text || '').length; k++) {
           if (token.text.charAt(k) === '(') {
             numUnclosedParens += 1;
             // keep track of the token/character position of the first "("
@@ -112,7 +112,8 @@ export default {
         // if we find some text ending with ", " and the next token is the start
         // of a new param, update this token text to replace the space with a
         // newline followed by 4 spaces
-        if (token.text.endsWith(', ') && nextToken && nextToken.kind === TokenKind.externalParam) {
+        if (token.text && token.text.endsWith(', ')
+          && nextToken && nextToken.kind === TokenKind.externalParam) {
           token.text = `${token.text.trimEnd()}\n    `;
           indentedParams = true;
         }
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -56,7 +56,7 @@ export default {
       language,
       formattedSwiftTokens,
       tokens,
-    }) => (language === Language.swift.api ? formattedSwiftTokens : tokens),
+    }) => (language === Language.swift.key.api ? formattedSwiftTokens : tokens),
     // Return a formatted version of the tokens array, with additional
     // indentation whitespace to break parameters onto individual lines for
     // improved readability and scanning of Swift functions/initializers.
@@ -84,6 +84,7 @@ export default {
       while (i < tokens.length) {
         // keep track of the current token and the next one (if any)
         const token = tokens[i];
+        const newToken = { ...token };
         const nextToken = j < tokens.length ? tokens[j] : undefined;
 
         // loop through the token text to look for "(" and ")" characters
@@ -114,11 +115,11 @@ export default {
         // newline followed by 4 spaces
         if (token.text && token.text.endsWith(', ')
           && nextToken && nextToken.kind === TokenKind.externalParam) {
-          token.text = `${token.text.trimEnd()}\n    `;
+          newToken.text = `${token.text.trimEnd()}\n    `;
           indentedParams = true;
         }
 
-        newTokens.push(token);
+        newTokens.push(newToken);
         i += 1;
         j += 1;
       }
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -12,7 +12,7 @@
   <pre
     ref="declarationGroup"
     class="source"
-    :class="{ indented: simpleIndent, [multipleLinesClass]: hasMultipleLines }"
+    :class="{ [multipleLinesClass]: hasMultipleLines }"
   ><code ref="code"><Token
     v-for="(token, i) in tokens"
     :key="i"
@@ -39,10 +39,6 @@ export default {
       type: Array,
       required: true,
     },
-    smartIndent: {
-      type: Boolean,
-      default: false,
-    },
     simpleIndent: {
       type: Boolean,
       default: false,
@@ -65,7 +61,7 @@ export default {
   async mounted() {
     if (hasMultipleLines(this.$refs.declarationGroup)) this.hasMultipleLines = true;
 
-    if (!this.smartIndent || !this.language) return;
+    if (!this.language) return;
     await this.$nextTick();
     indentDeclaration(this.$refs.code, this.language);
   },
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -91,8 +91,9 @@ export default {
         const nextToken = j < tokens.length ? tokens[j] : undefined;
 
         // loop through the token text to look for "(" and ")" characters
+        const tokenLength = (token.text || '').length;
         // eslint-disable-next-line no-plusplus
-        for (let k = 0; k < (token.text || '').length; k++) {
+        for (let k = 0; k < tokenLength; k++) {
           if (token.text.charAt(k) === '(') {
             numUnclosedParens += 1;
             // keep track of the token/character position of the first "("
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -57,6 +57,18 @@ export default {
       formattedSwiftTokens,
       tokens,
     }) => (language === Language.swift.api ? formattedSwiftTokens : tokens),
+    // Return a formatted version of the tokens array, with additional
+    // indentation whitespace to break parameters onto individual lines for
+    // improved readability and scanning of Swift functions/initializers.
+    //
+    // This is implemented in a single pass loop where text tokens are updated
+    // at certain key points to insert spaces and newlines so that each
+    // parameter of a multi-parameter function gets its own line.
+    //
+    // @param {Array} tokens The original syntax tokens.
+    //   See `DeclarationToken.props`
+    // @return {Array} A formatted version of the original tokens.
+    //   See `DeclarationToken.props`
     formattedSwiftTokens: ({ tokens }) => {
       let indentedParams = false;
       const newTokens = [];
@@ -68,14 +80,18 @@ export default {
       let closeParenCharIndex = null;
       let numUnclosedParens = 0;
 
+      // loop through every declaration token
       while (i < tokens.length) {
+        // keep track of the current token and the next one (if any)
         const token = tokens[i];
         const nextToken = j < tokens.length ? tokens[j] : undefined;
 
+        // loop through the token text to look for "(" and ")" characters
         // eslint-disable-next-line no-plusplus
         for (let k = 0; k < token.text.length; k++) {
           if (token.text.charAt(k) === '(') {
             numUnclosedParens += 1;
+            // keep track of the token/character position of the first "("
             if (openParenCharIndex == null) {
               openParenCharIndex = k;
               openParenTokenIndex = i;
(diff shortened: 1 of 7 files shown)


mportiz08:
@swift-ci test

bobergj:
Missing a test and example for a generic method declaration with a long list of where clauses.
```
func f(t: T, u: U) where T: Sequence, U: Sequence, T.Iterator.Element: Equatable, T.Iterator.Element == U.Iterator.Element
```
If it isn't handled, perhaps it should be done at the same time.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -52,6 +51,111 @@ export default {
...
+      // loop through every declaration token
+      while (i < tokens.length) {
+        // keep track of the current token and the next one (if any)
+        const token = tokens[i];
+        const newToken = { ...token };
+        const nextToken = j < tokens.length ? tokens[j] : undefined;
+
+        // loop through the token text to look for "(" and ")" characters
+        // eslint-disable-next-line no-plusplus
+        for (let k = 0; k < (token.text || '').length; k++) {
```
dobromir-hristov, line 99: I think you could extract the `.length` invocation outside the `for`, so it does not call it for every iteration.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -52,6 +51,111 @@ export default {
...
+      let openParenCharIndex = null;
+      let closeParenTokenIndex = null;
+      let closeParenCharIndex = null;
+      let numUnclosedParens = 0;
+
+      // loop through every declaration token
+      while (i < tokens.length) {
+        // keep track of the current token and the next one (if any)
+        const token = tokens[i];
+        const newToken = { ...token };
```
dobromir-hristov, line 94: You clone the token, so you can augment it, while still iterating over the real one, right?

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -52,6 +51,111 @@ export default {
...
+        // of a new param, update this token text to replace the space with a
+        // newline followed by 4 spaces
+        if (token.text && token.text.endsWith(', ')
+          && nextToken && nextToken.kind === TokenKind.externalParam) {
+          newToken.text = `${token.text.trimEnd()}\n${indent}`;
+          indentedParams = true;
+        }
+
+        newTokens.push(newToken);
+        i += 1;
```
dobromir-hristov, line 130: Why do you have both `i` and `j` ? Is `j` just `i + 1`?

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -52,6 +51,111 @@ export default {
...
+        j += 1;
+      }
+
+      // if we indented some params, we want to find the opening "(" symbol
```
dobromir-hristov, line 134: So the first part shifts params to a new line between each other, while these two passages move the params into new lines, in between the braces?
```
func(foo, 
  bar, 
  baz)
```
becomes
```
func(
  foo, 
  bar, 
  baz
)
```

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -52,6 +51,111 @@ export default {
...
+      // and add a newline and 4 spaces to the end of it, breaking the first
+      // param onto its own line
+      if (indentedParams && openParenTokenIndex !== null) {
+        const originalText = newTokens[openParenTokenIndex].text;
+        const begin = originalText.slice(0, openParenCharIndex);
+        const end = originalText.slice(openParenCharIndex);
+        const newText = `${begin}${end}\n${indent}`;
```
dobromir-hristov, line 141: Here wont it work to just append `/n${indent}`, why do we need to break up the text, then put it back together?

dobromir-hristov (commented):
Awesome job Marcus. Left a few comments so I am sure I understand the logic.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -52,6 +51,111 @@ export default {
...
+    // future with the raw text before it gets tokenized for syntax
+    // highlighting—however, this post-tokenization JavaScript logic should be
+    // an improvement until that kind of work can be integrated.
+    //
+    // @param {Array} tokens The original syntax tokens.
+    //   See `DeclarationToken.props`
+    // @return {Array} A formatted version of the original tokens.
+    //   See `DeclarationToken.props`
+    formattedSwiftTokens: ({ tokens }) => {
+      const numSpacesForIndent = 4; // maybe this could be configurable in the future
```
franklinsch, line 78: Maybe something that could be part of theme settings JSON? 👀

mportiz08:
> Missing a test and example for a generic method declaration with a long list of where clauses.
> 
> ```
> func f(t: T, u: U) where T: Sequence, U: Sequence, T.Iterator.Element: Equatable, T.Iterator.Element == U.Iterator.Element
> ```
> 
> If it isn't handled, perhaps it should be done at the same time.

Thanks for the feedback! I can add a unit test for this to ensure that the logic that inspects commas does not get tripped up by the multiple conditions in the where clause.

Based on the narrow focus of this PR (only breaking params), I would expect this particular declaration to be rendered as follows:
```swift
func f(
    t: T,
    u: U,
) where T: Sequence, U: Sequence, T.Iterator.Element: Equatable, T.Iterator.Element == U.Iterator.Element
```

That may not be your expectation since it is admittedly still kind of hard to quickly scan. However, doing more advanced formatting like splitting up the where clause conditions would be out of the scope of this PR—it would be great to expand on this in the future though!

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -52,6 +51,111 @@ export default {
...
+    // Ideally this should be implemented with a tool like SwiftFormat in the
+    // future with the raw text before it gets tokenized for syntax
+    // highlighting—however, this post-tokenization JavaScript logic should be
+    // an improvement until that kind of work can be integrated.
+    //
+    // @param {Array} tokens The original syntax tokens.
+    //   See `DeclarationToken.props`
+    // @return {Array} A formatted version of the original tokens.
+    //   See `DeclarationToken.props`
+    formattedSwiftTokens: ({ tokens }) => {
```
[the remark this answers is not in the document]
mportiz08: Hm, we can do that, although it seems a little unnecessary since this should only be called (and only is called today) when the language is Swift. Technically, if it _were_ to be called in this situation, the code would run, but the tokens would be returned unchanged as expected since none of the triggering conditions would apply to realistic Objective-C symbol declarations.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: Yep. I mainly added this to ensure that `formattedSwiftTokens` is non-mutating. In my first implementation, it was directly updating the existing token since I was working on the original object instead of copying it to return an array of new objects.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: Yes. `j` is just the next index (`i + 1`) so that we can peek at the next token to find the situation where one token is followed by another specific kind of token. I could remove the variable, but then we will have to repeat the `i + 1` expression in a couple of places—I don't mind making that change if it's confusing having so many variables.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: Exactly. There are 3 places where newlines are added:
* after each comma separating params (N - 1 times where N is number of params)
* after opening paren (1 time)
* before closing paren (1 time)

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: I probably was just thinking too hard. I'll look at making that update haha 😅

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: I think that sounds reasonable. I purposefully wrote this in a way where the number could be easily be updated by something along those lines as opposed to hardcoding in the spaces manually, but maybe I should just go ahead and make that just work.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: a20c63c9d51414407412887a076ada75e3f051ff 😆

mportiz08:
> > Missing a test and example for a generic method declaration with a long list of where clauses.
> > ```
> > func f(t: T, u: U) where T: Sequence, U: Sequence, T.Iterator.Element: Equatable, T.Iterator.Element == U.Iterator.Element
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If it isn't handled, perhaps it should be done at the same time.
> 
> Thanks for the feedback! I can add a unit test for this to ensure that the logic that inspects commas does not get tripped up by the multiple conditions in the where clause.
> 
> Based on the narrow focus of this PR (only breaking params), I would expect this particular declaration to be rendered as follows:
> 
> ```swift
> func f(
>     t: T,
>     u: U,
> ) where T: Sequence, U: Sequence, T.Iterator.Element: Equatable, T.Iterator.Element == U.Iterator.Element
> ```
> 
> That may not be your expectation since it is admittedly still kind of hard to quickly scan. However, doing more advanced formatting like splitting up the where clause conditions would be out of the scope of this PR—it would be great to expand on this in the future though!

[00a1e4adafa07a9345eb6fccc735102624fd591c](https://github.com/apple/swift-docc-render/pull/48/commits/00a1e4adafa07a9345eb6fccc735102624fd591c)

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -39,10 +42,6 @@ export default {
       type: Array,
       required: true,
     },
-    smartIndent: {
-      type: Boolean,
-      default: false,
-    },
     simpleIndent: {
```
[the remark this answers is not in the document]
mportiz08: yeah 64a340685a52acf82ff54ea40941d05784d91624

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -65,7 +169,7 @@ export default {
   async mounted() {
     if (hasMultipleLines(this.$refs.declarationGroup)) this.hasMultipleLines = true;
 
-    if (!this.smartIndent || !this.language) return;
+    if (!this.language) return;
```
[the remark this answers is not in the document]
mportiz08: sure 64a340685a52acf82ff54ea40941d05784d91624

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: 70e4bafaba541d782be6245e0d5d70e65f9eb550

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: I was initially hesitant to add this until we've been able to make this file more complete and better documented its use cases, but I've added support now with the knowledge that we will soon expose this information in a more discoverable way.

6c9b11d56bd9a4cdb0bda8fcbda399921b0e9aae

mportiz08:
@swift-ci test

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
dobromir-hristov: nah its fine, just wanted to make sure I understand. We can add a comment just for readability.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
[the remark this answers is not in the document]
dobromir-hristov: Alright, leave it as is.

dobromir-hristov (approved):

marinaaisa (approved):
Quite interesting and great implementation. It also works pretty well. Amazing job, Marcus 👏
Screenshot 2022-02-09 at 15 57 46

PR #51, opened 2022-02-16
Add a sidenav navigator to the documentation pages (WIP behind a feature flag, off by default)

dobromir-hristov:
Bug/issue #, if applicable: 89031049

## Summary

This PR introduces a few new UI changes, that are scoped only to the documentation pages. The features are all DISABLED, by default, but can be enabled by a feature flag in the theme-settings.json file, under `features.docs.navigator.enable`:

1. New adjustable layout
2. A sidebar tree navigator
3. Updated nav layout and breadcrumbs logic

This work covers both of the proposals:

https://forums.swift.org/t/new-swift-docc-content-styling/55326
https://forums.swift.org/t/swift-docc-sidebar/55250

## Dependencies

https://github.com/apple/swift-docc/pull/84

## Testing

Make sure you download and unzip the `.doccarchive` - [SlothCreator.doccarchive.zip](https://github.com/apple/swift-docc-render/files/8082216/SlothCreator.doccarchive.1.zip)

Steps:
1. run `VUE_APP_DEV_SERVER_PROXY=/path/to/SlothCreator.doccarchive npm run serve`
2. Visit http://localhost:8080/documentation/slothcreator
3. Assert that a sidebar tree navigator is rendered.
4. Assert that you can open/close items and click on the links to navigate to a page
5. Assert that navigating across pages does not loose open/closed state. Even on refresh.
5. Assert that going back/forward works and re-highlights the current item you are on, opening all parents so its visible.
6. Assert you can filter on items and its not slowing down the browser even for very broad queries.
8. Assert that the navigation layout does not go outside of the content container
9. Assert that the navigation breadcrumbs reduce in size and collapse, based on screen size.
9. Assert that the content has a maximum width, after which it gets centered (around 1800px)

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #68, opened 2022-03-04
Fix regression with page layout where it can wrap horizontally in certain scenarios

mportiz08:
Bug/issue #, if applicable: 89837079

## Summary

This fixes a regression with the page layout that was introduced in #51.

In certain edge cases for IDE targets, top level elements of documentation pages like the "Relationships" section may occasionally break out horizontally into a second column, which should never happen.

This seems to only happen in WebKit based browsers.

## Testing

Steps:
1. Configure `VUE_APP_TARGET=ide` when running the dev server with `npm run serve`.
2. Using Safari, find examples of problematic pages and verify that they now look normal as the window is resized from large to small.
3. View other pages and ensure that there are no new regressions with the layout.

## Checklist

- [ ] Added tests

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -351,7 +351,7 @@ export default {
   @include inTargetIde {
     min-height: 100vh;
     display: flex;
-    flex-flow: column wrap;
+    flex-flow: column nowrap;
     border: none;
 
     & > .contenttable:last-of-type {


mportiz08:
@swift-ci test

dobromir-hristov (approved):
I have introduced this when I added height:100% to a few of the outer wrapping divs. This was to make sure the content is always at least 100% of the screen.

I found other fixes but this one is deff the shortest so I think its good.

dobromir-hristov:
@swift-ci test

mportiz08:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #78, opened 2022-03-10
Enable the sidebar only for schema versions above certain value

dobromir-hristov:
Bug/issue #, if applicable: 89089597

## Summary

Enables the sidebar, only for RenderJSON that has a schemaVersion equal or above `0.3.0`.

## Dependencies

NA

## Testing

Steps:
1. Run against an older doccarchive. Assert it does not show the sidebar.
2. Run against a new doccarchive with version `0.3.0` and assert it shows the sidebar.

## Checklist

- [x] Added tests


src/utils/schema-version-check.js
```diff
@@ -10,14 +10,45 @@
...
+      return 1;
+    }
+    if (v1parts[i] > v2parts[i]) {
+      return 1;
+    }
+    if (v1parts[i] < v2parts[i]) {
+      return -1;
+    }
+  }
+  // the two versions have different lengths, return early
```
mportiz08, line 43: Was this meant to appear before the loop? If not, I'm not sure the comment about returning _early_ is very accurate :P

src/utils/schema-version-check.js
```diff
@@ -10,14 +10,45 @@
...
 }
 
+/**
+ * Compares two version strings
+ * @param {string} v1 - first version to compare
+ * @param {string} v2 - second version to compare
+ * @returns {Number<-1|0|1>}
+ */
+export function compareVersions(v1, v2) {
+  const v1parts = v1.split('.');
```
mportiz08, line 28: We should probably convert the parts into integers, even if the JS comparison of integer strings happens to work in the way that you might expect.

mportiz08, line 28: We should probably convert the parts into integers, even if the JS comparison of integer strings happens to work in the way that you might expect.

tests/unit/utils/schema-version-check.spec.js
```diff
@@ -75,4 +79,27 @@ describe('schema-version-check', () => {
...
     expect(warnSpy).toHaveBeenCalledTimes(0);
   });
+
+  describe('compareVersions', () => {
+    it.each([
+      ['1.1.1', '1.1.10', -1],
+      ['1.1.1', '1.2.10', -1],
+      ['1.1.20', '1.2.10', -1],
+      ['1.1.1', '2.1.1', -1],
+      ['1.1', '1.1.0', -1],
```
mportiz08, line 89: Shouldn't this return `0`? I think it's unlikely for us to see this kind of comparison with realistic data, but it still seems odd to me that we would treat `1.1` as being less than `1.1.0`.

mportiz08, line 89: Shouldn't this return `0`? I think it's unlikely for us to see this kind of comparison with realistic data, but it still seems odd to me that we would treat `1.1` as being less than `1.1.0`.

mportiz08:
(Sorry, not sure why GitHub duplicated all my comments)

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

mportiz08:
@swift-ci test

mportiz08:
@swift-ci test

PR #82, opened 2022-03-14
Fix navigator tag filtering logic - stop showing none-matching children

dobromir-hristov:
Bug/issue #, if applicable: 89588542

## Summary

Fixes the filtering logic of the navigator, stop showing children of a parent, that is a direct match of your current query. It fixes the logic for tags and title filtering. It was already fixed for API changes.

## Dependencies

NA

## Testing

Steps:
1. Type a filter that matches a parent item directly.
2. Assert none of it's children, that dont match the same query are shown.
3. This works the same for Tags. If you pick a tag, it should only show children with the same tag. (Their parents dont count, as they have to be rendered in order to show the children).

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -502,13 +502,11 @@ export default {
      */
     generateNodesToRender() {
       const {
-        children, filteredChildren, hasFilter, openNodes, apiChanges,
+        children, filteredChildren, hasFilter, openNodes,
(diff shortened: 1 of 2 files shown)


mportiz08:
Hm, running into issues testing this locally where the tagging functionality doesn't work anymore. I may be doing something wrong, but I'll reach out to you in person.

mportiz08 (approved):

PR #86, opened 2022-03-17
Make sure pages continue working, even if bad reference data is provided

dobromir-hristov:
Bug/issue #, if applicable: r90244085

## Summary

Prevents component from hanging, when there are missing references to the breadcrumb items.

## Dependencies

[SlothCreator.doccarchive.zip](https://github.com/apple/swift-docc-render/files/8283556/SlothCreator.doccarchive.zip)

NA

## Testing

Use the archive above

Steps:
1. Go to http://localhost:8080/documentation/slothcreator
2. Assert the page still gets rendered

## Checklist

- [x] Added tests


dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

src/components/DocumentationTopic/DocumentationNav/Hierarchy.vue
```diff
@@ -137,9 +137,11 @@ export default {
   computed: {
     parentTopics() {
       return this.parentTopicIdentifiers.map((id) => {
-        const { title, url } = this.references[id];
+        const { title, url } = this.references[id] || {};
```
mportiz08, line 140: Instead of simply not crashing in this situation, maybe we should additionally do something like `console.warn` to indicate that there was a missing reference since this is an unexpected/bad state where something went wrong with the compiler or otherwise with the data?

src/components/Navigator.vue
```diff
@@ -95,7 +95,10 @@ export default {
   computed: {
     // gets the paths for each parent in the breadcrumbs
     parentTopicReferences({ references, parentTopicIdentifiers }) {
-      return parentTopicIdentifiers.map(identifier => references[identifier]);
+      return parentTopicIdentifiers
+        .map(identifier => references[identifier])
```
mportiz08, line 99: Not a blocker, but it may be slightly more optimized to use a single `reduce` loop as opposed to the two loops needed for a `map` and subsequent `filter`.

src/views/DocumentationTopic.vue
```diff
@@ -197,7 +197,7 @@ export default {
     // hierarchy/breadcrumb for a given topic. We choose to render only the
     // first one.
     parentTopicIdentifiers: ({ topicProps: { hierarchy: { paths: [ids = []] = [] } } }) => ids,
-    navigatorParentTopicIdentifiers: ({ topicProps: { hierarchy: { paths = [] } } }) => (
+    navigatorParentTopicIdentifiers: ({ topicProps: { hierarchy: { paths = [[]] } } }) => (
```
mportiz08, line 200: Part of the reason I'm hesitant to add these kinds of defaults (for data required to be there by the spec) is that it can be easy to get into a rabbit hole of how far down we have to add these defaults.

For example, should we guard against when there is nothing in `hierarchy`? How about when `hierarchy` is not defined or `null`? Ok, but what about when `references` itself is null? We can guard every single layer if we choose to, but it quickly becomes a lot of added code for situations we shouldn't be hitting and don't always have an obvious fallback UI for.

In this case, we will prevent crashing in some very specific scenarios but the navbar itself will be empty even when that happens, which may be even more confusing for the user. Not saying we shouldn't do this, we just need to be careful about being overly defensive about data that is expected to be present.

src/views/DocumentationTopic.vue
```diff
@@ -207,17 +207,25 @@ export default {
...
+      // create an end-case fallback technology
+      const fallback = { title, url: $route.path };
+      // get the reference to the current page
+      const currentPageReference = references[identifier];
+      // if there are no parent topics at all, use the current reference or a fallback one
+      if (!parentTopicIdentifiers.length) return currentPageReference || fallback;
+      // try to use the first parent topic reference, if not a technology kind
       const first = references[parentTopicIdentifiers[0]];
-      if (first.kind !== 'technologies') return first;
+      if (first && first.kind !== 'technologies') return first;
```
mportiz08, line 218: I don't think this is necessary since there is already a guard for `parentTopicIdentifiers.length` above this—there is always guaranteed to be a first item here.

dobromir-hristov: Its not the same. That checks if there are any identifiers. Here I am making sure that there was a reference for that identifier at `0`.

src/views/DocumentationTopic.vue
dobromir-hristov: I understand what you mean. Would you like me to revert this path fallback?

src/components/DocumentationTopic/DocumentationNav/Hierarchy.vue
dobromir-hristov: Will add a `console.error` here. In the future we may want to have a more refined user notification process, but that should do for now.

dobromir-hristov:
@swift-ci test

src/views/DocumentationTopic.vue
mportiz08: no

src/views/DocumentationTopic.vue
```diff
@@ -207,17 +207,25 @@ export default {
...
+      // itself, manufacture a minimal one using other available data.
+      // This is to guard against bad data or missing references.
+      if (role === TopicRole.collection && !currentPageReference) {
+        return fallback;
       }
-
-      return references[parentTopicIdentifiers[1]] || references[identifier];
+      // use the second parent topic identifier, if there was a first,
+      // otherwise fallback the current page reference or the final fallback
+      return (first && references[parentTopicIdentifiers[1]]) || currentPageReference || fallback;
```
mportiz08, line 228: Since this will now always return something, what's the experience when the last-resort `fallback` doesn't make sense? For example, what if the current page used as the fallback is not actually a technology that would have an index? Do things just crash in a different way in subsequent code or would we already handle that gracefully? (I think it might already be covered by the data fetching logic but just making sure)

dobromir-hristov:
@swift-ci test

src/views/DocumentationTopic.vue
dobromir-hristov: Fetching will work, but the way we match the technology might not 🤔  We do a compare against the `currentTechnology.url === indexTechnology.path` and that might backfire. I will see if I can fix that logic as well.

mportiz08:
Your latest change may have broken a unit test.

> DocumentationTopic › renders the Navigator and AdjustableSidebarWidth when enabled

dobromir-hristov:
@swift-ci test

mportiz08:
@swift-ci test

mportiz08 (approved):

PR #87, opened 2022-03-17
Properly limit the tags on the FilterInput

dobromir-hristov:
Bug/issue #, if applicable: 90383936

## Summary

This fixes the logic to truncate the suggestedTags inside the FilterInput.

## Dependencies

NA

## Testing

Steps:
1. Make sure the FilterInput should never show more than 5 items, when ShouldTruncateTags is true.

## Checklist

- [x] Added tests


marinaaisa (approved):
Good catch! Too bad we didn't have a unit test for this 🤦‍♀️
Thanks for fixing it!

dobromir-hristov:
@swift-ci test

PR #88, opened 2022-03-21
Fix persisted state when navigating across technologies, while filter is applied

dobromir-hristov:
Bug/issue #, if applicable: 90558554

## Summary

Fixes an issue with the sidebar navigator, where it shows no children, even if there are. 
This can happen if you go to another technology, search and then go back.

## Dependencies

NA

## Testing

Steps:
1. Go to any page on a technology
2. In the navigator, click on a link that points to another technology
3. Search for something
4. Go back using the browser Back
5. Assert it does not apply the filter, (its cleared from SessionStorage as well), and it properly renders the children.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -360,6 +360,7 @@ export default {
       this.filter = '';
       this.debouncedFilter = '';
       this.selectedTags = [];
+      this.resetScroll = true;
     },
     scrollToFocus() {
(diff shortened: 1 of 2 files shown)


dobromir-hristov:
@swift-ci test

mportiz08:
@swift-ci test

mportiz08 (approved):
The code for this is getting pretty hard to follow, but the fix seems to be working well.

PR #89, opened 2022-03-21
[Regression] In sidenav input loses focus after updating the results

marinaaisa:
Bug/issue #90432388, if applicable: 

## Summary

In sidenav input loses focus after updating the results

## Dependencies

NA

## Testing

Steps:
1. Run any sidenav documentation
2. Search on filter input
3. Assert that when typing, focus doesn't go to sidenav
4. Assert that sidenav keyboard actions work properly

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -55,6 +55,7 @@
               :expanded="openNodes[item.uid]"
               :api-change="apiChangesObject[item.path]"
               :isFocused="focusedIndex === index"
+              :lastFocusWasInside="lastFocusWasInside"
               @toggle="toggle"
               @toggle-full="toggleFullTree"
               @navigate="setActiveUID"
@@ -332,6 +333,9 @@ export default {
     isLargeBreakpoint: ({ breakpoint }) => breakpoint === BreakpointName.large,
     hasNodes: ({ nodesToRender }) => !!nodesToRender.length,
     totalItemsToNavigate: ({ nodesToRender }) => nodesToRender.length,
+    lastFocusWasInside: ({ isInsideScroller, lastFocusTarget }) => (
+      isInsideScroller(lastFocusTarget)
+    ),
   },
   created() {
     this.restorePersistedState();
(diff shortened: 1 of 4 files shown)


marinaaisa:
@swift-ci please test

src/components/Navigator/NavigatorCard.vue
```diff
@@ -55,6 +55,7 @@
               :expanded="openNodes[item.uid]"
               :api-change="apiChangesObject[item.path]"
               :isFocused="focusedIndex === index"
+              :lastFocusWasInside="lastFocusWasInside"
```
dobromir-hristov, line 58: isnt this the same?
```suggestion
              :isFocused="lastFocusWasInside && focusedIndex === index"
```

dobromir-hristov (commented):
I can reproduce the problem but I dont understand what causes it. What changes the lastFocusItem to be the first item?

src/components/Navigator/NavigatorCard.vue
marinaaisa: It's not exactly the same because if we set lastFocusWasInside as a condition to focus we will never get to focus in the first place. You can try it out to see what I mean. But I'm open to solve it in another way.

dobromir-hristov (commented):
I did some further investigation, because I did not understand what causes this.

I noticed the watcher on the `activeIndex` gets triggered when we perform a search, which does NOT include the currentPage item, aka `activeUID`, in the nodesToRender.

This sets the `focusIndex` to `0`, which ofc causes the first item to get focused, and trigger the Focus handler.

At that point however, the scroller re-renders and that first item is no more in the DOM, but we already have a reference to it. 

Your fix just makes sure that reference we have stored, is the one that is in the DOM.

I think we should take a step back and see if we can do a proper fix, instead of just adding guards. If there is nothing else we can do, then this is also an option.

marinaaisa:
> I did some further investigation, because I did not understand what causes this.
> 
> I noticed the watcher on the `activeIndex` gets triggered when we perform a search, which does NOT include the currentPage item, aka `activeUID`, in the nodesToRender.
> 
> This sets the `focusIndex` to `0`, which ofc causes the first item to get focused, and trigger the Focus handler.
> 
> At that point however, the scroller re-renders and that first item is no more in the DOM, but we already have a reference to it.
> 
> Your fix just makes sure that reference we have stored, is the one that is in the DOM.
> 
> I think we should take a step back and see if we can do a proper fix, instead of just adding guards. If there is nothing else we can do, then this is also an option.

Hey @dobromir-hristov , I tried some other solutions that didn't reach any point :(
Do you want to give it a try?

dobromir-hristov:
Superseded in #112

PR #90, opened 2022-03-22
Allow toggling all siblings when cmd + click-ing on a tree item.

dobromir-hristov:
Bug/issue #, if applicable: 89797030

## Summary

When cmd + click-ing on a dropdown in the navigator, it should open it and all of its siblings.

## Dependencies

NA

## Testing

Steps:
1. Cmd + click on an item
2. Assert all of its siblings are open. 
3. Assert that previously open items stay open and dont get extra items appended.
4. Assert item order and hierarchy is not broken
5. Cmd + click to close
6. Assert all are closed and can then be individually opened.
7. Assert that root level items also open.
8. Assert that deeply open children of siblings are closed when cmd+click-ing on an item.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -57,6 +57,7 @@
               :isFocused="focusedIndex === index"
               @toggle="toggle"
               @toggle-full="toggleFullTree"
+              @toggle-siblings="toggleSiblings"
               @navigate="setActiveUID"
             />
           </RecycleScroller>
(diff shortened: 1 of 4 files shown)


dobromir-hristov:
@swift-ci test

src/components/Navigator/NavigatorCard.vue
```diff
@@ -461,9 +462,37 @@ export default {
...
+        if (isOpen) {
+          const children = this.getAllChildren(uid);
+          // remove all children
+          children.forEach((child) => {
+            delete openNodes[child.uid];
+          });
+          // remove the sibling as well
+          delete openNodes[uid];
+          // augment the nodesToRender
+          this.augmentRenderNodes({ uid, exclude: children.slice(1), include: [] });
```
marinaaisa, line 480: Why not excluding all children here instead?

marinaaisa (approved):
Interesting behaviour... it works as expected :)

src/components/Navigator/NavigatorCard.vue
dobromir-hristov: because getAllChildren returns the children + the current node. So I just slice off the current node.

dobromir-hristov:
@swift-ci test

PR #91, opened 2022-03-22
Change styling of selected/hover item states for sidenav items

marinaaisa:
Bug/issue #89558736, if applicable: 

## Summary

- [x] We should no longer aim for a rounded radius.
- [x] Selected and hover state of items should span the full width of the sidebar.
- [x] The top level items in the sidebar should also be aligned with the top level technology title.

## Dependencies

NA

## Testing

Steps:
1. Run `npm run serve` with your documentation
2. Enable sidenav
3. Assert that summary list is correct  

## Checklist

- [x] Added tests (not needed)

### DIFF
--- a/src/components/Navigator/NavigatorCardItem.vue
+++ b/src/components/Navigator/NavigatorCardItem.vue
@@ -211,7 +211,7 @@ $item-height: 32px;
   position: relative;
   display: flex;
   align-items: center;
-  border-radius: $border-radius;
+  padding-left: calc(var(--card-horizontal-spacing) / 2);
   flex: 1;
   min-width: 0;
   height: 100%;
--- a/src/styles/core/_nav.scss
+++ b/src/styles/core/_nav.scss
@@ -13,7 +13,7 @@ $nav-height: rem(52px);
 $nav-height-small: rem(48px);
 $nav-padding: rem(22px);
 $nav-padding-small: rem(16px);
-$nav-padding-wide: rem(30px);
+$nav-padding-wide: rem(20px);
 $nav-menu-item-left-margin: rem(24px);
 $nav-timingfunction: ease-in !default;
 $nav-menu-tray-transition: max-height 0.4s $nav-timingfunction 0s, visibility 0s linear 1s !default;
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -865,7 +865,6 @@ $filter-height: 64px;
   --card-horizontal-spacing: #{$navigator-card-horizontal-spacing};
   --card-vertical-spacing: #{$navigator-card-vertical-spacing};
 
-  padding: 0 var(--card-horizontal-spacing);
   // right padding is added by the items, so visually the scroller is stuck to the side
   padding-right: 0;
   flex: 1 1 auto;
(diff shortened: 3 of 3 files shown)


src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -211,7 +211,7 @@ $item-height: 32px;
   position: relative;
   display: flex;
   align-items: center;
-  border-radius: $border-radius;
+  padding-left: calc(var(--card-horizontal-spacing) / 2);
```
dobromir-hristov, line 214: I feel like the `--card-horizontal-spacing` var no longer serves the purpose it was meant for. We should do some cleanup.

I also dont understand why we use half of the spacing here. It feels like a happy coincidence.

src/styles/core/_nav.scss
```diff
@@ -13,7 +13,7 @@ $nav-height: rem(52px);
 $nav-height-small: rem(48px);
 $nav-padding: rem(22px);
 $nav-padding-small: rem(16px);
-$nav-padding-wide: rem(30px);
+$nav-padding-wide: rem(20px);
```
dobromir-hristov, line 16: I feel like we could have just synced the design with the normal spacing, and use 22px. Removing these 2px hardly makes a difference, just clutters the code.

src/components/Navigator/NavigatorCardItem.vue
marinaaisa: I tried using better and self-explanatory vars in this new commit: https://github.com/apple/swift-docc-render/pull/91/commits/5d63fd14d364ba4caa4911d01a2d5cca848a136a

src/styles/core/_nav.scss
marinaaisa: To simplify, I deleted the nav-padding-wide var and the design seems to be alright, can you check if I introduced any regression with it? 

https://github.com/apple/swift-docc-render/pull/91/commits/575e31a28aaef1ec94c57fa1b82ac54ee08c709d

dobromir-hristov: Well yes. Now its 2px off :) But honestly I think that we should just unify the rest to that.

CC @SamLanier 

image

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -182,36 +182,43 @@ export default {
 $item-height: 32px;
 
 .navigator-card-item {
+  --tree-toggle-padding: var(--card-horizontal-spacing-small);
+  --chevron-width: var(--card-horizontal-spacing);
+  --nesting-spacing: var(--card-horizontal-spacing) + var(--card-horizontal-spacing-small);
```
dobromir-hristov, line 187: This is technically not a valid addition. It only works, because the var is directly replaced with `a + b` in a `calc()`

dobromir-hristov (commented):
It may just be me, but I feel like all these Custom CSS Properties are unnecessary. I personally tend to use them, when I need to do some math (like we saw here), or something to be easily changed for a specific scope, like inside a breakpoint or from a class, for example. 

Here we dont do that, its just as easy to have a few sass vars. Not a big deal, just pointing it out.

I also noticed that with the latest changes to the styling, we also changed the spacing of the items, in relation to the parent. Will need a designer to look into this - @SamLanier 

image

marinaaisa:
Closed in favor of: https://github.com/apple/swift-docc-render/pull/96

PR #99, opened 2022-03-25
Allow toggling sidebar navigator on/off, for large breakpoints

dobromir-hristov:
Bug/issue #, if applicable: 89564104

## Summary

Allows toggling the navigator ON/OFF, for large resolutions. This is done through the Nav toggle icon or by dragging.
Users can drag the navigator all the way to the minimum allowed size. Once they pass 2x the minimum, the sidebar gets hidden. 

The user can expand it back out, if they have not let go of the mouse button upon dragging OR by clicking on the nav toggle. (should work similar to Mail.app)

Hidden state is persisted in localStorage, so it wont re-open for users who have closed it once.

## Dependencies

NA

## Testing

Steps:
1. Try to click on the toggle to open/close the navigator on Large screen sizes. Assert nothing bleeds out when hidden.
2. Make sure tabbing across the page does not focus anything inside the navigator
3. Refresh the page to assert its kept closed.
4. Open the navigator and start dragging all the way until it stops. Now drag 2x the left space and assert its toggled off.
5. Without releasing your drag, try to expand out again. It should snap to the smallest possible size again.
6. If you drag until closed, it should persist in LS the closed state, and later allow you to toggle open again.

## Checklist

- [x] Added tests


dobromir-hristov:
@swift-ci test

biamx3:
This is a great start.

My biggest piece of feedback is we need to enable the sidebar side by side with the content in M viewports. I know this is a pain and you'll need to rework a lot of the logic, but when I designed that behavior, the sidebar button wasn't in scope. 

Screen Shot 2022-03-25 at 6 13 38 PM

S viewports should remain as they are. If you click on the sidebar icon it will bring up the modal with the filter on top. 

My other feedback are small visual things: 
1. The. sidebar icon should be glyph-gray. It looks disabled right now. 
2. The sidebar icon should align with the items in the sidebar. 
3.  Let's add a separator between the icon and the Documentation item of the breadcrumb. 

Screen Shot 2022-03-25 at 5 54 58 PM

biamx3:
This is a great start.

My biggest piece of feedback is we need to enable the sidebar side by side with the content in M viewports. I know this is a pain and you'll need to rework a lot of the logic, but when I designed that behavior, the sidebar button wasn't in scope.

S viewports should remain as they are. If you click on the sidebar icon it will bring up the modal with the filter on top.

My other feedback are small visual things:
1. The sidebar icon should be glyph-gray. It looks disabled right now.
2. The sidebar icon should align with the items in the sidebar.
3. Let's add a separator between the icon and the Documentation item of the breadcrumb.

Screen Shot 2022-03-25 at 6 32 33 PM

dobromir-hristov:
@biamx3 I updated the PR to your specs. Let me know how it looks.

marinaaisa (commented):
Nice job, congrats! 😄 

It works pretty well, I left couple of comments tho:

- I noticed that the Sidenav toggle's focus outline is wider than the hover style.
Screenshot 2022-03-30 at 22 52 55

- The navigator head item's focus outline has been modified as well. It seems that in main is not working totally right either, but here is worse. Let me know if you prefer that I create another radar for this one.

Screenshot 2022-03-30 at 22 58 52

src/components/Navigator.vue
```diff
@@ -24,7 +24,9 @@
       @close="$emit('close')"
     />
     <div v-else class="loading-placeholder">
-      Fetching...
+      <div class="loading-placeholder-content">
```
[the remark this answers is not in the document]
dobromir-hristov: You can, but then it wont apply the overflow. The container gets squished to almost 0, and elements that have padding continue to show the content. 

In this case, the wording was getting smushed but still partially visible for a brief moment as you toggle, which was annoying.

dobromir-hristov:
> I noticed that the Sidenav toggle's focus outline is wider than the hover style.

Will try to fix that one.

> The navigator head item's focus outline has been modified as well. It seems that in main is not working totally right either, but here is worse. Let me know if you prefer that I create another radar for this one.

Thats because I probably added `overflow: hidden`, so things dont shine through, when we toggle the navigator to the minimum size.

dobromir-hristov:
@marinaaisa those should be fixed now.

dobromir-hristov:
@swift-ci test

marinaaisa (approved):
Looks good, thanks Dobri!

dobromir-hristov:
@swift-ci test

biamx3:
Awesome improvements Dobri! This is looking great. 

I noticed that when I click on the navigator  button, a secondary button shows up right next to it which is jarring. 

Screenshot 2022-07-01 at 2 42 19 PM

Could we make it look like there's only one button that moves to the left and sticks there?

dobromir-hristov:
Well they are different components, so its going to be allot of pixel guessing to get a transition between the two.

What about something like this?

Screen Recording 2022-07-02 at 8 23 13
?

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

SamLanier (approved):
Design approved.

dobromir-hristov:
@swift-ci test

src/components/DocumentationTopic/DocumentationNav/LanguageToggle.vue
```diff
@@ -125,12 +125,12 @@ export default {
   },
   mounted() {
     // on resize, re-calculate the width of the select.
-    const cb = debounce(async () => {
+    const cb = throttle(async () => {
```
marinaaisa, line 128: Could you explain me this change? Thanks

marinaaisa (changes requested):
Code looks good!
I just found couple of AX improvements:

1) Focus state of the button to close the navigation should be a rectangle around the button. Now it looks like this Screenshot 2022-08-01 at 15 33 23

2) Focus state gets lost when open and close navigator buttons are triggered. As a keyboard user I would expect to keep the focus state in the same element

https://user-images.githubusercontent.com/8567677/182160086-a492a561-e29b-41d0-8fc1-0a81a53333d4.mov

src/components/DocumentationTopic/DocumentationNav/LanguageToggle.vue
dobromir-hristov: I think Bea noticed the language toggle was not getting the new screen size, when we resize the page by dragging the adjustable sidebar. Debounce can skip the last function hit, while throttle will always fire the last function invocation.

dobromir-hristov:
@marinaaisa first issue I can probably figure something out, though its another hack 😆 I hate overflow: hidden.

For the second one, what do you suggest we do? Should the focus move from one toggle to the other?

marinaaisa:
> For the second one, what do you suggest we do? Should the focus move from one toggle to the other?

Yes, I understand that there is two different toggles but the user only see one. We should keep the focus in the same one.

dobromir-hristov:
@swift-ci test

PR #100, opened 2022-03-25
Prevent arrows from animating, when scrolling too fast

dobromir-hristov:
Bug/issue #, if applicable: 89840827

## Summary

This PR prevents the arrows from animating open/closed, when scrolling long lists too fast.

## Dependencies

NA

## Testing

Steps:
1. Open a few levels in a tree in different hierarchies.
2. Start scrolling fast up/down
3. Assert it does not animate the arrows upon rendering on the screen.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCardItem.vue
+++ b/src/components/Navigator/NavigatorCardItem.vue
@@ -196,8 +196,8 @@ export default {
       }
     },
     async expanded() {
-      // wait a few frames for animations queues to finish
-      await waitFrames(8);
+      // wait for 9 frames (60hz * 0.15ms = 9), for animations queues to finish.
+      await waitFrames(9);
       // set the opening animation as ended
       this.idState.isOpening = false;
     },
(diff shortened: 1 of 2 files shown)


dobromir-hristov:
@swift-ci test

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -176,6 +195,12 @@ export default {
         this.selfFocus();
       }
     },
+    async expanded() {
+      // wait a few frames for animations queues to finish
+      await waitFrames(8);
```
mportiz08, line 200: I'm guessing there isn't, but is there possibly a more reliable way of doing this as opposed to waiting for 8 frames?

If not, maybe it should actually be 9 since the the animation duration is 0.15 seconds and the refresh rate is probably most commonly 60hz (60 * .15 = 9). Even if my math is wrong, it might be good to add to the comment that the number of frames is largely estimated to match the duration of the CSS transition.

dobromir-hristov: Right, will do.

marinaaisa: Thanks for the math comment, I had no idea how to calculate those frames 😅

marinaaisa (approved):
Interesting the use of IdState and waiting frames! 💯

mportiz08:
@swift-ci test

mportiz08 (approved):

PR #104, opened 2022-03-29
Fix tutorial title truncation in nav

dobromir-hristov:
Bug/issue #, if applicable: 90843609

## Summary

This PR fixes the titles on Tutorial pages navigation bar from being truncated on small viewport sizes.

This is a regression from introducing the pre-title slot for the sidenav toggle.

## Dependencies

NA

## Testing

Steps:
1. Go to a tutorial
2. Resize screen to a mobile viewport
3. Assert it does not truncate the title anymore

## Checklist

- [ ] Added tests

### DIFF
--- a/src/components/Tutorial/NavigationBar.vue
+++ b/src/components/Tutorial/NavigationBar.vue
@@ -218,6 +218,13 @@ export default {
     .mobile-dropdown-container {
       display: block;
     }
+    /deep/ .nav-title {
+      grid-area: title;
+    }
+
+    /deep/ .pre-title {
+      display: none;
+    }
   }
 }
 


src/components/Tutorial/NavigationBar.vue
```diff
@@ -218,6 +218,13 @@ export default {
     .mobile-dropdown-container {
       display: block;
     }
+    /deep/ .nav-title {
+      grid-area: title;
+    }
+
+    /deep/ .pre-title {
```
marinaaisa, line 225: We don't have any place where we set `pre-title` to a `grid-area` but we use it in `grid-template-areas`, which shouldn't work.

marinaaisa (changes requested):

src/components/Tutorial/NavigationBar.vue
marinaaisa: I mean, this PR works but I don't understand why we use `pre-title` if it's not set to a grid-area.

dobromir-hristov: I think its assumed to be in that, because its the first item? The grid follows the natural order of items, top left to bottom right.

I think the bug was happening, because the `.title` was being mushed into the `pre-title` grid-area, because of a `grid-column: 1` rule we had defined earlier. Even though it has a `grid-area: title` that was getting overridden.

dobromir-hristov:
@swift-ci test

PR #105, opened 2022-03-29
Fix to not highlight Tutorials, when navigating to them from the navigator links, then going back.

dobromir-hristov:
Bug/issue #, if applicable: 90840717

## Summary

This fix prevents keeping tutorials highlighted, when clicking on them in the Navigator, then going back.

This was caused because we kept an activeUID for that item upon clicking, but it navigates you to a different page type altogether. 

The fix is a more generic one, to clear the cached state, if the activeUID path is NOT the same as the current route path.

**Alternative**
An alt fix could be to not emit a click event for Tutorial types, but that is a slippery slope I think.

## Dependencies

NA

## Testing

Steps:
1. Open an archive with a tutorial
2. Click on a none-tutorial item
3. Then click on a tutorial
4. Assert you get navigated to the tutorial
5. Go back with browser Back
6. Assert the tutorial is no longer highlighted, but the current page is.

## Checklist

- [x] Added tests

### DIFF
--- a/app/public/theme-settings.json
+++ b/app/public/theme-settings.json
@@ -58,7 +58,7 @@
         "hide": false
       },
       "navigator": {
-        "enable": true
+        "enable": false
       }
     }
   }
--- a/app/public/theme-settings.json
+++ b/app/public/theme-settings.json
@@ -58,7 +58,7 @@
         "hide": false
       },
       "navigator": {
-        "enable": false
+        "enable": true
       }
     }
   }
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -104,7 +104,7 @@ import { TopicTypes } from 'docc-render/constants/TopicTypes';
 import FilterInput from 'docc-render/components/Filter/FilterInput.vue';
 import { BreakpointName } from 'docc-render/utils/breakpoints';
 import keyboardNavigation from 'docc-render/mixins/keyboardNavigation';
-import { last } from '@/utils/arrays';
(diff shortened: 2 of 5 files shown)


src/components/Navigator/NavigatorCard.vue
```diff
@@ -104,6 +104,7 @@ import { TopicTypes } from 'docc-render/constants/TopicTypes';
 import FilterInput from 'docc-render/components/Filter/FilterInput.vue';
 import { BreakpointName } from 'docc-render/utils/breakpoints';
 import keyboardNavigation from 'docc-render/mixins/keyboardNavigation';
+import { last } from '@/utils/arrays';
```
mportiz08, line 107:
```suggestion
import { last } from 'docc-render/utils/arrays';
```

src/utils/arrays.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const last = array => array[array.length - 1];
```
mportiz08, line 2: I realize it's very simple, but we should add unit tests for this if we're going to add this helper. Also, this new file will need the Swift open source license preamble.

app/public/theme-settings.json
```diff
@@ -58,7 +58,7 @@
         "hide": false
       },
       "navigator": {
-        "enable": false
```
marinaaisa, line 61: I don't think this change was intended for this PR :)

marinaaisa (approved):
Looks good! I left a small comment, thanks Dobri! ✨

mportiz08:
@swift-ci test

PR #117, opened 2022-04-04
Fix issues with losing state when navigating or opening/closing items, while filtering.

dobromir-hristov:
Bug/issue #, if applicable: 90843762

## Summary

This PR fixes a few issues we have noticed during testing:

- [x] rdar://90841554 ([TESTFEST] Navigation collapses on searched term when performing a search, expanding and then selecting a subsection)
- [x] rdar://90843762 ([TESTFEST] Filtering for Sample Code yields inconsistent UI experience)
- [x] Navigating across pages while filtering, re-opens all previously closed items.

## Dependencies

NA

## Testing

You can use any doccarchive with an index.

Steps:
1. Apply a generic enough filter
2. Start opening/closing items.
3. Assert there are no items being lost when opening/closing.
4. Try to find a parent item that has no children matches. Assert it renders as closed initially, but you can expand and see it's children.
5. Try cmd + click to open all siblings. Assert its all working as expected.
6. Try alt + click to open all nested children. Assert they are all rendered in correct order.
7. Close some items and start navigating across links in the page OR by clicking navigator items. 
9. Assert the collapsed items are not opened, unless your current page is in one of them.
10. Try the same with going back/forward in the browser.
11. Clear the filters and assert the only open items are the ones leading to the current page.
12. Assert that opening/closing items without a filter works as expected.

## Checklist

- [x] Added tests


src/components/Navigator/NavigatorCard.vue
```diff
@@ -838,6 +860,9 @@ export default {
         this.focusIndex(0);
       }
     },
+    convertChildrenArrayToObject(children) {
+      return Object.fromEntries(children.map(child => [child.uid, child]));
```
mportiz08, line 864: Not blocking, but as a very minor performance improvement, you may want to rewrite this in a single pass loop using `reduce` instead of looping twice (one for `fromEntries` and one for `map`

mportiz08, line 864: Not blocking, but as a very minor performance improvement, you may want to rewrite this in a single pass loop using `reduce` instead of looping twice (one for `fromEntries` and one for `map`

src/utils/arrays.js
```diff
@@ -8,5 +8,8 @@
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
-// eslint-disable-next-line import/prefer-default-export
 export const last = array => array[array.length - 1];
+
+export const isEqual = (first = [], second = []) => (
```
mportiz08, line 13: Minor thing, but I can't think of a good use case for this function having the default arguments of empty arrays.

mportiz08, line 13: Minor thing, but I can't think of a good use case for this function having the default arguments of empty arrays.

tests/unit/utils/arrays.spec.js
```diff
@@ -18,4 +18,37 @@ describe('arrays', () => {
...
+
+    it('returns false, if the two arrays are different', () => {
+      expect(isEqual(first, second)).toBe(false);
+    });
+
+    it('returns false, if the array does not have the same item types', () => {
+      expect(isEqual(first, firstStrings)).toBe(false);
+    });
+
+    it('does not error out, if passed undefined', () => {
```
mportiz08, line 43: I think you could take this out since I don't think this would ever be an expected way to call this function (with only a single array).

mportiz08, line 43: I think you could take this out since I don't think this would ever be an expected way to call this function (with only a single array).

src/components/Navigator/NavigatorCard.vue
```diff
@@ -316,6 +316,34 @@ export default {
...
+     * This is used on both toggling, as well as on navigation and filtering.
+     * @return {Object.<string, NavigatorFlatItem>}
+     */
+    renderableChildNodesMap({ filteredChildrenUpToRootSet, childrenMap, hasFilter }) {
+      if (!hasFilter) return childrenMap;
+      const all = [];
+      // create a set of all matches and their parents
+      filteredChildrenUpToRootSet.forEach((current) => {
+        // if it has no children, just add it
+        if (!current.childUIDs.length) return all.push(current);
```
mportiz08, line 337: This is probably working the way that you expect, but it's a little confusing to read since the `forEach` isn't expected to actually return anything, and you're possibly just using the returns to break the control flow(?). Also, `Array.push` returns a count which adds to the confusing way this reads, to me at least.

It might be helpful to rewrite this either using `Array.reduce` with `Array.concat` to add/return items or just a simple for loop without return statements for imperative actions like `Array.push`.

Or am I maybe just misunderstanding something here?

mportiz08, line 337: This is probably working the way that you expect, but it's a little confusing to read since the `forEach` isn't expected to actually return anything, and you're possibly just using the returns to break the control flow(?). Also, `Array.push` returns a count which adds to the confusing way this reads, to me at least.

It might be helpful to rewrite this either using `Array.reduce` with `Array.concat` to add/return items or just a simple for loop without return statements for imperative actions like `Array.push`.

Or am I maybe just misunderstanding something here?

mportiz08:
Sorry, GitHub decided to double all my review comments...

dobromir-hristov:
@swift-ci test

src/components/Navigator/NavigatorCard.vue
dobromir-hristov: Yeah it was to break to flow. Refactored it a bit. I cant use `reduce`, because I am looping a `Set` and I wanted to not convert it back to an array again.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

PR #122, opened 2022-04-05
Suggest only tags in the Navigator, that can return results

dobromir-hristov:
Bug/issue #, if applicable: 89482723

## Summary

Change the navigator filter to only show tags, that have a match in the nav items

## Dependencies

NA

## Testing

Steps:
1. Open a page with the navigator
2. Click on the Nav filter
3. Assert the tags that are shown, only the ones we have content for. Based on on your archive, it may be all, some or even none.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #138, opened 2022-04-08
[5.7] Suggest only tags in the Navigator, that can return results

dobromir-hristov:
- **Rationale:** We should only suggest tags that can yield results
- **Risk:** Low
- **Risk Detail:** The change is only scoped to the Navigator filters, but it does need to iterate over all items of a technology.
- **Reward:** High
- **Reward Details:** Users will not see tags, that would not yield any results
- **Original PR:** https://github.com/apple/swift-docc-render/pull/122
- **Issue:** rdar://89482723
- **Code Reviewed By:** @hqhhuang 
- **Testing Details:** Added Unit Tests + manually tested in the browser

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -250,11 +250,35 @@ export default {
       if (errorFetching) return ERROR_FETCHING;
       return NO_CHILDREN;
     },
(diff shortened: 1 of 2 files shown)


franklinsch (approved):

dobromir-hristov:
@swift-ci test

PR #147, opened 2022-04-11
Highlight current item in Navigator, when there is a trailing slash in the URL

dobromir-hristov:
Bug/issue #, if applicable: 91443710

## Summary

This PR fixes a bug Ethan found on [Argument Parser docs](https://ethan-kusters.github.io/swift-argument-parser/documentation/argumentparser/argumenthelp/shoulddisplay/), where docc-render would not highlight the current page in the navigator. 

This turns out to be caused by the trailing slash in the URL, which was causing the current page matching to fail.

## Dependencies

NA

## Testing

Steps:
1. Using any docc-archive, visit a page and just add a trailing slash at the end of the URL
2. Make sure you clear any SessionStorage cache
3. Assert the current page gets highlighted

## Checklist

- [x] Added tests

### DIFF
--- a/tests/unit/components/Navigator.spec.js
+++ b/tests/unit/components/Navigator.spec.js
@@ -6,7 +6,7 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
- */
+*/
 
 import Navigator from '@/components/Navigator.vue';
 import { shallowMount } from '@vue/test-utils';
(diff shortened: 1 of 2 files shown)


tests/unit/components/Navigator.spec.js
```diff
@@ -6,7 +6,7 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
-*/
+ */
```
mportiz08, line 9:
```suggestion
*/
```

(Probably auto modified by your editor)

mportiz08 (approved):
Nice catch!

Looks like one unexpected (to me) consequence of using directories and an `index.html` for each page (as opposed to just using the page name as the HTML file name) is that the GitHub Pages server auto redirects to the version of the URL with the trailing slash, so this particular bug would impact any site deployed there 😅

dobromir-hristov:
@swift-ci test

PR #148, opened 2022-04-11
Simplify the matching pattern in the Navigator filter

dobromir-hristov:
Bug/issue #, if applicable: 91443929

## Summary

This PR simplifies the filter matching pattern on the Navigator. Now it wont highlight leading/trailing spaces on potential matches.

**Old Behaviour:**

`FooBar` would match `FooBar` and `Foo Bar`.
`Foo` would match `Foo` and ` Foo `

**Now:**
`FooBar` matches only `FooBar`
`Foo` matches only `Foo`

Matches are case-insensitive, so casing doesnt matter much.

## Dependencies

NA

## Testing

Steps:
1. Serve using any archive with an index
2. Type in the filter input on the Navigator a single word
3. Assert only the word is highlighted as matching in the results.

## Checklist

- [x] Added tests

### DIFF
--- a/src/utils/strings.js
+++ b/src/utils/strings.js
@@ -186,11 +186,14 @@ export function whiteSpaceIgnorantRegex(stringToSanitize) {
     char = trimmedString[i];
     // if the character is an escape char, pass it and the next character
     if (char === '\\') {
+      // pass both escape char and char, with an empty space match before, only if not first char
+      collector.push(`${i === 0 ? '' : spaceMatch}${char}`);
+      collector.push(trimmedString[i + 1]);
       // skip one character in next iteration
       i += 1;
-      // pass both escape char and char, with an empty space match before
-      collector.push(`${spaceMatch}${char}`);
-      collector.push(trimmedString[i]);
+    } else if (i === 0) {
+      // skip the first character, if its not a `\`
+      collector.push(char);
       // add anything else, but empty spaces
     } else if (char !== singleSpace) {
       collector.push(`${spaceMatch}${char}`);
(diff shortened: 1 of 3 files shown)


dobromir-hristov:
@swift-ci test

biamx3:
Getting less matches for a query seems like a design regression. The original issue that needed a fix was just the highlighting of space before a word. For example, if a query is " help", the highlight should only be on top of "help", not the space. We should keep highlighting the space between two different words, like for example "foo bar".

dobromir-hristov:
OK Let me explain a bit more. The matching pattern we had before would match a query with any spaces, no mater where and how much. So if I have `func help`, `funchel p` or `funchelp`, they would all match `func help` text. 
Because we match any number of spaces, between all the letters in the query, it was also highlighting the space before `help`, when just typing `help`.

My change removes that space ignoring, and tries to match what you typed.

**Before:**

image
image
image
image

**After**
image
image
image
image


**Alternative:**
If we really want to keep the old more relaxed matching, one other easy way to fix this would be to stop matching spaces before the first character of the query. This can be done in the `whiteSpaceIgnorantRegex` function, where we can just pass over the first char, if its not a `\`.

cc @mportiz08 @marinaaisa @hqhhuang

biamx3:
Got it. I'm still concerned folks will accidentally type in an extra space or miss a space, so my preference is to go for the more forgiving previous matching. 

The second solution you described makes sense to me.

dobromir-hristov:
Updated. @biamx3 @mportiz08 @hqhhuang give it another look.

dobromir-hristov:
@swift-ci test

hqhhuang (changes requested):
Change looks good, I think we still need to fix a failing test in: `search-utils.spec.js`

hqhhuang:
@swift-ci test

hqhhuang (approved):
LGTM, let's get @SamLanier and @biamx3 to review again before merging.

SamLanier (approved):
Design approved.

dobromir-hristov:
@swift-ci test

PR #155, opened 2022-04-12
Improve the children flattening performance in the Navigator

dobromir-hristov:
Bug/issue #, if applicable: 91610791

## Summary

This PR refactors the tree flattening logic to improve its performance.

**Results**
For 40k children, I saw about 100ms reduction, from ~340ms to sub ~220ms in Chrome. Similar results also in Safari.

## Dependencies

NA

## Testing

Serve with a very large doccarchive, with lots of index items.

Steps:
1. You can use the Performance benchmark of you browser. Chrome has one to test on page load.

## Checklist

- [ ] ~Added tests~ - no performance tests so far


mportiz08:
(The test index I used is one of the biggest ones we have available to us and seems pretty realistic)

dobromir-hristov:
Here are some results using an archive with ~40k index items.

Going to track the time it takes to process the downloaded data `fetchDataIndex` (make it reactive), flatten it down `flatChildren` and then render the items in the virtual scroller `RecycleScroller patch`.

### No Changes

**900ms (223ms fetchDataIndex , 404ms flatChildren, 126ms RecycleScroller patch)**

image

### Changes from this PR

**862ms (335ms fetchDataIndex, 271ms flatChildren, 120ms RecycleScroller patch)** 

We see the index process time took another 100ms (kind of random), but the `flatChildren` is now almost in half.

image

### Using `Object.freeze`

This is something I meant to use but forgot yesterday. 

**345ms (0.27ms fetchDataIndex, 113ms flatChildren, 109ms RecycleScroller patch)**

As you can see, the improvement is even better, and is very much noticeable on page refresh.

image

mportiz08:
Sorry, I was incorrectly reporting the "Self Time" value instead of the "Total Time" value, but I think they still tell the same story since the change here only impact the function itself and none of the functions it calls.

Either way, I tried again and got these results (for Total Time of `flattenNestedData`):
351.2ms (39f18f2276c07234e25d01b7f13a8d0ec5aa988f, parent commit of this PR)
385.9ms (1d3dbe974bf83b62f3afe84262fe89bf8045bdee, `for` loop commit)
266.3ms (ecaab0094e1d742d9897b8e4f4015cc040bef592, `Object.freeze` commit)

Note that I actually got slightly _worse_ performance with the for loop. That's not to say that it's any worse by any means—I think it just points to the fact that the differences we're both seeing are more likely just statistical variance and not any direct performance impact of the change itself.

I _did_ see a consistent improvement with the `Object.freeze` commit. However, I'm also not convinced that it had anything to do with the change to freezing the tree, since I was actually testing with a different `NavigatorDataProvider` entirely which wouldn't even use that code...I'm not sure what that means lol

With all that being said, I don't have a problem with merging any of these changes—I just can't see how they would have any major performance improvement, and I haven't actually seen that improvement on the test runs I've done on my own machine.

mportiz08 (approved):

marinaaisa:
Really interesting conversation and tools on this PR, thank you both of you! 👏

dobromir-hristov:
@swift-ci test

PR #160, opened 2022-04-12
add nvmrc

QuietMisdreavus:
This PR adds an `.nvmrc` file so that users can use [nvm](https://github.com/nvm-sh/nvm) to quickly install the versions of node and npm requested in the `package.json` file. This can greatly speed up onboarding new contributors to docc-render, since they can use nvm to get the right toolset and not worry about incorrect versions causing problems when processing dependencies.


QuietMisdreavus:
@swift-ci Please test

mportiz08 (approved):

mportiz08:
@swift-ci test

PR #47, opened 2022-01-27
Add more extendability to tutorial navs

dobromir-hristov:
Bug/issue #, if applicable: 88095565

## Summary

Fixes minor styling leftovers from prior refactors and allows for the tutorial navigations to be more extendable by external clients.

## Dependencies

NA

## Testing

Steps:
1. Mount using an archive with tutorials
2. Make sure the navigation looks identical on both TutorialOverview and TutorialTopic pages.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #162, opened 2022-04-12
Fix uneven divider in nav bar for small viewport 

hqhhuang:
Bug/issue #, if applicable: 91372980

## Summary
Fix uneven divider in nav bar for the small viewport. This only happens when we use `NavMenuItem` component (not `NavMenuItems`
Fixed by overriding line height for the small viewport.

## Checklist

- [ ] Added tests - NA


hqhhuang:
@swift-ci test

dobromir-hristov (commented):
This is the exact change that broke it. I am not sure whether we should revert that `min-height` to `height` or keep this fix. https://github.com/apple/swift-docc-render/pull/47/files#diff-0e3113fe0f5966efbcbf73dbccb8c910fd794de61da439d6063299dc39feee9cR46

I am OK with both. 

CC @marinaaisa

marinaaisa:
> This is the exact change that broke it. I am not sure whether we should revert that `min-height` to `height` or keep this fix. https://github.com/apple/swift-docc-render/pull/47/files#diff-0e3113fe0f5966efbcbf73dbccb8c910fd794de61da439d6063299dc39feee9cR46
> 
> I am OK with both.
> 
> CC @marinaaisa

I don't have the context of why `height` was changed to `min-height` in the first place. If reverting it doesn't create a regression, I would prefer not adding new CSS. Since you were the one introducing the min-height change, can you tell if there is any regression when reverting it, @dobromir-hristov ?

dobromir-hristov:
Not that i could see. I think it was so that we could have items larger than that height as children without cutoff.

hqhhuang:
> Not that i could see. I think it was so that we could have items larger than that height as children without cutoff.

Do you have an example of that? I could open another PR and keep both `height` and `min-height`?

dobromir-hristov (approved):
I think I added the change, so that the links on the mobile Tutorial nav, are as height as they need to be, based on the amount of children they have. When it used to be a hard `height` it was going into a weird state where the children are visible, but the parent is only 42px.

The fix is OK, I will approve it.

hqhhuang:
@swift-ci test

PR #176, opened 2022-04-15
Convert `left` rule to `padding` rule to fix inset issue

mportiz08:
Bug/issue #, if applicable: 89775264

## Summary

For some reason, there is a CSS `left` property assignment that is inconsistently being evaluated as expected on devices with left inset values.

I've simply updated this to be a `padding` rule, which seems to resolve this issue and the spacing works consistently now.

## Testing

Steps:
1. Open a page with the sidebar on a device that has a left inset when in landscape orientation (like the iPhone 13 Pro as an example).
2. Open the mobile sidebar and verify that the close icon is not obscured by the device.

## Checklist

- [ ] ~Added tests~~


dobromir-hristov (approved):

mportiz08:
@swift-ci test

PR #177, opened 2022-04-16
Update padding in S viewport

hqhhuang:
Bug/issue #, if applicable: rdar://90844182

## Summary
Update padding in S viewport to use variable

## Checklist

- [ ] Added tests - NA

### DIFF
--- a/src/styles/core/_vars.scss
+++ b/src/styles/core/_vars.scss
@@ -67,4 +67,3 @@ $card-horizontal-spacing-small: $card-horizontal-spacing / 2;
 
 // Small vp padding and width
 $small-viewport-content-padding: 20px !default;
-$small-viewport-content-width: 100% !default;
--- a/src/styles/core/_vars.scss
+++ b/src/styles/core/_vars.scss
@@ -64,3 +64,7 @@ $total-columns: 12 !default;
 $card-horizontal-spacing: 10px;
 $card-horizontal-spacing-large: $card-horizontal-spacing * 2;
 $card-horizontal-spacing-small: $card-horizontal-spacing / 2;
+
+// Small vp padding and width
+$small-viewport-content-padding: 20px !default;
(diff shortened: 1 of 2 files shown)


hqhhuang:
@swift-ci test

dobromir-hristov (approved):
LGTM

src/styles/core/_vars.scss
```diff
@@ -64,3 +64,7 @@ $total-columns: 12 !default;
 $card-horizontal-spacing: 10px;
 $card-horizontal-spacing-large: $card-horizontal-spacing * 2;
 $card-horizontal-spacing-small: $card-horizontal-spacing / 2;
+
+// Small vp padding and width
+$small-viewport-content-padding: 20px !default;
+$small-viewport-content-width: 100% !default;
```
mportiz08, line 70: I'm not a CSS expert, so hopefully @dobromir-hristov and/or @marinaaisa can add some better advice on this:

I _think_ it might be slightly odd to have 100% width _and_ the 20px padding combined together (may run into issues depending on the `box-sizing`).

One possible alternative: don't set `$small-viewport-content-width` here—you can define it only when you need to adjust it and then change the code using it conditionally assign it if it's defined.

Another possible alternative: use `auto` instead of 100%. I think this would work but again I think someone else should comment on the best approach.

hqhhuang: This is why I originally wanted to set it to `unset`, but that's also not the best option. Maybe `auto` is a better approach? I set up some time to discuss this PR with Dobri tomorrow.

hqhhuang:
@swift-ci test

hqhhuang:
@swift-ci test

dobromir-hristov (approved):
LGTM

hqhhuang:
@swift-ci test

PR #179, opened 2022-04-16
[Bug Fix] Fix issue where chevron icon in collapsed top navbar is not visible (SR-16120)

sofiaromorales:
Bug/issue #, if applicable: SR-16120

## Summary

In certain viewports, and zoom levels in Safari, when in collapse range, the chevron icon from the top navbar dropdown menu is not visible to the user. The `span nav-menucta-chevron` element still exists in the DOM and can be clicked but the `:before` and `:after` elements are not visible.

## Dependencies

NA

## Testing

1. run `$ VUE_APP_DEV_SERVER_PROXY=/path/to/SlothCreator.doccarchive npm run serve`
2. Visit http://localhost:8080/documentation/slothcreator on Safari
3. Enter `Responsive Design Mode` and set the following values
- `800 x 700` VP
- `1x` res
- Safari 15.3 - macOS - Mac
4. Zoom out to 85% 

[SlothCreator.doccarchive.zip](https://github.com/apple/swift-docc-render/files/8498988/SlothCreator.doccarchive.zip)

(I was unable to upload screenshots showing the bug into the Jira platform due to a 403 Forbidden error)

## Checklist

- [ ] Added tests NA
- [ ] Updated documentation if necessary NA

### DIFF
--- a/src/components/AdjustableSidebarWidth.vue
+++ b/src/components/AdjustableSidebarWidth.vue
@@ -57,8 +57,8 @@ import { baseNavStickyAnchorId } from 'docc-render/constants/nav';
 export const STORAGE_KEY = 'sidebar';
 
 // the maximum width, after which the full-width content does not grow
-export const MAX_WIDTH = 1800;
-export const ULTRA_WIDE_DEFAULT = 500;
+export const MAX_WIDTH = 1920;
+export const ULTRA_WIDE_DEFAULT = 543;
 
 export const eventsMap = {
   touch: {
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -11,7 +11,11 @@
 <template>
   <div class="doc-topic">
     <main class="main" id="main" role="main" tabindex="0">
-      <DocumentationHero :type="role" :enhanceBackground="enhanceBackground">
+      <DocumentationHero
+        :role="role"
+        :enhanceBackground="enhanceBackground"
+        :extraPadding="extraPadding"
+      >
         <template #above-content>
           <slot name="above-hero-content" />
         </template>
@@ -37,7 +41,7 @@
         />
       </DocumentationHero>
       <div v-if="showContainer" class="container">
-        <div class="description">
+        <div class="description" :class="{ 'after-enhanced-hero': enhanceBackground }">
           <RequirementMetadata
             v-if="isRequirement"
             :defaultImplementationsCount="defaultImplementationsCount"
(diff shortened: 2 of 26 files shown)


hqhhuang:
Hey Sofia! Nice catch! I investigated this bug and I think the issue causing the icon to disappear is that we set the maximum height of `.nav-actions`, the container containing the icon, to be 48px. In this scenario, the actual `height` of `.nav-actions` is greater than 48px (see orange box), causing an overflow. I think a better way to fix this is to set `height` to be `inherit` instead. Let me know what you think!
Screen Shot 2022-04-18 at 1 10 26 PM

sofiaromorales:
@hqhhuang Sure, sounds good to me, for what I could notice the `max-height` is only taking effect when the VP is in the medium breakpoint, now, if the parent container `nav-actions` is going to be higher (only when medium breakpoint applies), should I set the original `margin-top: 2px` to only apply on the small breakpoint? that way the chevron icon would stay in place

sofiaromorales:
@hqhhuang I implemented your suggested change and seems to work great, thanks! :)

hqhhuang:
@swift-ci test

dobromir-hristov (changes requested):
The problem is still visible on the Tutorials pages - http://localhost:8080/tutorials/slothcreator/creating-custom-sloths

Removing that margin-top does fix the problem. I think we can do that, and let @SamLanier do a review. We may have added the margin to do some optical alignment, but we can do that other ways if necessary.

sofiaromorales:
@dobromir-hristov Thanks for testing on tutorials! I ended up removing the margin-top and max-height properties, and visually I cannot tell the difference of those 2px off. @SamLanier ready for review 😄

SamLanier:
This solution will not work, since the icon in the nav bar should be aligned with the middle of the text:
mobile-nav-alignment
:

sofiaromorales:
Hi @SamLanier,  you're right, the icon is not aligned properly with the title, but after checking on the master branch it doesn't seems to be aligned on there either, if that is the case the 2px are not the ones causing the misalignment, but probably has to do with the absolute positioning of the icon. I'm happy to work on a fix for that but I'm wondering if that should still be part of this PR considering that is not related to the original bug.

This is a screenshot from the master branch without any of these PR changes applied, if you could confirm the misalignment there would be great.

Screen Shot 2022-04-20 at 11 02 37 PM

dobromir-hristov (approved):
I think its fine now. Visually I dont see much of a diff.

hqhhuang (approved):
Looks good to me! We can file a separate bug to resolve the misalignment. @SamLanier 
Thank you Sofia!

hqhhuang:
@swift-ci test

PR #181, opened 2022-04-18
Cast to lower case the current path in the navigator, to ensure matching paths with capital letters.

dobromir-hristov:
Bug/issue #, if applicable: 91824384

## Summary

Fixes an issue, where if you visit a url that has a capital case letter, it would not find a match if the current page you are on.
It also fixes a small issue with stale activeUID in cache, where it may be null, but we could have actual items to match against.

## Dependencies

NA

## Testing

Steps:
1. Serve any archive
2. Go to a page the archive supports, but change some letters in the URL to capitalcase.

## Checklist

- [x] Added tests


mportiz08 (approved):

dobromir-hristov:
@swift-ci test

PR #182, opened 2022-04-18
Add possible API changes as suggested tags in the sidebar Navigator

dobromir-hristov:
Bug/issue #, if applicable: 91762298

## Summary

This PR adds available API change types as tags to the Navigator. Only types, that yield results will be available to the users to pick.

## Dependencies

NA

## Testing

Not yet testable, rely on Unit Tests

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -111,6 +111,7 @@ import FilterInput from 'docc-render/components/Filter/FilterInput.vue';
 import { BreakpointName } from 'docc-render/utils/breakpoints';
 import keyboardNavigation from 'docc-render/mixins/keyboardNavigation';
 import { isEqual, last } from 'docc-render/utils/arrays';
+import { ChangeNames, ChangeNameToType } from 'docc-render/constants/Changes';
 
 const STORAGE_KEYS = {
   filter: 'navigator.filter',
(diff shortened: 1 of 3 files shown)


src/components/Navigator/NavigatorCard.vue
```diff
@@ -337,8 +352,14 @@ export default {
         // check if `title` matches the pattern, if provided
         const titleMatch = filterPattern ? filterPattern.test(title) : true;
         // check if `type` matches any of the selected tags
-        const tagMatch = selectedTags.length
-          ? tagsSet.has(TOPIC_TYPE_TO_TAG[type]) : true;
+        let tagMatch = true;
+        if (selectedTags.length) {
+          tagMatch = tagsSet.has(TOPIC_TYPE_TO_TAG[type]);
+          // if there are API changes and there is no tag match, try to match change types
+          if (apiChanges && !tagMatch) {
```
hqhhuang, line 359: Why do we only try to match change types when there are API Changes and no tag match? 
I was thinking maybe we should always match both Topic and API Changes? So we can further filter the content down.

hqhhuang (approved):
API Changes are only available for symbol type pages, right? So once a `changeType` tag is selected, we don't need to worry about `Topictype` tags?

dobromir-hristov:
We allow only only one tag to be picked at once. So if you pick `Added` you cannot pick `Article` for example. For now...

dobromir-hristov:
@swift-ci test

PR #187, opened 2022-04-19
Prevent Safari from reading the navigator group markers as clickable

dobromir-hristov:
Bug/issue #, if applicable: 

## Summary

VO on Safari reads any element with an event listener attached as a clickable element.

To prevent that we have to remove the listener, or add a `role`. - https://pauljadam.com/demos/voiceover-clickable.html

We cannot assign `role` on the Reference, because that is an actual prop it has.

That is why I created a simple `OptionalLink` component, that renders a span instead of a `router-link` if `disabled` or no url is passed. This is a very very basic version of the

## Dependencies

NA

## Testing

Steps:
1. Make sure links in the navigator are passing over query params properly
2. Make sure labels are not read out as `clickable` in Safari by VO

## Checklist

- [x] Added tests


src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -67,21 +67,22 @@
         >
           {{ item.index + 1 }} of {{ item.siblingsCount }} symbols inside
         </span>
-        <Reference
```
mportiz08, line 70: Do you know if there was a specific reason we were using `Reference` instead of `router-link` before?

If so, maybe it would be more appropriate to create an `OptionalReference` that uses `Reference` instead of `router-link`

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -67,21 +67,22 @@
...
+        <OptionalLink
```
mportiz08, line 70: Did you consider just using the `isGroupMarker` to conditionally use a span or link here directly instead of creating a new "Optional..." component?

I don't know if it would necessarily be better or worse off the top of my head—just wondering if you had already tried that.

mportiz08, line 70: Did you consider just using the `isGroupMarker` to conditionally use a span or link here directly instead of creating a new "Optional..." component?

I don't know if it would necessarily be better or worse off the top of my head—just wondering if you had already tried that.

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -67,21 +67,22 @@
...
           :id="item.uid"
           :url="item.path || ''"
```
mportiz08, line 72: It looks like the `Reference` component already has logic to utilize a span when it isn't active and/or doesn't have a URL. Are you sure we're not duplicating logic by creating a new component?

mportiz08, line 72: It looks like the `Reference` component already has logic to utilize a span when it isn't active and/or doesn't have a URL. Are you sure we're not duplicating logic by creating a new component?

dobromir-hristov:
@mportiz08 to answer all of your questions at once:

I used `Reference`, so that I can use the inactive link feature (replace with `span`) when the link is not provided or we have a `groupMarker`, along with adding the query params, like the API changes and language params.

The only reason I could not reuse the Reference, is because I cannot set a `role` attribute, because we already have a `role` prop.

If you are OK with me renaming the `role` prop on `Reference` to something else, then I can remote the OptionalLink.

mportiz08:
> @mportiz08 to answer all of your questions at once:
> 
> I used `Reference`, so that I can use the inactive link feature (replace with `span`) when the link is not provided or we have a `groupMarker`, along with adding the query params, like the API changes and language params.
> 
> The only reason I could not reuse the Reference, is because I cannot set a `role` attribute, because we already have a `role` prop.
> 
> If you are OK with me renaming the `role` prop on `Reference` to something else, then I can remote the OptionalLink.

I think that would be fine, as long as you agree that it would better to not create an additional component for this (totally ok if you don't agree).

I'm wondering if we can reduce the risk of renaming an existing prop of a heavily used component though—would it be possible to have it additionally accept an optional `ariaRole` prop that it uses to set the actual `role` HTML attribute wherever that's being set?

dobromir-hristov:
@mportiz08 `ariaRole` sounds good to me. Will try it today.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@mportiz08 great suggestion, much better than my initial work.

src/components/ContentNode/Reference.vue
```diff
@@ -91,6 +91,10 @@ export default {
       type: String,
       required: false,
     },
+    ariaRole: {
+      type: String,
+      default: null,
```
mportiz08, line 96: Minor, non-blocking thing: the `null` default probably isn't needed if this is already an optional prop.

mportiz08 (approved):
The code looks good, but I'm wondering if @marinaaisa or @hqhhuang can help review this as well, because I'm having trouble reproducing the original behavior where these items are being described as "clickable" by VoiceOver, so it's hard to verify that this resolves that.

hqhhuang:
I also can not reproduce the issue. For me, VoiceOver is showing for main and this fix:
 Screen Shot 2022-04-21 at 2 37 13 PM

dobromir-hristov:
@swift-ci test

marinaaisa (approved):
Looks good to me!
Before:
Screenshot 2022-04-22 at 15 16 28

After:
Screenshot 2022-04-22 at 15 17 32

marinaaisa:
> I also can not reproduce the issue. For me, VoiceOver is showing for main and this fix: Screen Shot 2022-04-21 at 2 37 13 PM

Hi Hanqing!
This is happening because you are using the keyboard navigation with VoiceOver instead of VoiceOver navigation.

To use VO navigation, move thought the page using the Interaction commands listed here:
https://support.apple.com/en-gb/guide/voiceover/cpvokys07/mac
Note: The default VO keys are Control-Option

The keyboard navigation will be changed at rdar://91649787

marinaaisa (changes requested):
After thinking a bit more about the role presentation proposed here in this radar, I think group markers should have heading 3 roles and "navigator-head" should be a heading 2. Example of headings inside navigations: https://html.spec.whatwg.org/multipage/sections.html#the-nav-element

Example of headings distributed through sections: https://www.w3.org/WAI/tutorials/page-structure/headings/

Let me know if you have any question!

marinaaisa:
I changed the implementation of this PR, instead of using the role presentation that gives no semantics to the group makers, I set them to be a `h2` and also the title of the nav becomes a `h1`. Having multiple h1s can be posible because this one is inside a nav element.

Screenshot 2022-04-22 at 20 51 03

dobromir-hristov (commented):
It no longer calls the headings as `clickable`, but I am not sure the H1 is good there.

Everything else looks fine to me :))

marinaaisa:
> It no longer calls the headings as clickable, but I am not sure the H1 is good there.

@dobromir-hristov Yep, I was unsure of using another h1 as well... I read some articles that said that it should be fine it's inside another section of the page like the navigation but then these articles from W3 recommends using h2 inside the nav so I'm reverting it. https://www.w3.org/TR/WCAG20-TECHS/H42
https://www.w3.org/WAI/tutorials/page-structure/example/

marinaaisa:
@swift-ci please test

tests/unit/components/Navigator.spec.js
```diff
@@ -112,6 +112,12 @@ describe('Navigator', () => {
   beforeEach(() => {
     jest.clearAllMocks();
   });
+
+  it('renders a nav', () => {
```
dobromir-hristov, line 116: This assertion could have been part of the `it('renders the Navigator'` test. I dont see a reason to have it separate.

dobromir-hristov (commented):
This looks good to me now. Left a small comment on one of the assertions.

dobromir-hristov:
@swift-ci test

tests/unit/components/Navigator.spec.js
marinaaisa: Sounds good :)

marinaaisa (approved):
Approving myself 😅 Thanks Dobri too!

PR #209, opened 2022-04-22
VoiceOver should read button and link in NavigatorCardItem as independent items keeping keyboard navigation as it is

marinaaisa:
Bug/issue #91649787, if applicable: 

## Summary
VoiceOver should read button and link in NavigatorCardItem as independent items keeping keyboard navigation as it is

- [x] Button should have aria-expanded + current aria-describedby without instructions
- [x] Link should have the current aria-describedby with instructions
- [x] Keyboard navigation has to remain the same so we need to fake the focus state of the link to cover the whole item
- [x] tabindex="-1" should be removed from link and button
- [x] tabindex=“0” should be removed from navigator-card-item
- [x] Roving tabindex should be performed on the link, not on navigator-card-item

### Before

Screenshot 2022-04-22 at 21 11 14

### After
Screenshot 2022-04-22 at 21 10 47
Screenshot 2022-04-22 at 21 10 57

## Dependencies

NA

## Testing

Steps:
1. Open a .doccarchive with navigator
2. Turn on VoiceOver and navigate using the VO Interaction commands listed here:
https://support.apple.com/en-gb/guide/voiceover/cpvokys07/mac
Note: The default VO keys are Control-Option
3. Assert that keyboard navigation has not changed
4. Assert that VO navigation has changed and button and link are separated

## Checklist

- [x] Added tests


src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -72,8 +70,10 @@
           :url="item.path || ''"
           :isActive="!isGroupMarker"
           :class="{ bolded: isBold }"
+          :tabindex="isFocused ? '0' : '-1'"
+          :aria-describedby="`${ariaDescribedBy} ${usageLabel}`"
+          :aria-hidden="isRendered ? null : 'true'"
```
dobromir-hristov, line 75: We should probably keep the aria-hidden on the parent. This is a failsafe in case you try to focus something that is used to render content.

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -241,14 +241,14 @@ $depth-spacer-base-spacing: (
 $nesting-spacing: $card-horizontal-spacing + $card-horizontal-spacing-small;
 
 .navigator-card-item {
-  overflow: hidden;
   height: $item-height;
   display: flex;
   align-items: center;
 
-  @include on-keyboard-focus {
+  &:focus-within {
```
dobromir-hristov, line 248: This is a bit risky, as it got a wide spread adoption recently, but its OK on Safari based devices - https://caniuse.com/?search=focus-within

dobromir-hristov (approved):
Looking good to me.

Left two small none-blockers, that we should probably look into.

src/components/Navigator/NavigatorCardItem.vue
marinaaisa: Moved :) https://github.com/apple/swift-docc-render/pull/209/commits/fd0a2b94c3d4094ce4c33a2c1a514afd55e6bd89

src/components/Navigator/NavigatorCardItem.vue
marinaaisa: I'm not very worried about its browser compatibility which it's similar to the one CSS Grid has.

dobromir-hristov (approved):
LGTM now. Maybe a designer should look into it too.

PR #215, opened 2022-04-25
Don't render default footer for IDE target

mportiz08:
Bug/issue #, if applicable: 89998805

## Summary

Fixes an issue where the default footer is incorrectly rendered for IDE targets in certain edge case scenarios.

Right now, this is happening when navigating very quickly between pages where occasionally there is an error encountered when attempting to load data, which is handled by calling `next(false)` in a vue router guard to abort the navigation—when this happens, a blank page is loaded, but the footer is still shown unexpectedly.

_Note: marking this as a draft PR for the moment, since I am still doing some of my own manual testing to verify that these changes resolve the issue without introducing any new regressions._

## Testing

Steps:
1. Run `VUE_APP_TARGET=ide npm run serve`
2. Open http://localhost:8080/documentation/foobar and verify that a blank page is shown with no color scheme toggle footer 
3. Run `VUE_APP_DEV_SERVER_PROXY=https://ethan-kusters.github.io/swift-argument-parser npm run serve`
4. Open http://localhost:8080/documentation/argumentparser and verify that the color scheme toggle footer is still shown
5. Open http://localhost:8080/documentation/foobar and verify that the color scheme toggle is still shown

## Checklist

- [X] Added tests


src/styles/base/_typography.scss
```diff
@@ -21,7 +21,7 @@ html {
 
 body {
   @include font-styles(body);
-  background-color: var(--color-fill);
+  background-color: var(--color-text-background);
```
hqhhuang, line 24: Why is this change necessary?

mportiz08: It's slightly more correct in that we want the text and text-background colors as the default body color for the color/background-color respectively. Technically in swift-docc-render, these are the same (fill is used for text-background), but other applications extending this may choose to have different semantic colors for the same thing.

dobromir-hristov:
Interesting 🤔

hqhhuang (approved):
Spent more time to test this a little more carefully and the change looks good!

dobromir-hristov (approved):

mportiz08:
@swift-ci test

PR #221, opened 2022-04-27
Fix bug in Safari, where the navigator cannot be scrolled by dragging scrollbar

dobromir-hristov:
Bug/issue #, if applicable: 92337769

## Summary

Navigator cannot be scrolled, by dragging the scrollbar puck in Safari. This PR fixes that by adding a one line CSS fix, forcing the browser to do some stacking changes. 

## Dependencies

NA

## Testing

Steps:
1. Open enough items in the navigator, to cause a scroll
2. Try to click and drag the scrollbar using the mouse or touchpad in Safari

## Checklist

- [ ] Added tests - no need, CSS only


marinaaisa (approved):
Works well! 👍

src/components/Navigator/NavigatorCard.vue
```diff
@@ -1121,6 +1121,10 @@ $navigator-head-background-active: var(--color-fill-tertiary) !default;
   @include breakpoint(medium, nav) {
     padding-bottom: $nav-menu-items-ios-bottom-spacing;
   }
+  // hack to allow scrolling on Safari
```
mportiz08, line 1124: Can you explain the "hack" better in this comment? I think this will be very confusing to anyone else reading this in the future since it's not clear why we're doing the translation without a different position. Your PR mentions that this affects the stacking context for Safari—it would be good to describe the specific problem in this comment and how this actually resolves that. Thanks!

dobromir-hristov: Here is where I got the fix from - https://github.com/Akryum/vue-virtual-scroller/issues/387#issuecomment-833304471

dobromir-hristov:
@swift-ci test

PR #230, opened 2022-04-28
Improve legibility of typography elements for base Helvetica font

dobromir-hristov:
Bug/issue #, if applicable: 86116620

## Summary

Improve legibility of typography elements for base Helvetica font

## Dependencies

NA

## Testing

Steps:
1. Assert fonts look better, without visual regressions

## Checklist

- [ ] Added tests


SamLanier (approved):
Design approved.

src/styles/core/typography/_font-attribute-utils.scss
```diff
@@ -53,6 +53,10 @@
   @return $line-height / $size;
 }
 
+@function -transform-letter-spacing($letter-spacing) {
+  @return $letter-spacing * 0;
```
mportiz08, line 57: Hm. This technically works for what you're trying to accomplish, but it seems like a pretty roundabout approach. It seems a little strange to be configuring all these specific letter-spacing values for different types of typography styles in this codebase and then purposefully zeroing them out here and never actually using them.

We should probably consider removing the `letter-spacing` attribute from all of the `$font-attributes` items and re-consider a cleaner way of allowing these to be added/updated by other applications or themes.

What do you think?

dobromir-hristov: Right, I was looking for the fastest and easiest approach. What you are proposing is better, but does introduce more risk for third party themes, as they would now have to re-declare all those font-attributes.

I am not saying I am opposed that, I just want to get a second opinion if that's an OK risk to take.

mportiz08:
@swift-ci test

mportiz08:
@swift-ci test

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

PR #234, opened 2022-04-29
Use the `hierarchy.path`, closest to the current page URL

dobromir-hristov:
Bug/issue #, if applicable: 92105815

## Summary

Ensures we show breadcrumbs that are as close to the current page as possible, in cases where one page is linked from multiple places.

## Dependencies

NA

## Testing

Steps:
1. Make sure no regressions in normal doccarchives

## Checklist

- [x] Added tests


src/views/DocumentationTopic.vue
```diff
@@ -196,8 +196,20 @@ export default {
...
-    parentTopicIdentifiers: ({ topicProps: { hierarchy: { paths: [ids = []] = [] } } }) => ids,
+    // one, that has the same path as the current URL.
+    parentTopicIdentifiers: ({ topicProps: { hierarchy: { paths = [] }, references }, $route }) => {
+      if (!paths.length) return [];
+      return paths.find((iteratedPaths) => {
+        // ge the first item
+        let pathToUse = references[iteratedPaths[0]];
+        // if its a technology, get the next one
+        if (pathToUse && pathToUse.kind === 'technologies') {
+          pathToUse = references[iteratedPaths[1]];
```
mportiz08, line 207: This technically works because of the fact that the array always has at least 2 items in this situation in practice with real data, but the `pathToUse` logic seems a little odd to me because of that assumption.

Could we simplify it to something slightly more general and safe like this pseudo code?
```
const rootPath = iteratedPaths.find( .. not-technology-kind-ref... )
```

mportiz08 (approved):
Seems to be working well.

I have one minor comment about the code.

src/views/DocumentationTopic.vue
dobromir-hristov: Why is it bad? We just pick the first item and in most cases it will be the one used anyway. In the off-chance its a `technology` kind, we just use the next one, if available. If the second item is empty, we skip the entire iteration and continue with the next one (which shouldnt happen I think).

I guess a `find` solution would be:

```js
    parentTopicIdentifiers: ({ topicProps: { hierarchy: { paths = [] }, references }, $route }) => {
      if (!paths.length) return [];
      return paths.find((iteratedPaths) => {
        const rootPath = iteratedPaths.find(item => references[item] && references[item].kind !== 'technologies');
        // if there is an item, check if the current url starts with it
        return rootPath && $route.path.toLowerCase().startsWith(rootPath.url.toLowerCase());
      }) || paths[0];
    },
```

mportiz08: Not saying that it's "bad", just potentially unsafe in that it relies on the length of certain arrays without doing checks. I think what you have might actually be fine since that will just return undefined in JavaScript. I was probably thinking of other languages that crash at runtime when that happens, sorry. It shouldn't happen either way with the data that we expect.

The `find` solution seems very slightly safer to me for that reason and is also a little easier to read as a bonus, although that's more subjective.

mportiz08: Either way works.

dobromir-hristov: Applied the find :)

dobromir-hristov:
@swift-ci test

PR #235, opened 2022-04-29
Persist navigator data across technologies

dobromir-hristov:
Bug/issue #, if applicable: 91188362

## Summary

If there is an item in the navigator, that links to another technology/documentation, going back should still re-apply the old filters.

## Dependencies

https://github.com/apple/swift-docc-render/pull/234

## Testing

Steps:
1. Assert there are no regressions when working with normal doccarchives.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -63,7 +63,7 @@
               @toggle="toggle"
               @toggle-full="toggleFullTree"
               @toggle-siblings="toggleSiblings"
-              @navigate="setActiveUID"
+              @navigate="handleNavigationChange"
               @focus-parent="focusNodeParent"
             />
           </RecycleScroller>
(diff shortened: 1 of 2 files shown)


src/components/Navigator/NavigatorCard.vue
```diff
@@ -835,7 +859,16 @@ export default {
     setActiveUID(uid) {
       this.activeUID = uid;
       this.resetScroll = false;
-      sessionStorage.set(STORAGE_KEYS.activeUID, uid);
+    },
+    /**
+     * Handles the `navigate` event from NavigatorCardItem, guarding from selecting an item,
+     * that points to another technology.
+     */
+    handleNavigationChange(uid) {
```
marinaaisa, line 867: If a user doesn't navigate through the technology (they go straight away to a new different technology instead by just filtering) we never set the activeUID and therefore we cannot persist the filtering data :(

We need a solution also for those cases.

marinaaisa (changes requested):
Left couple of comments. One is about a use case that it's still not persisting the data but maybe that's why the PR is still on draft. :)

src/components/Navigator/NavigatorCard.vue
dobromir-hristov: It was called, but I had some mistakes in my logic. Should be fixed now.

dobromir-hristov (commented):
Try again, should be fixed now.

marinaaisa (approved):
Testing works great and code too. Let me know again to give another look when you finish the PR and open it (if it's not finished yet)

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #244, opened 2022-05-02
Use spinner icon to indicate indefinite progress while loading navigator

mportiz08:
Bug/issue #, if applicable: 92438279

## Summary

This updates the navigator UI to use a spinner icon to indicate indefinite loading progress instead of the **"Fetching..."** text. \cc @biamx3

spinner

## Testing

You can use the following patch to simulate a scenario where the navigator is taking infinitely long to fetch index data:

```diff
diff --git a/src/views/DocumentationTopic.vue b/src/views/DocumentationTopic.vue
index 4770b61..c22020b 100644
--- a/src/views/DocumentationTopic.vue
+++ b/src/views/DocumentationTopic.vue
@@ -42,7 +42,7 @@
                 <Navigator
                   :parent-topic-identifiers="parentTopicIdentifiers"
                   :technology="slotProps.technology || technology"
-                  :is-fetching="slotProps.isFetching"
+                  :is-fetching="true"
                   :error-fetching="slotProps.errorFetching"
                   :api-changes="slotProps.apiChanges"
                   :references="topicProps.references"

```

Steps:
1. Copy the patch above and apply it with `pbpaste | git apply`
2. Run the dev server with example data using `env VUE_APP_DEV_SERVER_PROXY=https://ethan-kusters.github.io/swift-argument-parser npm run serve`
3. Open an example URL like http://localhost:8080/documentation/argumentparser and verify that the navigator shows the spinner icon instead of the text **Fetching...**

## Checklist

- [X] Added tests


dobromir-hristov (approved):
LGTM

src/components/Navigator.vue
```diff
@@ -204,7 +206,18 @@ export default {
...
+  padding: $section-spacing-single-side 0;
+  text-align: center;
+  width: 100%;
+}
+
+.loading-spinner {
+  --spinner-size: 40px;
+
+  height: var(--spinner-size);
+  position: sticky;
```
marinaaisa, line 219: Why using position sticky instead of just centering with flex? Is there any use case that I'm missing?

biamx3:
It looks great! Can we add a small delay so the indicator doesn't appear unless the page doesn't load instantaneously/near instantaneously?

src/components/Navigator.vue
mportiz08: I was trying to think of a simpler approach like that at first, but I couldn't think about how to make it work right since it needs to vertically center in the sidebar, which may expand beyond the viewport. I'll try to see if I can come up with something simpler like that and make it work.

mportiz08:
> It looks great! Can we add a small delay so the indicator doesn't appear unless the page doesn't load instantaneously/near instantaneously?

I can look into that, yeah. There may still be instances where the page happens to load just after the time we use for the small delay though which would then have the same visual problem, since the exact load time is not something we can predict.

src/components/Navigator.vue
mportiz08: Yeah, I'm not sure how to do this with a flex layout. The trouble happens when you scroll down the page, we want to still show the loading indicator in the center of the sidebar instead of having it scroll out of view.

Let me know if you think of a better way of doing that though please!

mportiz08:
> It looks great! Can we add a small delay so the indicator doesn't appear unless the page doesn't load instantaneously/near instantaneously?

Added a delay so that the spinner doesn't show unless it has taken at least a second to load the data (and it hasn't loaded yet).

dobromir-hristov (approved):

src/components/Navigator.vue
marinaaisa: Alternatively, you could do:

```css
.loading-placeholder {
  color: var(--color-figure-gray-secondary);
  position: sticky;
  top: $nav-height;
  height: calc(100vh - #{$nav-height});
  width: 100%;
  display: flex;
  align-items: center;
  justify-content: center;

  @include breakpoint(medium, nav) {
    position: static;
    height: 100%;
  }
}

.loading-spinner {
  --spinner-size: 40px;

  height: var(--spinner-size);
  width: var(--spinner-size);
}
```

just as we do for class `.navigator-card-inner`
and extract the common CSS into a SCSS mixin.

It implies to use position sticky but at least you don't have to center the position with `top`
As you prefer :)

I'm approving the PR.

marinaaisa (approved):

mportiz08:
> Alternatively, you could do:
> 
> ```css
> .loading-placeholder {
>   color: var(--color-figure-gray-secondary);
>   position: sticky;
>   top: $nav-height;
>   height: calc(100vh - #{$nav-height});
>   width: 100%;
>   display: flex;
>   align-items: center;
>   justify-content: center;
> 
>   @include breakpoint(medium, nav) {
>     position: static;
>     height: 100%;
>   }
> }
> 
> .loading-spinner {
>   --spinner-size: 40px;
> 
>   height: var(--spinner-size);
>   width: var(--spinner-size);
> }
> ```
> 
> just as we do for class `.navigator-card-inner`
> and extract the common CSS into a SCSS mixin.
> 
> It implies to use position sticky but at least you don't have to center the position with `top`
> As you prefer :)
> 
> I'm approving the PR.

Sounds like a good idea to me—I extracted the shared styles into a `NavigatorCardInner` component with bfa060241aaefc6b8c0a79256742028e84f06ef4 to address this.

dobromir-hristov (approved):
I saw it leaves some extra space for the filter, which we probably dont need for the loading screen? Apart from that, refactors looks good.

image

src/components/Navigator/NavigatorCardInner.vue
```diff
@@ -0,0 +1,35 @@
...
+
+<style scoped lang="scss">
+@import 'docc-render/styles/_core.scss';
+
+$filter-height: 71px;
+
+.navigator-card-inner {
+  position: sticky;
+  top: $nav-height;
+  height: calc(100vh - #{$nav-height} - #{$filter-height});
```
marinaaisa, line 27: We shouldn't rest the `filter-height` when the navigator component is loading. This is why the UI doesn't reach the bottom of the page on its loading state. @dobromir-hristov pointed it out before.

marinaaisa (changes requested):

src/components/Navigator/NavigatorCardInner.vue
mportiz08: 353c1dfd30aafdb57b02f74bb710b17f95403f67

src/components/Navigator.vue
```diff
@@ -203,8 +209,22 @@ export default {
...
-  @include font-styles(body-reduced);
+  justify-content: center;
+}
+
+.loading-spinner {
+  --spinner-size: 40px; // used for both width and height
+  --spinner-delay: 1s; // don't show spinner until this much time has passed
+
+  height: var(--spinner-size);
+  visibility: visible;
```
marinaaisa, line 222: `visibility: visible;` is the default value. I don't think we need to specify it, right?

mportiz08: https://github.com/apple/swift-docc-render/pull/244/commits/fdbea73c9a062e2ae868f1426bd750d199f6c6c4

mportiz08:
@swift-ci test

marinaaisa (approved):
The last thing, we should delete `padding: 40px 0;` from `.loading-placeholder` class since we are not using it anymore.

After that everything seems good 💯  Thanks for working on it, Marcus! 👏

mportiz08:
I don't believe it has any padding anymore.

PR #266, opened 2022-05-06
Adjust min-height value to match the lowest height to support WatchOS' screen ratio on the narrowest client

marinaaisa:
Bug/issue #76652132, if applicable: 

## Summary

Adjust min-height value to match the lowest height to support WatchOS' screen ratio on the narrowest client

### Before
Screenshot 2022-05-06 at 18 29 03

### After
Screenshot 2022-05-06 at 18 28 45

## Dependencies

NA

## Testing

Steps:
1. Run a .doccarchive with tutorials containing screenshots of watchOS
2. In the narrowest client (320px width) assert that previews containing screenshots of watchOS look correct

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Tutorial/MobileCodePreview.vue
+++ b/src/components/Tutorial/MobileCodePreview.vue
@@ -132,7 +132,9 @@ $-preview-padding: 60px;
   /deep/ img:not(.file-icon) {
     border-radius: $border-radius;
     box-shadow: 0 0 3px rgba(0, 0, 0, 0.4);
-    min-height: 320px;
+    // Min-height matches the lowest height to support
+    // WatchOS' screen ratio on the narrowest client: 320px width
+    min-height: 243px;
     max-height: 80vh;
     width: auto;
     display: block;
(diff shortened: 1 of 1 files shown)


src/components/Tutorial/MobileCodePreview.vue
```diff
@@ -132,7 +132,9 @@ $-preview-padding: 60px;
   /deep/ img:not(.file-icon) {
     border-radius: $border-radius;
     box-shadow: 0 0 3px rgba(0, 0, 0, 0.4);
-    min-height: 320px;
+    // Min-height matches the lowest height to support
+    // WatchOS' screen ratio on the narrowest client: 320px width
+    min-height: 243px;
```
mportiz08, line 137: I don't totally understand where the `243px` value is coming from or if it's just a small enough value to work for a particular asset.

Either way...do we actually need to set a `min-height` value here at all? I know you didn't add it, but does it maybe make sense to just remove it entirely? (Not sure myself, just thinking out loud)

dobromir-hristov: Yeah, 243 is very specific 🤣 Removing the min-height doesnt seem to break anything from a first glance 🤔

mportiz08 (approved):

PR #285, opened 2022-05-10
Make sure ParametersTable changes dont go outside of bounds

dobromir-hristov:
Bug/issue #, if applicable: 92961798

## Summary

ParameterTable changes should not go outside of bounds.

## Dependencies

NA

## Testing

Not testable in current version.

## Checklist

- [ ] Added tests


marinaaisa (approved):
Tested and works great.

Using `display: block;` instead `display: inline-block;` would work too instead of adding `max-width` but I don't know if it would introduce a regression in other use cases.

dobromir-hristov:
@swift-ci test

PR #291, opened 2022-05-11
Tags and text should scroll as a unique element in filter input

marinaaisa:
Bug/issue #90896758, if applicable: 

## Summary

Tags and text should scroll as a unique element in filter input

## Dependencies

NA

## Testing

Steps:
1. Open a .doccarchive that has navigation
2. Select a tag and write a long string in the filter input
3. Assert that when scrolling, tags and input scroll along the way. Before only the text was scrolling.
4. Assert that AX is correct

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Filter/FilterInput.vue
+++ b/src/components/Filter/FilterInput.vue
@@ -553,13 +553,14 @@ $input-height: rem(28px);
     width: 100%;
     height: var(--input-height);
     padding: var(--input-vertical-padding) 0;
-    min-width: 130px; // set a min width, so user can select the area
 
     &::after {
       content: attr(data-value);
       visibility: hidden;
       width: auto;
       white-space: nowrap;
+      min-width: 130px; // set a min width, so user can select the area
+      display: block;
     }
   }
 
--- a/src/components/Filter/FilterInput.vue
+++ b/src/components/Filter/FilterInput.vue
@@ -553,6 +553,7 @@ $input-height: rem(28px);
     width: 100%;
     height: var(--input-height);
     padding: var(--input-vertical-padding) 0;
+    min-width: 130px; // set a min width, so user can select the area
 
     &::after {
       content: attr(data-value);
@@ -577,7 +578,6 @@ $input-height: rem(28px);
     height: var(--input-height);
     border: none;
     width: 100%;
-    min-width: 130px; // set a min width, so user can select the area
     position: absolute;
     background: transparent;
     z-index: 1;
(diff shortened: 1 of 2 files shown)


src/components/Filter/FilterInput.vue
```diff
@@ -56,36 +56,36 @@
           <label
             id="filter-label"
             :for="FilterInputId"
-            class="visuallyhidden"
-            aria-hidden="true"
+            :input-value="modelValue"
```
dobromir-hristov, line 59: I would suggest the same, as I did for @hqhhuang. Use a `data-` attribute, so note its a custom data attribute. Now it looks like a vue prop.

src/components/Filter/FilterInput.vue
```diff
@@ -56,36 +56,36 @@
           <label
             id="filter-label"
             :for="FilterInputId"
-            class="visuallyhidden"
-            aria-hidden="true"
```
dobromir-hristov, line 60: Does removing `aria-hidden` cause it to be read out now?

dobromir-hristov (commented):
LGTM, just left a few small comments.

src/components/Filter/FilterInput.vue
marinaaisa: https://github.com/apple/swift-docc-render/pull/291/commits/34c3848b2287a7b62d7639a4bd1ff8df5614ec9a

src/components/Filter/FilterInput.vue
```diff
@@ -563,13 +577,12 @@ $input-height: rem(28px);
     height: var(--input-height);
     border: none;
     width: 100%;
-    min-width: 130px; // set a min width, so it does not get crushed by tags
+    min-width: 130px; // set a min width, so user can select the area
+    position: absolute;
```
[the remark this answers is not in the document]
marinaaisa: Not really need because the after has `visibility: hidden;`

src/components/Filter/FilterInput.vue
```diff
@@ -563,13 +577,12 @@ $input-height: rem(28px);
     height: var(--input-height);
     border: none;
     width: 100%;
-    min-width: 130px; // set a min width, so it does not get crushed by tags
+    min-width: 130px; // set a min width, so user can select the area
```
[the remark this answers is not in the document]
marinaaisa: Yep, I moved it to the filter label instead. It's not needed in the input because it's doing the 100% of the label now.

marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

dobromir-hristov (approved):
I think we have a fix :)

dobromir-hristov:
@swift-ci test

PR #292, opened 2022-05-11
Calculate image sizes when not provided by Render JSON

mportiz08:
Bug/issue #, if applicable: 88195563

## Summary

This adds logic to calculate the optimal width for rendering an image when the size information has not been provided in the Render JSON.

Since DocC does not provide image size information, there may be a situation where a user only provides a 2x version of an image—on 1x displays, this image may end up looking twice as big as was intended. By calculating the appropriate width in the renderer, it can be shown at the expected size regardless of the pixel density of the display.

## Testing

Steps:
1. Download/unzip [FakePackage.doccarchive.zip](https://github.com/apple/swift-docc-render/files/8674002/FakePackage.doccarchive.zip)
2. Run the dev server with `env VUE_APP_DEV_SERVER_PROXY=/path/to/unzipped/FakePackage.doccarchive npm run serve`
3. Open http://localhost:8080/documentation/fakepackage
4. Using real displays and/or responsive design modes, verify that the image is always rendered at 50 pixels wide, regardless of whether or not the display/tool is in 1x or 2x mode
5. Test other existing content in realistic scenarios if possible and verify that there are no regressions with sizing of images when the Render JSON already includes size information

## Checklist

- [X] Added tests


dobromir-hristov:
I am noticing issues on my 1x screen in Chrome. I think Chrome is smart enough to pick out images with `@2x` in the name, and adjusts the size accordingly.

I modified the archive to adda another image, just changed the name to `example.png`, nothin else and it renders it 50x50, the original one with `@2x` is rendered 25x25. When I inspect the naturalWidth its 50, not 100 like on other browsers.

image

dobromir-hristov:
nvm, it does it again if I refresh, even with a `-2x` in the name instead.

image

dobromir-hristov:
I think in general Chrome is trying to be smart and scale them images down

image

mportiz08:
I did some testing and it seems that you're right @dobromir-hristov — it seems like Safari/Firefox both treat the `src` fallback as a 1x image always whereas Chrome tries to be smart and treat it as a 2x image because of the filename, so this workaround does not work in that case since the return value of `naturalWidth` differs between browsers.

After some thinking, I'm thinking that instead of the changes here, the best solution might actually be to remove the `src` fallback entirely and rely only on `srcset` — that way, we would get the same behavior in all browsers with no workaround needed to downscale images for 1x devices. The only caveat with that is that we would be dropping support for any older browsers that don't support `srcset` yet, and the image wouldn't even render in those browsers. The compatibility for `srcset` does seem pretty well supported, but I'm not sure yet if that's the right move.

(Alternatively we could potentially just enforce that a 1x version of images needs to be provided in addition if 2x images are provided, but that seems like it would be too strict.)

I don't like the idea of trying to add logic to this PR to detect if the browser is Chrome and avoid doing this logic there—seems too fragile.

dobromir-hristov:
Cant we just drop `srcset` if there are no size hints, and only one image is provided? We already have different tags for color variants. We would use `srcset` only in case we have an 2x and 1x variants of an image?

mportiz08:
> Cant we just drop `srcset` if there are no size hints, and only one image is provided? We already have different tags for color variants. We would use `srcset` only in case we have an 2x and 1x variants of an image?

Maybe but that doesn't really address any problem that we have today since that scenario already works fine today (only 1 1x image provided).

The issue I would like to resolve is a common scenario where we only have one 2x resolution image provided, and the image appears twice as big on 1x devices in non-Chrome browsers. Without being provided image dimensions beforehand, `srcset` is needed to ensure the image is rendered at the appropriate size with higher res images.

dobromir-hristov:
if we drop the `srcset` in cases, where its just one image, wont your calculation with `.naturalWidth` work though?

mportiz08:
I think maybe the following might work as the best option to reduce the visibility of this bug? I don't know if there's a way to eliminate it entirely without the server providing image sizes based on what we found here.

* Add JS logic to detect if the browser supports the `srcset` attribute
  * If yes, *only* set `srcset` and use `naturalWidth` for width (should always report the right size in all browsers and no additional calculation is needed since the device pixel density is accounted for)
  * If no, *only* set `src` (bug with image sizes still happens but much less often since `srcset` compatibility is pretty good)

mportiz08:
Actually, I may have found an alternate method of calculating the actual image size without using `naturalWidth` that might be working more reliably across browsers—still need to do more testing though: 08ff27a5281f7a4ea82c19d4337a190f3a0c07df

dobromir-hristov:
GJ Marcus, i saw this when I did my investigation, but was not sure whether it would work as expected.

Tried on 1x and 2x screen, on Chrome and looks great. The image is cached, so it's not downloaded again from the server.

dobromir-hristov:
@swift-ci test

dobromir-hristov (approved):
I can see the flash on 1x screens in Safari, as you mentioned. Chrome manages to prevent that, which is also good.

I suppose we cant hide the flashing/jumping, while JS fetches the image, because we dont know what space to give it. 

We could put the `src` in the DOM, only after we know the size of the image. That should prevent the flashing of big images, but not sure how it will be with caching and all.

mportiz08:
> I can see the flash on 1x screens in Safari, as you mentioned. Chrome manages to prevent that, which is also good.
> 
> I suppose we cant hide the flashing/jumping, while JS fetches the image, because we dont know what space to give it.
> 
> We could put the `src` in the DOM, only after we know the size of the image. That should prevent the flashing of big images, but not sure how it will be with caching and all.

I would be a little uneasy making that change since the `src` attribute is required for the `` tag, although I guess it is technically possible. There would still be a jump as the space goes from nothing to the actual size, but it would be less noticeable than the flash going from a larger image to a smaller one I suppose.

At the end of the day, I still consider this more of a workaround for the real issue where DocC should be providing the image size when possible so that it doesn't need to be calculated client-side. Also, this will only happen on 1x devices when a 1x alternative for the image hasn't been provided.

mportiz08:
@swift-ci test

PR #294, opened 2022-05-12
Ensure navigator unlocks scrolling when navigating away

### LINKED ISSUE #293: Bug - View not scrollable when accessing from sidebar navigator
**Describe the bug**

When using the full-width sidebar navigator, if an item that redirects to a new route different from the current one gets selected, the sidebar gets dismissed, but the scrolling inside the selected page doesn't work.

**Expected behavior**

When an item in the full-width sidebar navigator gets selected, the sidebar should get dismissed, and the view should be scrollable if the body content is higher than the VP.

**Steps to reproduce**

1. Open a `.doccarchive` with sidebar navigator enabled and a tutorials page
2. In Safari, select Develop -> Open Responsive Design Mode  and select an iPhone size (or any other VP in collapse range) 
3. Open the sidebar navigator
4. Select `Meet SlothCreator` or any other tutorial item
5. Once on the tutorials page, try to scroll the page

**Related file**

In `src/components/AdjustableSidebarWidth.vue` there's a method that locks/unlocks the view scroll adding some fixed values to the `body` element.

```js
// src/components/AdjustableSidebarWidth.vue
toggleScrollLock(lock) {
     ...
      if (!scrollLockContainer) return;
      if (lock) {
        scrollLock.lockScroll(scrollLockContainer);
        ...
      } else {
        scrollLock.unlockScroll(scrollLockContainer);
        ...
      }
}
```  
On a relative path change, the scroll gets unlocked, but since the tutorial items redirect to a new route, with path `/tutorials/`, the `AdjustableSidebarWidth` component is removed from the DOM and the `scrollLockID` element id is not found, keeping the lock on.
 
```js
// src/components/AdjustableSidebarWidth.vue
this.$once('hook:beforeDestroy', () => {
      ...
      if (this.openExternally) {
        this.toggleScrollLock(false);
      }
});
```

A possible solution could be to not require `scrollLockID` to unlock the scroll.

**Environment**

Safari 15.3
Testing locally using `swift-docc-render` repo with HEAD in commit id 47b17909e440161619de31d369eeb5dfa0c55ff6 

**Attachment**

Screen Shot 2022-05-11 at 11 45 28 PM

dobromir-hristov:
Bug/issue #, if applicable:  closes #293

## Summary

Fixes a bug where the navigator would not unlock scrolling on mobile devices.

## Dependencies

NA

## Testing

You need an archive with Tutorials, like sloth creator

Steps:
1. From a doc page open the navigator on a mobile device
2. Click on a link, that leads to a tutorial
3. Assert scrolling is possible again

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/AdjustableSidebarWidth.vue
+++ b/src/components/AdjustableSidebarWidth.vue
@@ -280,7 +280,6 @@ export default {
      */
     toggleScrollLock(lock) {
       const scrollLockContainer = document.getElementById(this.scrollLockID);
-      if (!scrollLockContainer) return;
       if (lock) {
         scrollLock.lockScroll(scrollLockContainer);
         // lock focus
--- a/src/utils/scroll-lock.js
+++ b/src/utils/scroll-lock.js
@@ -131,7 +131,7 @@ export default {
     if (!isIosDevice()) {
       simpleLock();
     } else {
-      advancedLock(targetElement);
+      advancedLock(targetElement || {});
     }
     isLocked = true;
   },
@@ -144,7 +144,7 @@ export default {
 
     if (isIosDevice()) {
       // revert the old scroll position
-      advancedUnlock(targetElement);
+      advancedUnlock(targetElement || {});
     } else {
       // remove all inline styles
       document.body.style.cssText = '';
--- a/src/utils/scroll-lock.js
+++ b/src/utils/scroll-lock.js
@@ -63,8 +63,10 @@ function simpleLock() {
 function advancedUnlock(targetElement) {
   /* eslint-disable no-param-reassign */
   // remove the touch listeners on the target
-  targetElement.ontouchstart = null;
-  targetElement.ontouchmove = null;
+  if (targetElement) {
+    targetElement.ontouchstart = null;
+    targetElement.ontouchmove = null;
+  }
   // remove the body event listener
   document.removeEventListener('touchmove', preventDefault);
 }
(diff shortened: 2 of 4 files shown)


dobromir-hristov:
@swift-ci test

src/utils/scroll-lock.js
```diff
@@ -131,7 +131,7 @@ export default {
     if (!isIosDevice()) {
       simpleLock();
     } else {
-      advancedLock(targetElement);
```
mportiz08, line 134: I may be misunderstanding what's happening here, but if `targetElement` may be null/undefined, shouldn't we just change the `else` to `else if (targetElement)` and/or fix the places that might be calling `lockScroll` without an element?

It seems a little odd to me that we would even try calling the method without an element.

src/components/AdjustableSidebarWidth.vue
```diff
@@ -280,7 +280,6 @@ export default {
      */
     toggleScrollLock(lock) {
       const scrollLockContainer = document.getElementById(this.scrollLockID);
-      if (!scrollLockContainer) return;
```
mportiz08, line 283: This is being removed because we are always guaranteed to have rendered an element with this ID, correct?

dobromir-hristov: So in this case, the component is already gone from the DOM, by the time the `AdjustableSidebarWidth` is destroyed, so we were not calling the unlock scripts.

src/utils/scroll-lock.js
dobromir-hristov: Right, but if we locked previously and try to unlock, even if the lock target is no longer in the DOM, we wont be able to. But we still want to clean up the rest (there are some even listeners attached to the body).

What I can do is just inside the `advancedLock` check if `!targetElement` and apply code conditionally.

mportiz08 (approved):
Thanks for catching this bug @sofiaromorales!

dobromir-hristov:
@swift-ci test

PR #296, opened 2022-05-12
[AX] Keyboard navigation shouldn't tag on hidden items in the navigator

marinaaisa:
Bug/issue #91660503, if applicable: 

## Summary

VO should not read hidden items in the navigator

## Dependencies

NA

## Testing

Steps:
1. Open a .doccarchive with navigator
2. In a small or medium viewport, close the navigator
3. Assert that when tabbing or navigating with VO, the navigator isn't rendered

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator.vue
+++ b/src/components/Navigator.vue
@@ -9,34 +9,29 @@
 -->
 
 <template>
-  <transition name="delay-hiding">
-    <nav
-      v-show="showNavigator"
-      class="navigator"
-    >
-      <NavigatorCard
-        v-if="!isFetching"
-        :technology="technology.title"
-        :technology-path="technology.path || technology.url"
-        :type="type"
-        :children="flatChildren"
-        :active-path="activePath"
-        :scrollLockID="scrollLockID"
-        :error-fetching="errorFetching"
-        :breakpoint="breakpoint"
-        :api-changes="apiChanges"
-        @close="$emit('close')"
-      />
-      <NavigatorCardInner v-else class="loading-placeholder">
-        <transition name="delay-showing" appear>
-          <SpinnerIcon class="loading-spinner" />
-        </transition>
-      </NavigatorCardInner>
-      <div aria-live="polite" class="visuallyhidden">
-        Navigator is {{ isFetching ? 'loading' : 'ready' }}
-      </div>
-    </nav>
-  </transition>
+  <nav class="navigator">
+    <NavigatorCard
+      v-if="!isFetching"
+      :technology="technology.title"
+      :technology-path="technology.path || technology.url"
+      :type="type"
+      :children="flatChildren"
+      :active-path="activePath"
+      :scrollLockID="scrollLockID"
+      :error-fetching="errorFetching"
+      :breakpoint="breakpoint"
+      :api-changes="apiChanges"
+      @close="$emit('close')"
+    />
+    <NavigatorCardInner v-else class="loading-placeholder">
+      <transition name="delay-showing" appear>
+        <SpinnerIcon class="loading-spinner" />
+      </transition>
+    </NavigatorCardInner>
+    <div aria-live="polite" class="visuallyhidden">
+      Navigator is {{ isFetching ? 'loading' : 'ready' }}
+    </div>
+  </nav>
 </template>
 
 <script>
@@ -94,10 +89,6 @@ export default {
       default: '',
     },
     errorFetching: {
-      type: Boolean,
-      default: false,
-    },
-    isOpen: {
       type: Boolean,
       default: false,
     },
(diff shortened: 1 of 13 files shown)


marinaaisa:
@swift-ci test

src/views/DocumentationTopic.vue
```diff
@@ -43,6 +43,7 @@
                   :parent-topic-identifiers="parentTopicIdentifiers"
                   :technology="slotProps.technology || technology"
                   :is-fetching="slotProps.isFetching"
+                  :is-open="openExternally"
```
dobromir-hristov, line 46: This is the same as `isSideNavOpen`, as it is defined in the DocTopic.

src/components/Navigator.vue
```diff
@@ -9,7 +9,10 @@
 -->
 
 <template>
-  <nav class="navigator">
+  <nav
+    v-if="displayNavigator"
```
dobromir-hristov, line 13: I am worried this will cause allot of unnecessary re-rendering. Wont applying an `aria-hidden` in the DocumentationTopic be enough?

marinaaisa: Applying only `aria-hidden` will prevent VO navigation but not keyboard navigation. An alternative is applying `display: none` which will prevent keyboard navigation too but I thought it would be better to just not rendering it at all.

dobromir-hristov: This would also I think remove any transition animations we have, when opening/closing the nav on mobile, so we need to think of a better way.

marinaaisa: Updated using v-show: https://github.com/apple/swift-docc-render/pull/296/commits/a70e5c79fa65edb0d93ff6323456f6529d15b427

marinaaisa:
@swift-ci test

src/components/Navigator.vue
hqhhuang: I tested on mobile and the animation seemed fine.

src/components/Navigator.vue
```diff
@@ -210,6 +221,11 @@ export default {
     position: static;
     transition: none;
   }
+
+  &.delay-hiding-leave-active {
+    // don't hide navigator until delay time has passed
+    transition: visibility var(--delay);
```
dobromir-hristov, line 227: This is such an exploit on the Vue transition component 🤣  

We never apply `visibility`, so that we can transition it, but the `display: none` on the `v-show` is delayed, as much as the duration of this transition declaration. It basically waits for it to "finish".

dobromir-hristov (approved):

src/views/DocumentationTopic.vue
```diff
@@ -39,30 +39,36 @@
...
-        <Topic
-          v-bind="topicProps"
-          :key="topicKey"
-          :objcPath="objcPath"
-          :swiftPath="swiftPath"
-          :isSymbolDeprecated="isSymbolDeprecated"
-          :isSymbolBeta="isSymbolBeta"
-          :languagePaths="languagePaths"
-        />
+        <transition name="delay-hiding">
```
dobromir-hristov, line 60: We should be applying an aria-hidden to everything except the navigator. This has been what we do for Modals up to now too, isnt that sufficient?

src/views/DocumentationTopic.vue
```diff
@@ -39,30 +39,36 @@
...
+          <Topic
+            v-show="!isSideNavOpen"
```
dobromir-hristov, line 62: not 100% sure if we flip this boolean back to false, on large breakpoints.

src/views/DocumentationTopic.vue
marinaaisa: No, it's not. Aria-hidden doesn't prevent keyboard navigation, it only prevents VO navigation.
I'll fill bugs for the modals too.

dobromir-hristov: what do you mean by keyboard navigation? Tabbing?

marinaaisa: Yes

dobromir-hristov: Does the FocusTrap work?

marinaaisa: That's a good point. I've been checking the FocusTrap class and actually `event.preventDefault();` inside `onFocus` function is not preventing from focusing outside the container. Is this a regression? Can you test it too?

dobromir-hristov (approved):

dobromir-hristov:
@swift-ci test

marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

PR #305, opened 2022-05-16
Fix scrollbars not showing for CodeListing in Safari

dobromir-hristov:
Bug/issue #, if applicable: 78783950

## Summary

Fixes a bug where codelisting scrollbars are sometimes hidden in Safari, until you resize the screen.

## Dependencies

NA

## Testing

You can use SlothCreator to test.

Try setting scrollbars to `Always Show` in macOS settings.

Steps:
1. Use a resolution that would show horizontal scrollbars. Expand the navigator as much as possible.
2. Go to http://localhost:8082/documentation/slothcreator/sloth
3. Assert that scrollbars are always visible in safari and its fully scrollable
4. Try the same here - http://localhost:8082/tutorials/slothcreator/creating-custom-sloths

## Checklist

- [ ] ~Added tests~ - no need CSS only

### DIFF
--- a/src/components/ContentNode/CodeListing.vue
+++ b/src/components/ContentNode/CodeListing.vue
@@ -154,7 +154,8 @@ export default {
 pre {
   padding: $code-listing-with-numbers-padding;
   display: flex;
-  overflow: auto;
+  // set as `unset` to fix a Safari issue, where the scrollbar is hidden until you resize window
+  overflow: unset;
   -webkit-overflow-scrolling: touch;
   white-space: pre;
   word-wrap: normal;


dobromir-hristov:
@swift-ci test

src/components/ContentNode/CodeListing.vue
```diff
@@ -154,7 +154,8 @@ export default {
 pre {
   padding: $code-listing-with-numbers-padding;
   display: flex;
-  overflow: auto;
+  // set as `unset` to fix a Safari issue, where the scrollbar is hidden until you resize window
+  overflow: unset;
```
mportiz08, line 158: This is to counteract an issue where a parent of this element has also set `overflow: auto`, right?

In that case, should we consider removing the rule in the parent(s) instead? This change would mean that the default behavior of the component doesn't behave as expected with regards to overflow and scrolling if I'm understanding correctly and would only work when the overflow has been set by the thing using this component.

dobromir-hristov: Yes, but removing the `overflow: auto` on the parent and leaving it here (which is ofc what I did initially 😆 ) results in the scrollbar going outside of the borders of the component, looking VERY strange :/

mportiz08: I can imagine, haha.

Could the parent maybe do something like this though?

```
...
overflow: auto;
...
/deep/ .code-listing pre {
  overflow: unset;
}
```

That way we keep the proper default in this component and only unset it where needed? Sorry if I'm not quite understanding the exact issue!

dobromir-hristov: Unfortunately, by parent, I meant parent element -`.code-listing`, which is inside this component.

If you look here, it has an overflow as well - https://github.com/apple/swift-docc-render/blob/ce69a8c1e4bdc5662dba7c99923ba376a94aca95/src/components/ContentNode/CodeListing.vue#L190

mportiz08: Ah, got it. In that case the default behavior thing I mentioned at first isn't a concern since anyone using this component would still get the overflow rule from the code-listing element.

I'm fine with this change then...but should we maybe just remove the `overflow` rule from the pre entirely? Or does that also break things? :P

mportiz08 (approved):

src/components/ContentNode/CodeListing.vue
dobromir-hristov: The pre has an overflow rule by default, from the browsers :/

dobromir-hristov:
@swift-ci test

PR #306, opened 2022-05-16
Fix issue with navigating up/down in navigator

dobromir-hristov:
Bug/issue #, if applicable: 93332275

## Summary

Fixes a bug where if you navigate down in the navigator, it skips the Headings and throws errors in the console.

This is caused by a recent refactor, where we call `$refs.reference.$el.focus()`, but the `reference` can be a Vue component or a DOM element, in the second case it throws an error. That would be when we render an H3 instead of the Reference component.

## Dependencies

NA

## Testing

Steps:
1. Navigate down the navigator using arrow keys.
2. Assert no errors are visible in console.

## Checklist

- [ ] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCardItem.vue
+++ b/src/components/Navigator/NavigatorCardItem.vue
@@ -205,10 +205,10 @@ export default {
       this.toggleTree();
     },
     clickReference() {
-      this.$refs.reference.$el.click();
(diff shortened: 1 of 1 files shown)


marinaaisa (approved):
Thanks for fixing this!

hqhhuang (approved):

dobromir-hristov:
@swift-ci test

PR #328, opened 2022-05-23
Fix iOS navigator not calculating height properly

dobromir-hristov:
Bug/issue #, if applicable: 93506244

## Summary

Exposes the viewport height as a Custom CSS Property, so it can be used for CSS calc expressions, instead of using `100vh` units.

This is to fix a mis-calculation on iPads, where the height of the top and bottom safari navbars are not accounted for in the height of the navigator

## Dependencies

NA

## Testing

Please test this on Android, Safari and on Desktop on the popular browsers. This could not only add a performance degradation, but as you scroll, it may shift the height of the navigator, causing stuttering.

Steps:
1. Open a page with a navigator on iPad
2. Assert you can scroll all the way down and see the last navigator items, while still having the Safari nav items visible

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/AdjustableSidebarWidth.vue
+++ b/src/components/AdjustableSidebarWidth.vue
@@ -104,6 +104,7 @@ export default {
   },
   data() {
     const windowWidth = window.innerWidth;
+    const windowHeight = window.innerHeight;
     const breakpoint = BreakpointName.large;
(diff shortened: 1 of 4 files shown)


dobromir-hristov:
@swift-ci test

PR #333, opened 2022-05-26
Remove nav background flashing, when changing between color modes

dobromir-hristov:
Bug/issue #, if applicable: 90712731

## Summary

Extract the nav bg transition timing to a scss variable. This way clients can control the transition timing in their templates.

## Dependencies

NA

## Testing

Steps:
1. Go to a documentation page
2. Switch between color modes (light/dark)
3. Assert the nav no longer has a transition/delay in the background color
4. Assert that colors are still shifted where expected, like on TutorialOverview, where we apply a blur to the background

## Checklist

- [x] Added tests


marinaaisa (approved):

dobromir-hristov:
@swift-ci test

PR #337, opened 2022-05-27
Use references to build secondary navigation urls in NavigationBar

dobromir-hristov:
Bug/issue #, if applicable: 63628221

## Summary

Modifies the NavigationBar to use paths from the references to build the URLs for SecondaryDropdown and MobileDropdown elements.

## Dependencies

NA

## Testing

Steps:
1. Go to a technologies page, like in Slothcreator
2. Assert links on the page in the navigator are generated and properly working

## Checklist

- [x] Added tests


src/components/Tutorial/NavigationBar.vue
```diff
@@ -117,11 +126,26 @@ export default {
...
+      return linkableSections.map((linkableSection, index) => {
+        const singleSection = allSections[index];
+        const referenceObject = this.references[singleSection.reference];
+        const { url, title } = referenceObject || singleSection;
+        return {
+          depth: linkableSection.depth,
+          visibility: linkableSection.visibility,
+          sectionNumber: linkableSection.sectionNumber,
+          title,
+          path: url,
```
hqhhuang, line 146: Is url a required "key" for the renderjson? If not, should we keep the old logic to make sure if no url is provided, we fall back to use the anchor?

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@mportiz08 do you mind taking a look at this? Make sure I am not missing something.

src/components/Tutorial/NavigationBar.vue
```diff
@@ -117,11 +126,26 @@ export default {
   },
   computed: {
     pageSections() {
-      const tutorialPath = this.$route.path.replace(/#$/, '');
-      return this.tutorialState.linkableSections.map(section => ({
-        ...section,
-        path: `${tutorialPath}#${section.anchor}`,
-      }));
+      const projectReferences = this.chapters
```
mportiz08, line 129: I think it could help to create a new computed property (that depends on `chapters` and `identifierUrl`) which returns the the current project from the chapters list representing this tutorial, which can just be used here as `project.sections` (etc)

src/components/Tutorial/NavigationBar.vue
```diff
@@ -117,11 +126,26 @@ export default {
...
+
+      const allSections = [FirstSectionItem, ...projectReferences.sections];
+      const { linkableSections } = this.tutorialState;
+
+      return linkableSections.map((linkableSection, index) => {
+        const singleSection = allSections[index];
+        const referenceObject = this.references[singleSection.reference];
+        const { url, title } = referenceObject || singleSection;
+        return {
+          depth: linkableSection.depth,
```
mportiz08, line 142: Possibly these 3 lines could be simplified to just `...linkableSection`?

mportiz08 (commented):
Seems like the right approach to me.

The code is a little confusing, but that's just because the top level `sections` and `hierarchy` data are just kind of separate from each other and the main sections data doesn't contain any actual full references to the sections like it should (I think the `hierarchy` field came later which explains why there's kind of some redundancy between the two fields)

dobromir-hristov:
@swift-ci test

mportiz08 (approved):

PR #338, opened 2022-05-28
Fix preview box getting cut off for Tutorial pages

hqhhuang:
Bug/issue #, if applicable: rdar://93843834

## Summary
Fixes issue where tutorial preview frame looks broken after hiding and re-expanding preview. 

## Testing
Steps:
1. Navigate to a tutorial page with previews
2. Verify that UI is consistent even after hiding and re-expanding the preview box. 

## Checklist

- [ ] Added tests


dobromir-hristov (commented):
I dont think this is the correct visual we are after. The box shadow goes away as it animates out to 0. The design team should take a look too.

image

hqhhuang:
@swift-ci test

dobromir-hristov (commented):
I think we should do a deeper dive into this.

If a preview is not sent with a `size`, the entire calculation goes off, because it calculates height as `Nan` and then we do weird stuff on top of that.

Your fix just removes the `height` property entirely, if `height` is not available, which makes the outer shadow element no animate anymore. This may be the only way, given we dont know the image height upfront, so we just wait for it to expand the content, but we lose the animation.

Adding a hard height fallback may be a solution as well, although that may look off for variable aspect ratios of images, as we already have a hard set width.

marinaaisa:
What animation do you speak about @dobromir-hristov ?

Using `initial` as height is the same as `auto` and it looks good to me, but I'm not sure what animation I'm missing out.

Screenshot 2022-06-01 at 17 06 37

biamx3:
The animation looks a little jarring to me. The issue I'm seeing is the divider line between the title container and the white area is appearing too soon, and the outline of the white area is disappearing altogether. 

What should happen is the container resizes, and we don't see the line animating in at all. 

https://user-images.githubusercontent.com/11503099/171734236-49dbb5fa-3a5e-4bac-b357-6e2b9f382d47.mov
Screen Shot 2022-06-02 at 1 42 05 PM

dobromir-hristov:
Checkout - https://github.com/hqhhuang/swift-docc-render/pull/19, for the improved transitions

hqhhuang:
@swift-ci test

marinaaisa (approved):
Thanks, Hanqing :)

hqhhuang:
@swift-ci test

PR #341, opened 2022-05-30
[AX] Localnav should not FocusTrap for documentation pages

marinaaisa:
Bug/issue #94044029, if applicable: 

## Summary

Localnav should not FocusTrap for documentation pages but we should keep it for the rest of the pages.

This is because localnav in small viewports in documentation pages is very small, only has 2 items, and it doesn’t feel like a modal, it’s more like a collapse element.

## Dependencies

NA

## Testing

Steps:
1. Using SlothCreator, run DocC Render
2. Go to http://localhost:8080/documentation/slothcreator
3. Go into a small viewport
4. Open the localnav
5. Assert that when tabbing focus doesn't get trap in the localnav

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/NavBase.vue
+++ b/src/components/NavBase.vue
@@ -146,6 +146,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    preventFocusTrap: {
+      type: Boolean,
+      default: false,
+    },
   },
   mixins: [onIntersect],
   data() {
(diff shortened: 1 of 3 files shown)


src/components/NavBase.vue
```diff
@@ -226,6 +232,7 @@ export default {
       this.isTransitioning = false;
       if (this.isOpen) {
         this.$emit('opened');
+        if (this.preventFocusTrap) return;
```
dobromir-hristov, line 235: should we also prevent scroll locking? At this point we are going into mobile viewport territory.

dobromir-hristov (commented):
I am not an AX expert, so I cant say what is right or wrong here, I just want to point out that there are occasions, where docs can have 6 links with breadcrumbs, language toggles and so on.

marinaaisa:
Closed in favor of https://github.com/apple/swift-docc-render/pull/427

PR #347, opened 2022-06-03
Fix replay video link needing to be clicked twice on Safari Desktop

dobromir-hristov:
Bug/issue #, if applicable: 94304930

## Summary

Fixes a randomly popping up issue, where the video would freeze, if we click `replay`. This happened only on Safari Desktop.

This was caused by us reseting the `currentTime` of the video and then calling the `play` method.

`play` already resets the video to the start if it was ended, and we only show that after it has ended anyway.

## Dependencies

NA

## Testing

Use an archive that has a tutorial with vidoes in it and Safari on Desktop

Steps:
1. wait for the video to finish playing.
2. Click `replay`
3. Do that a few times and assert it never freezes. Assert there are no other regressions from the change.

## Checklist

- [x] Added tests


mportiz08 (approved):

mportiz08:
@swift-ci test

PR #352, opened 2022-06-09
<rdar://74717074> Fix fetch redirects

dobromir-hristov:
Bug/issue #, if applicable: 74717074

## Summary

Properly follow redirects with Vue Router. It assumes the redirects come back in the format `/data/documentation/foo.json?optionalQuery=foo`, from which we extract `/documentation/foo?optionalQuery=foo` and redirect Vue router to it.

Then data fetching would commence as normal.

## Dependencies

NA

## Testing

To test this you need to have a server that responds with a 301 for a JSON with a location pointing to the new location.
The server should include the baseURL in the location header.

## Checklist

- [x] Added tests


src/utils/data.js
```diff
@@ -71,6 +87,12 @@ export async function fetchDataForRouteEnter(to, from, next) {
       throw false;
     }
 
+    if (error instanceof RedirectError) {
+      // throw the redirect location, so it's passed to the `next` error handler and
+      // vue router redirects to that location
+      throw transformDataPathToRoutePath(error.location);
```
mportiz08, line 93: I'm guessing you already tried this and it didn't work, but could we just explicitly call `next(error.location)` here instead of rethrowing the string and relying on the navigation guard `catch` to call `next`?

src/errors/FetchError.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
```
mportiz08, line 4:
```suggestion
 * Copyright (c) 2022 Apple Inc. and the Swift project authors
```

src/errors/RedirectError.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
```
mportiz08, line 4:
```suggestion
 * Copyright (c) 2022 Apple Inc. and the Swift project authors
```

src/utils/data.js
mportiz08: Er, I mean `next(transformDataPathToRoutePath(error.location))`

dobromir-hristov: It will work, but the promise from `fetchDataForRouteEnter` would resolve and because we return nothing, for a brief moment, we would get an empty page, because `vm.topicData = undefined` basically. 

We can also return the JSON and call next, then in the `beforeRouteUpdate` somehow detect what happened and skip fetching, but at this point it feels overcomplicated.

mportiz08 (approved):

mportiz08:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #355, opened 2022-06-14
Show group markers in search

dobromir-hristov:
Bug/issue #, if applicable: 90655322

## Summary

Group markers should show up in results, along with all the items they label, when the query matches exactly the group name

## Dependencies

NA

## Testing

Use SlothCreator as an example

Steps:
1. Assert you can search directly for the name of a Group Marker (Label). Assert all of it's children are rendered, and you can open them to render more children.
2. Assert searching for parts of a group do not show it or it's children.
3. Assert you can still apply tags and those dont apply
4. Assert clicking "Hide Deprecated" continues to hide the deprecated tags, 
5. Assert deprecated items are hidden, even after matching a group name
6. Assert nothing else is broken in the filtering

## Checklist

- [x] Added tests


biamx3:
I think there shouldn't need to be an exact match for us to see results, the titles should have the same rules as the rest of the content. It's easier for folks to find content by looking up keywords, rather than expecting them to remember the whole title. So for example, in the "Sloth Creator" framework, if I type in "care", I should see the "Care and Feeding" topic title.

biamx3:
Tested again, felt really natural and good. Approved on my end.

dobromir-hristov:
@swift-ci test

marinaaisa (approved):
Code wise, it looks good to me! thanks Dobri!

dobromir-hristov:
@swift-ci test

PR #357, opened 2022-06-16
Refactor of url resolvers and add param for custom url resolver in the metadata mixin

marinaaisa:
Bug/issue #94575097, if applicable: 

## Summary

Refactor of url resolvers and add param for custom url resolver in the metadata mixin

## Dependencies

NA

## Testing

Steps:
1. Run docc-render and verify that everything works correctly

## Checklist

- [x] Added tests

### DIFF
--- a/src/utils/assets.js
+++ b/src/utils/assets.js
@@ -69,7 +69,7 @@ export function pathJoin(parts) {
  * @param {String} url
  * @return {String}
  */
-export function normalizeAssetUrl(url) {
+export function normalizeUrl(url) {
   if (!url || typeof url !== 'string' || url.startsWith(baseUrl) || !url.startsWith('/')) {
     return url;
   }
--- a/src/utils/assets.js
+++ b/src/utils/assets.js
@@ -11,7 +11,7 @@
 /**
  * Utility functions for working with Assets
  */
-import { baseUrl } from 'docc-render/utils/theme-settings';
+const { baseUrl } = window;
 
 /**
  * Separate array of variants by light/dark mode
--- a/src/utils/data.js
+++ b/src/utils/data.js
@@ -9,7 +9,9 @@
 */
 
 import { pathJoin } from 'docc-render/utils/assets';
-import { queryStringForParams, areEquivalentLocations, resolveAssetsAbsoluteUrl, resolveAbsoluteUrl } from 'docc-render/utils/url-helper';
+import {
+  queryStringForParams, areEquivalentLocations, getAssetsAbsoluteUrl, getAbsoluteUrl,
+} from 'docc-render/utils/url-helper';
 import emitWarningForSchemaVersionMismatch from 'docc-render/utils/schema-version-check';
 import { baseUrl } from 'docc-render/utils/theme-settings';
 import RedirectError from 'docc-render/errors/RedirectError';
@@ -30,7 +32,7 @@ export async function fetchData(path, params = {}) {
     return !response.ok;
   }
 
-  const url = resolveAbsoluteUrl(path);
+  const url = getAbsoluteUrl(path);
   const queryString = queryStringForParams(params);
   if (queryString) {
     url.search = queryString;
@@ -132,6 +134,6 @@ export function clone(jsonObject) {
 }
 
 export async function fetchIndexPathsData() {
-  const path = resolveAssetsAbsoluteUrl('index/index.json');
+  const path = getAssetsAbsoluteUrl('/index/index.json');
   return fetchData(path);
 }
--- a/src/utils/data.js
+++ b/src/utils/data.js
@@ -10,10 +10,9 @@
 
 import { pathJoin } from 'docc-render/utils/assets';
 import {
-  queryStringForParams, areEquivalentLocations, getAssetsAbsoluteUrl, getAbsoluteUrl,
+  queryStringForParams, areEquivalentLocations, getAbsoluteUrl,
 } from 'docc-render/utils/url-helper';
 import emitWarningForSchemaVersionMismatch from 'docc-render/utils/schema-version-check';
-import { baseUrl } from 'docc-render/utils/theme-settings';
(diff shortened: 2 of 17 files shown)


mportiz08:
I'm not sure that this is the right approach since we _do_ actually want to include a base path (if provided) with metadata URLs in addition to URLs for static assets requests.

Thinking about an example GitHub Pages hosted site, like [this one](https://ethan-kusters.github.io/swift-argument-parser/documentation/argumentparser/):

Note how it includes the following:
```
<meta property="og:url" content="https://ethan-kusters.github.io/swift-argument-parser/documentation/argumentparser/">
```

When I test the changes from this PR, the `/swift-argument-parser/` prefix gets dropped, and the URL would then be incorrect.

I think we would need to find an alternate approach to the problem you're trying to solve here.

src/utils/assets.js
```diff
@@ -11,7 +11,7 @@
 /**
  * Utility functions for working with Assets
  */
-import { baseUrl } from 'docc-render/utils/theme-settings';
+const { baseUrl } = window;
```
dobromir-hristov, line 14: I dont think this is necessary?

dobromir-hristov (commented):
I really like the overall refactor, but as Marcus said, dropping the `baseUrl` from some of those places is a regression, like the index or the `resolveAbsoluteUrl`.

src/utils/assets.js
marinaaisa: Importing the baseURL from theme-settings creates a dependency cycle between `url-helper`, `theme-settings` and `assets` util files. More info: https://github.com/import-js/eslint-plugin-import/blob/v2.23.4/docs/rules/no-cycle.md 

Since window is a global variable, I decided to just take it as it is.

marinaaisa:
@swift-ci test

src/utils/data.js
```diff
@@ -55,8 +56,7 @@ export async function fetchData(path, params = {}) {
 }
 
 function createDataPath(path) {
-  const dataPath = path.replace(/\/$/, '');
-  return `${pathJoin([baseUrl, 'data', dataPath])}.json`;
+  return `${pathJoin(['/data', path.replace(/\/$/, '')])}.json`;
```
dobromir-hristov, line 59: by removing the `baseUrl` here, wont this break deployments to none root paths? Like https://ethan-kusters.github.io/swift-argument-parser/documentation/argumentparser/

src/utils/data.js
```diff
@@ -132,6 +132,6 @@ export function clone(jsonObject) {
 }
 
 export async function fetchIndexPathsData() {
-  const path = new URL(`${pathJoin([baseUrl, 'index/index.json'])}`, window.location.href);
+  const path = '/index/index.json';
```
dobromir-hristov, line 135: same baseUrl question here.

src/utils/theme-settings.js
```diff
@@ -27,7 +28,7 @@ export const { baseUrl } = window;
  * @return {Promise<{}>}
  */
 export async function fetchThemeSettings() {
-  const url = new URL(`${baseUrl}theme-settings.json`, window.location.href);
+  const url = resolveAbsoluteUrl('/theme-settings.json');
```
dobromir-hristov, line 31: Here you are also removing the baseUrl, same question as above

dobromir-hristov (commented):
Hey Marina, refactor looks good, but I am genuinely worried about those baseURLs :D 

If we have `new URL('/theme-settings.json, 'http://mywebsite.com/mydocs/documentation')`, it will resolve to `http://mywebsite.com/theme-settings.json` which would be wrong, if my baseURL, is `/mydocs/`. 

The word `baseURL` is a bit misleading, because its a basePath, not fully qualified url.

src/utils/data.js
marinaaisa: Not really because I'm using `getAbsoluteUrl` in https://github.com/apple/swift-docc-render/pull/357/files#diff-3cf45312eb60039df9dfa67d90bc0936e830f0d3424cf0b09943b1e3b09d700fR34 which contains the baseUrl on it. Before we weren't adding the baseUrl inside the `fetchData` function but now we do.

src/utils/theme-settings.js
marinaaisa: https://github.com/apple/swift-docc-render/pull/357#discussion_r901793839

marinaaisa:
I answered your question, @dobromir-hristov , please review again, thanks!

src/utils/data.js
dobromir-hristov: mmmm no, thats not correct.

https://github.com/apple/swift-docc-render/pull/357/files#diff-3cf45312eb60039df9dfa67d90bc0936e830f0d3424cf0b09943b1e3b09d700fR34 - this creates a fully qualified URL from a path, but for the most part, that just sticks the path to whatever the current `window.location.href` is. Since our paths start with `/` it will remove the entire `pathname` from the current `location.href`.

The path we pass is transformed by `createDataPath` which was adding the baseUrl prefix, which you removed. 

Here you can see what I mean, you are essentially introducing a bug with deployments to subpaths.

image

marinaaisa: I believe there is a misunderstanding on the `url-helper.js` file since the param for the domain path was also called `baseURL` but it's just the `domainPath`. I changed it here: https://github.com/apple/swift-docc-render/pull/357/commits/52827fbd6f966bee5a16e4ae801b31a9ccfedea6

I didn't remove the baseURL prefix, it happens in the `normalizeUrl` function inside `getAbsoluteUrl`.

The baseURL prefix happens inside the `path` const and in the case of `theme-settings.json` it's a file that is hosted in the source root so it makes sense to delete the rest of the pathname in the window.location.href

Screenshot 2022-06-21 at 15 38 08

marinaaisa:
Closed in favor of https://github.com/apple/swift-docc-render/pull/379

PR #363, opened 2022-06-21
LanguageToggle dropdown should have a cursor-pointer

dobromir-hristov:
Bug/issue #, if applicable: 89923444

## Summary

Language toggle should have a pointer cursor icon.

## Dependencies

NA

## Testing

Open an archive that has language versions

Steps:
1. Assert that hovering over the language toggle shows a pointer cursor

## Checklist

- [ ] Added tests

### DIFF
--- a/src/components/DocumentationTopic/DocumentationNav/LanguageToggle.vue
+++ b/src/components/DocumentationTopic/DocumentationNav/LanguageToggle.vue
@@ -240,6 +240,7 @@ $nav-menu-toggle-label-margin: 6px;
     padding: 0 $dropdown-icon-padding 0 4px;
     margin-left: -4px;
     @include font-styles(nav-toggles);
+    cursor: pointer;
 
     // remove the default focus styles, and re-add them on keyboard navigation, only.
     &:focus {
--- a/src/components/DocumentationTopic/DocumentationNav/LanguageToggle.vue
+++ b/src/components/DocumentationTopic/DocumentationNav/LanguageToggle.vue
@@ -241,6 +241,8 @@ $nav-menu-toggle-label-margin: 6px;
     margin-left: -4px;
     @include font-styles(nav-toggles);
     cursor: pointer;
+    position: relative;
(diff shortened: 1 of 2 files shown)


biamx3:
It's odd to me that when I hover over the "language:" label, the cursor doesn't switch to the hand but the language itself (like Swift) changes color to blue. 

Can you please make it so the color doesn't change when you hover the "language:" label? 

Also, since the chevron is part of the button, on hover, the cursor should also change to the hand and the language color should switch to blue.

dobromir-hristov:
@biamx3 I fixed the pointer on the chevron, but not the blue color on label hover.

That looks like part of the general browser behaviour, as the label is linked to the dropdown, for accessibility reasons. Otherwise it wont be read out in VO when you focus on the dropdown.

biamx3:
Got it, the accessibility aspect makes sense. With that in mind, I think we should make the label clickable as well as the language and chevron. So if I hover on top of any part of the component, the language label will turn to blue and the cursor will change to pointing hand. 

FYI, I tested the branch again today on Safari and Chrome and I'm not seeing the pointing hand cursor anymore – could be just me, but wanted to give you a heads up.

dobromir-hristov:
Hey Bea, 

Cursor hand is still there, just tried it out.

However I have bad news about the clickable label. Turns out this is a browser limitation, and you cannot expand the select dropdown, when the user clicks on the label.

It should have already worked like that, but as you saw, it doesnt.

biamx3:
> Cursor hand is still there, just tried it out.
Weird! I can't get it to show up for me anymore. 

> However I have bad news about the clickable label. Turns out this is a browser limitation, and you cannot expand the select dropdown, when the user clicks on the label.
I see... Bummer! Ok then, let's stick to the previous version where we have the pointing hand + blue label color on hover of the language (Swift) and chevron, and just the blue color on hover on the language (language) label.

dobromir-hristov:
@swift-ci test

PR #384, opened 2022-07-13
Add issue templates to docc-render

marinaaisa:
Bug/issue #96950970, if applicable: 

## Summary

Add issue templates to docc-render following https://github.com/apple/swift-docc/tree/main/.github/ISSUE_TEMPLATE but adapted to DocC Render

## Dependencies

NA

## Testing

Steps:
1. Review that text added for issue templates is correct

## Checklist

- [ ] Added tests


.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,61 @@
...
+    description: |
+      Explanation of how to reproduce the incorrect behavior. 
+      
+      This could include a reduced version of your documentation bundle, or a link to the documentation (or code) that is exhibiting the issue.
+    placeholder: |
+      1. In this browser version...
+
+      2. With this viewport size...
+      
+      3. Run '...'
```
franklinsch, line 52: Not sure what this would be referring to, since we don't run commands in Swift-DocC-Render. Is something like "Perform action…" more appropriate?

.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yml
```diff
@@ -0,0 +1,47 @@
...
+    required: true
+- type: textarea
+  attributes:
+    label: Description
+    description: |
+      A description of your proposed feature.
+
+      Design proposals to show what's missing, or what new capabilities will be possible, are very helpful! Provide links to existing issues or external references/discussions, if appropriate.
+
+      How does it behaves in different viewports?
```
franklinsch, line 24:
```suggestion
      How does it behave in different viewports?
```

franklinsch: I'm not sure this is generally relevant for feature requests. This would only apply to specific kind of UI-related requests, so I'm not sure it should be included in the template.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
marinaaisa: I was thinking of a step like:

3. Run `VUE_APP_DEV_SERVER_PROXY=[path to docc] npm run serve`

.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yml
marinaaisa: To me it's a question for the user who is writing the proposal to think about how the feature that wants to implement also works in different viewports. Does it make sense?

.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yml
```diff
@@ -0,0 +1,47 @@
+name: Feature Request
+description: A suggestion for a new feature
+labels: [enhancement]
```
mportiz08, line 3: I think there's an explicit new feature label if that is what this is pulling from.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
franklinsch: I expect that most issues will be filed after noticing a bug in some hosted version of a DocC Render website somewhere, rather than running things locally, so I don't think serving DocC Render locally should be part of the template.

.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yml
franklinsch: Maybe we can also add something like "Are there accessibility considerations for this feature?"

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
marinaaisa: Makes sense! deleted d4ec0eeeda23beee35f7d5df4a662e79450e7d26

marinaaisa:
@swift-ci test

PR #390, opened 2022-07-14
Plist links in topics section should not be in code voice

hqhhuang:
Bug/issue #, if applicable: 62751444

## Summary
Plist links in the Topics section should not be in code voice

## Testing
Manually verify in browser 
Steps:
1. Navigate to a documentation with links to property list pages in the topics section
2. Verify that the link is rendered in plain text instead of code voice

## Checklist
- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic/TopicsLinkBlock.vue
+++ b/src/components/DocumentationTopic/TopicsLinkBlock.vue
@@ -180,8 +180,8 @@ export default {
       if (topic.titleStyle === TitleStyles.title) {
         return topic.ideTitle ? 'span' : 'code';
       }
-      // Framework name links should not be code voice
-      if (topic.role && topic.role === TopicRole.collection) return 'span';
+      // Framework name and property list links and should not be code voice
+      if (topic.role && (topic.role === TopicRole.collection || topic.role === TopicRole.dictionarySymbol)) return 'span';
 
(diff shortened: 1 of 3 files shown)


src/components/DocumentationTopic/TopicsLinkBlock.vue
```diff
@@ -180,8 +180,8 @@ export default {
       if (topic.titleStyle === TitleStyles.title) {
         return topic.ideTitle ? 'span' : 'code';
       }
-      // Framework name links should not be code voice
-      if (topic.role && topic.role === TopicRole.collection) return 'span';
+      // Framework name and property list links and should not be code voice
+      if (topic.role && (topic.role === TopicRole.collection || topic.role === TopicRole.dictionarySymbol)) return 'span';
```
mportiz08, line 184: I was initially worried that this might not work, because other symbols like REST objects also have a `dictionarySymbol` role, and we _do_ want to use code voice for those. However, after testing a bit, it looks like this is handled by the template, since these topics have `fragments`, which means the `DecoratedTopicTitle` component is used and this code never runs in those instances.

All that to say, seems like this is working even though I didn't expect it to 😅

mportiz08 (approved):

hqhhuang:
@swift-ci test

PR #395, opened 2022-07-19
Fix multiline detection in DeclarationSource for ObjC

dobromir-hristov:
Bug/issue #, if applicable: 97024618

## Summary

Moves the multiline code detection after we indent ObjC code, so we can properly detect newly created new lines.

## Dependencies

NA

## Testing

This bug is seen in Safari only, so use that to test.

Steps:
1. Go to a page that has a multiline declaration in ObjC language
2. Assert with the inspector, that there is a `.has-multiple-lines` class on the `.source` class.

## Checklist

- [x] Added tests


dobromir-hristov:
@swift-ci test

marinaaisa (approved):

PR #396, opened 2022-07-20
[GSoC] Quick navigation UI

sofiaromorales:
Bug/issue #, if applicable: NA

## Summary

UI for Quick Navigation feature with keyboard navigation support

## Dependencies

NA

## Testing

Steps:
1. Set the `quickNavigation.enable` flag to true inside `theme-settings.json`
2. Open the quick navigation modal component by clicking on the button inside the filter input, `/` key or `⌘ + shift + o`
3. Type something inside the input bar to trigger the fuzzy search
4. Assert that the symbols listed correspond to your input
5. Use arrow keys to move through the results list
6. Press `Enter` to access the symbol

Demo: https://sofiaromorales.github.io/swift-argument-parser/documentation/argumentparser/

## Checklist

- [ ] Added tests


sofiaromorales:
@marinaaisa ready for review :)

marinaaisa (changes requested):
Thanks, Sofía!
I started reviewing your PR, I'll continue tomorrow. I leave you some comments in the meantime.

src/components/Filter/FilterInput.vue
```diff
@@ -16,7 +16,7 @@
     :aria-labelledby="searchAriaLabelledBy"
     :class="{ 'focus': showSuggestedTags }"
     @blur.capture="handleBlur"
-    @focus.capture="showSuggestedTags = true"
+    @focus.capture="showSuggestedTags = true; $emit('toggleIsFilterInputFocused')"
```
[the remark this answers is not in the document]
sofiaromorales: Resolved in https://github.com/apple/swift-docc-render/pull/396/commits/83c9145e7637f7ad05904dd32af1cbd3580eb97c

src/components/Filter/FilterInput.vue
```diff
@@ -430,6 +440,29 @@ $input-height: rem(28px);
...
+      color: var(--input-text);
+      border: solid 1px;
+      border-radius: $border-radius;
+      border-color: var(--color-grid);
+      display: flex;
+      align-items: center;
+      .filter__open-modal-key {
+        text-decoration: none;
+        margin: auto;
+        font-size: rem(12px);
```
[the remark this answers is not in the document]
sofiaromorales: Changed for `body-reduced-tight` font style instead in https://github.com/apple/swift-docc-render/pull/396/commits/83c9145e7637f7ad05904dd32af1cbd3580eb97c

src/components/Navigator/NavigatorCard.vue
```diff
@@ -90,14 +90,25 @@
             :clear-filter-on-tag-select="false"
             class="filter-component"
             @clear="clearFilters"
+            @toggleIsFilterInputFocused="toggleIsFilterInputFocused"
```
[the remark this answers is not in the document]
sofiaromorales: Reverted this change in https://github.com/apple/swift-docc-render/pull/396/commits/83c9145e7637f7ad05904dd32af1cbd3580eb97c

src/components/Filter/FilterInput.vue
```diff
@@ -276,6 +279,9 @@ export default {
         'paste-tags': handlePaste,
       }
     ),
+    enableQuickNavigation: () => (
```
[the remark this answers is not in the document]
sofiaromorales: Is being used now to render the quick nav button and add some event listeners https://github.com/apple/swift-docc-render/pull/396/commits/83c9145e7637f7ad05904dd32af1cbd3580eb97c

src/components/Filter/FilterInput.vue
```diff
@@ -207,6 +209,7 @@ export default {
       SuggestedTagsId,
       AXinputProperties,
       showSuggestedTags: false,
+      store: QuickNavigationStore,
```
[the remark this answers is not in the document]
sofiaromorales: I made some code cleanup in this commit https://github.com/apple/swift-docc-render/pull/396/commits/34e2bd41ec13c72a3b04bd31703c6a09b8a506b5 thanks!

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,41 +9,107 @@
...
+    >
+    </div>
+    <div
+      class="quick-navigation__container"
+    >
+      <div class="quick-navigation__input-container">
+        <div class="quick-navigation__magnifier-icon-container">
+          <MagnifierIcon />
+        </div>
+        <input
```
marinaaisa, line 30: How is that we aren't using FilterInput here? We could reuse most of the HTML and CSS code so this component would become much smaller.

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -132,57 +215,179 @@ export default {
         });
       }).filter(Boolean)
     ),
+    handleKeyDown() {
```
marinaaisa, line 218: We have a mixing called keyboardNavigation that we use across the app to handle keyboard navigation 😃 Could we reuse it for this implementation too so we don't duplicate code?

src/views/DocumentationTopic.vue
```diff
@@ -9,10 +9,17 @@
...
-    <div v-show="quickNavigationStore.state.showQuickNavigation">
-      <QuickNavigationModal />
-    </div>
+  <div>
+    <PortalTarget name="modal-destination" multiple />
+    <GenericModal
+      :codeBackgroundColorOverride="'transparent'"
+      :isFullscreen=true
+      :showClose=false
+      :theme="'code'"
```
marinaaisa, line 18:
```suggestion
      codeBackgroundColorOverride="transparent"
      :isFullscreen="true"
      :showClose="false"
      theme="code"
```

tests/unit/components/Navigator/QuickNavigationModal.spec.js
```diff
@@ -9,37 +9,89 @@
...
 
 import { shallowMount } from '@vue/test-utils';
+import NavigatorLeafIcon from '@/components/Navigator/NavigatorLeafIcon.vue';
 import QuickNavigationHighlighter from '@/components/Navigator/QuickNavigationHighlighter.vue';
 import QuickNavigationModal from '@/components/Navigator/QuickNavigationModal.vue';
 import Reference from '@/components/ContentNode/Reference.vue';
 
+jest.mock('docc-render/utils/theme-settings', () => ({
+  getSetting: () => true,
+}));
```
marinaaisa, line 19: This fallback may be more useful than just returning true :)

```suggestion
jest.mock('docc-render/utils/theme-settings', () => ({
  getSetting: jest.fn((_, fallback) => fallback),
}));
```

marinaaisa (changes requested):
Hi Sofía, thanks a lot for this huge PR! It's great work 🙌 I added some comments.

My main concerns is to see how we can do FilterInput more generic without adding specific logic to it. I left comments about it. Also, there is an AX issue with the keyboard tabbing backwards, I left another comment.

Congratulations on the tests, I'll see if we can add more and let you know. But your current progress on testing is very very good :)

src/components/Filter/FilterInput.vue
```diff
@@ -395,6 +414,23 @@ export default {
         this.$emit('focus-prev');
       }
     },
+    openQuickNavigationModal() {
+      this.quickNavigationStore.toggleShowQuickNavigationModal(true);
+      return true;
```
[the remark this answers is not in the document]
sofiaromorales: Resolved in https://github.com/apple/swift-docc-render/pull/396/commits/aaa7e4f770e7fbb1c3daeb07fb0ea43867bc231f

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: Refactor made in https://github.com/apple/swift-docc-render/pull/396/commits/eb4841068d99750abcfe827c5935efe2862901a8, thx for the suggestion!

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: sure thanks for pointing that out, resolved in https://github.com/apple/swift-docc-render/pull/396/commits/aaa7e4f770e7fbb1c3daeb07fb0ea43867bc231f

src/views/DocumentationTopic.vue
sofiaromorales: https://github.com/apple/swift-docc-render/pull/396/commits/eb4841068d99750abcfe827c5935efe2862901a8

tests/unit/components/Navigator/QuickNavigationModal.spec.js
sofiaromorales: https://github.com/apple/swift-docc-render/pull/396/commits/aaa7e4f770e7fbb1c3daeb07fb0ea43867bc231f

src/components/Filter/FilterInput.vue
```diff
@@ -98,6 +98,20 @@
             <ClearRoundedIcon />
           </button>
         </div>
+        <button
```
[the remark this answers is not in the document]
sofiaromorales: Since I moved out the `/` button I made a lot of cleanup in this component, and added some customization points through `<slot/>` instead

src/components/Filter/FilterInput.vue
```diff
@@ -395,6 +414,23 @@ export default {
...
+      if (
+        event.key === '/'
+        || (event.key === 'o' && event.shiftKey && (event.metaKey || event.ctrlKey))
+      ) {
+        this.openQuickNavigationModal();
+        event.preventDefault();
+      }
+    },
+    handleFocus(event) {
+      if (event.target.className === 'filter__quick-navigation-container') return;
```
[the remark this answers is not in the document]
sofiaromorales: Removed here https://github.com/apple/swift-docc-render/pull/396/commits/155ef513b18705e57b563e8ec00719141ec3c00b

sofiaromorales:
Hi @marinaaisa, appreciate all your comments 😄

I addressed all the suggested changes from the last review, I had to make some minor changes to `FilterInput` to make it more generic and reuse it for the QN modal, but otherwise, it has made the component way cleaner.

+Some improvements to the `flattenIndex` object, and after testing with the SwiftUI index file seems like the performance problem no longer happens. 

I did some UI/UX changes with the feedback I received from the forums pitch, including reducing the debouncing time for faster results, resolving issues #410, #411, and #412, and moving the open modal button `/` to the navigation bar.

I will keep iterating on the`QuickNavigationBar` component, but I would appreciate it if you could give this a round of code review to see if we can merge this, thanks a lot!

src/views/DocumentationTopic.vue
```diff
@@ -61,7 +70,15 @@
             :isWideFormat="enableNavigator"
             :sidenavHiddenOnLarge="sidenavHiddenOnLarge"
             @toggle-sidenav="handleToggleSidenav"
-          />
+          >
+            <template v-slot:menu-items>
```
marinaaisa, line 74: You can write it like
```suggestion
            <template #menu-items>
```
too

src/views/DocumentationTopic.vue
```diff
@@ -9,10 +9,19 @@
...
-  <div :class="{ 'modal-open': quickNavigationStore.state.showQuickNavigation }">
-    <div v-show="quickNavigationStore.state.showQuickNavigation">
-      <QuickNavigationModal />
-    </div>
+  <div>
+    <PortalTarget name="modal-destination" multiple />
+    <GenericModal
+      codeBackgroundColorOverride="transparent"
+      theme="code"
+      :isFullscreen="true"
```
marinaaisa, line 17: You can write it like
```suggestion
      isFullscreen
```
too

src/components/Filter/FilterInput.vue
```diff
@@ -395,12 +400,14 @@ export default {
...
     },
+    handleFocus() {
+      this.showSuggestedTags = true;
+    },
   },
   created() {
     if (
       this.focusInputWhenCreated
       && document.activeElement !== this.$refs.input
-      && this.inputIsNotEmpty
```
marinaaisa, line 403: This change potentially breaks other use cases for users who want the input to focus only when it's not empty.
Could you create another prop that make sure that we focus input when it's empty too?
Something like `focusInputWhenEmpty` and then the condition would be:

`this.inputIsNotEmpty && !focusInputWhenEmpty`

Thinking out loud. You may find a better way to do it :)

src/components/Filter/FilterInput.vue
```diff
@@ -76,7 +76,7 @@
               @keydown.left="leftKeyInputHandler"
               @keydown.right="rightKeyInputHandler"
               @keydown.delete="deleteHandler"
-              @keydown.meta.a.prevent="selectInputAndTags"
+              @keydown.meta.a.prevent.stop="selectInputAndTags"
```
marinaaisa, line 79: Why are you stopping the event propagation?

marinaaisa (changes requested):
Hi Sofía!
Your work is looking quite well, thanks for the hard work!
I've added some new comments for you to check. :)

Also I've found some AX issues that we could improve:
1- The circle on the X button seems off:
Screenshot 2022-08-29 at 21 17 27

2- Right now 2 tabs are needed to tab through the list, we should make it only 1.
https://user-images.githubusercontent.com/8567677/187281586-cd3f4fbf-baaf-470b-84b1-776517b7566a.mov


 I know you had some new UI updates on the way, I'll review them once you push them!
Thanks!

src/components/Filter/FilterInput.vue
```diff
@@ -14,9 +14,9 @@
     role="search"
     tabindex="0"
     :aria-labelledby="searchAriaLabelledBy"
-    :class="{ 'focus': showSuggestedTags }"
+    :class="{ 'focus': showSuggestedTags && !preventFocusStyle }"
```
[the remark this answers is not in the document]
sofiaromorales: @marinaaisa this would work in this case, but what about use cases where you want to keep the focus even when there are no tags? I think that we are tying up two unrelated things (the existence of tags with the focus logic)

src/components/Navigator.vue
```diff
@@ -41,6 +41,7 @@
 </template>
 
 <script>
+
```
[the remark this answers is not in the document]
sofiaromorales: Resolved in [54a559e](https://github.com/apple/swift-docc-render/pull/396/commits/54a559ef81ba62970191dfba7f9be5fd92e3a6a6)

src/views/DocumentationTopic.vue
sofiaromorales: Resolved in [54a559e](https://github.com/apple/swift-docc-render/pull/396/commits/54a559ef81ba62970191dfba7f9be5fd92e3a6a6)

src/views/DocumentationTopic.vue
sofiaromorales: Resolved in [54a559e](https://github.com/apple/swift-docc-render/pull/396/commits/54a559ef81ba62970191dfba7f9be5fd92e3a6a6)

src/components/Filter/FilterInput.vue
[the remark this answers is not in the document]
sofiaromorales: Ended up removing the prop and leaving the focus style on the element ([bedd4e0](https://github.com/apple/swift-docc-render/pull/396/commits/bedd4e06f51824468835b932b2a9640aad3afbd3))

src/components/Navigator.vue
```diff
@@ -162,7 +159,7 @@ export default {
     }) => {
       const flatIndex = flattenNestedData(technology.children || [], null, 0, technology.beta);
       if (enableQuickNavigation) {
-        store.setFlattenIndex(flatIndex);
+        store.setFlattenIndex(Object.freeze(flatIndex));
```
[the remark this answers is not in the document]
sofiaromorales: Thanks for the suggestion, I went with this approach and it seems to work fine. I removed the `QuickNavigationStore` entirely, moved the `flatIndex` logic into the `NavigatorDataProvider`, and changed the tests accordingly. Thanks Marina! ([bedd4e0](https://github.com/apple/swift-docc-render/pull/396/commits/bedd4e06f51824468835b932b2a9640aad3afbd3))

src/components/Filter/FilterInput.vue
sofiaromorales: [54a559e](https://github.com/apple/swift-docc-render/pull/396/commits/54a559ef81ba62970191dfba7f9be5fd92e3a6a6)

src/components/Filter/FilterInput.vue
sofiaromorales: The event in the QN filter input is bubbling up and selecting the entire DOM in `cmd+a` instead of just the input content, that's why I'm stopping the event propagation.

marinaaisa (changes requested):
Looking quite good! I like it :)

I left you a comment about a wrong SVG that I fixed so you don't have strange margins on it.

I will try to get the rest of the team to review this PR too, since it's quite big.

Thanks, Sofía, have a nice weekend!

marinaaisa:
I also see that you still have to fix some failing tests, let me know if I can help with that.

tests/unit/components/Navigator/__snapshots__/NavigatorDataProvider.spec.js.snap
```diff
@@ -0,0 +1,160 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`NavigatorDataProvider counts the amount of deprecated items a groupMarker has 1`] = `
```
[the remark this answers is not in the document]
sofiaromorales: It got autogenerated when I was moving some tests from `Navigator` to `NavigatorDataProvider`

sofiaromorales:
> I also see that you still have to fix some failing tests, let me know if I can help with that.

Which test is failing? I just merged main and tested again but all seem working well for me

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
+            >
+              <MagnifierIcon />
+            </div>
+          </template>
+          <template #input-menu-items>
+            <div
+              v-if="!userInput.length"
+              class="quick-navigation__open-key-container"
+            >
+              <kbd>Cmd+Shift+o</kbd>
```
mportiz08, line 47: This text was a little confusing for me when I was trying this feature out since it wasn't clear that this a the keyboard shortcut for opening the modal since there was just the text on its own. You have to close the modal before the shortcut does anything, which is why it was a little confusing. It was also a little unclear to me why there were multiple shortcuts that could be used—this one and the `/` key. Also, the `cmd` modifier key might be specific to keyboards used in Apple platforms.

Minor detail. And TIL about the `<kbd>` element :P

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
-              {{ symbol.title.slice(symbol.start + symbol.matchLength) }}
-            </span>
-          </Reference>
+          <div
+            v-if="noResultsWereFound"
+            class="no-results"
+          >
+            <p>
+              No results found.
+          </p>
```
mportiz08, line 61:
```suggestion
            </p>
```

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
+                        :matcherText="debouncedInput"
+                      />
+                      <span v-text="symbol.title.slice(symbol.start + symbol.matchLength)">
+                      </span>
+                    </div>
+                  </div>
+                  <div class="symbol-path">
+                    <div
+                      v-for="(parent, index) in symbol.parents"
+                      :key="index"
```
mportiz08, line 101: Possibly the title might make for a better `key` here.

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
+          >
+            <p>
+              No results found.
+          </p>
+          </div>
+          <div
+            v-for="(symbol, index) in filteredSymbols"
+            :class="{ 'selected' : index == focusedIndex }"
+            :key="index"
+            :id="MatchId"
```
mportiz08, line 67: This `MatchId` constant is getting assigned to multiple div elements, and all `id` attributes should be unique, so we should consider changing this or dropping it if it isn't needed.

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
+            class="no-results"
+          >
+            <p>
+              No results found.
+          </p>
+          </div>
+          <div
+            v-for="(symbol, index) in filteredSymbols"
+            :class="{ 'selected' : index == focusedIndex }"
+            :key="index"
```
mportiz08, line 66: The symbol title might possibly make for a better key here.

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
+    },
+    startingPointHook() {
+      this.focusedIndex = this.totalItemsToNavigate - 1;
+    },
   },
 };
 </script>
 
-<style scoped lang="scss">
+<style lang="scss">
```
mportiz08, line 275: Is there a specific reason why you removed the `scoped` attribute for these styles? It can be a little dangerous since these styles might now interfere with other components on the page that might contain the same classnames (even though it does seem like the majority here are very specific).

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
+    border: solid $base-border-width var(--color-fill-gray);
+    border-radius: $border-radius;
+    filter: drop-shadow(0px 7px 50px rgba(0, 0, 0, 0.25));
+    margin: auto;
+    max-width: rem(800px);
+    .filter__wrapper {
+      border: none;
+    }
+    > * {
+      --input-text: var(--color-figure-gray-secondary) !important;
```
mportiz08, line 297: Have you looked into other ways of making this work without the need for the `!important`?

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
 -->
 
 <template>
-  <div class="quick-navigation-modal">
-    <div class="container">
+  <div
```
mportiz08, line 12: You may have already discussed this with Marina, but is there a reason why this modal seems to have been written mostly from scratch and doesn't utilize the existing `GenericModal` component? (There's likely a good reason, just curious)

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
+    .no-results {
+      margin: $filter-padding auto $filter-padding auto;
+      width: fit-content;
+    }
+    .selected {
+      background-color: var(--color-navigator-item-hover);
+    }
+  }
+  &__modal-shadow {
+    background-color: var(--color-quick-navigation-modal-shadow);
```
mportiz08, line 337: We may want to see what colors the existing `GenericModal` component uses for this to reduce the number of component-specific colors we have (and either use that one or change them both to use this color).

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
     position: fixed;
-    top: 0;
     bottom: 0;
     left: 0;
     right: 0;
-    z-index: 100;
-    padding: 100px;
-    overflow: scroll;
+    top: 0;
+    padding: min(rem(10px));
```
mportiz08, line 343: Is `min()` needed here?

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,387 @@
...
+    margin-top: auto;
+    margin-bottom: auto;
+    margin-right: rem(15px);
+    color: var(--color-figure-gray-secondary);
   }
-  .symbol-match {
-    width: fit-content;
+  &__reference:hover {
+      text-decoration: none;
+    }
```
mportiz08, line 354:
```suggestion
  }
```

mportiz08 (commented):
Very cool! I played around with the feature and found it to be helpful and very easy to use.

I also did a quick pass at reviewing most of the code, and I mainly had minor comments from what I looked at. I might want to do a closer look at the `DocumentationTopic` view changes since those are fairly big and could impact pages even without the feature enabled, but things seem to generally be looking/working like normal without the feature enabled so far.

Thanks for all your hard work on this!

src/components/Filter/FilterInput.vue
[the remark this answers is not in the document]
marinaaisa: Nice! I think it's better in terms of AX :)

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: Resolved in https://github.com/apple/swift-docc-render/pull/396/commits/62db4a637515e8fd67f83266a028ced48f825d01

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: Resolved in https://github.com/apple/swift-docc-render/pull/396/commits/62db4a637515e8fd67f83266a028ced48f825d01

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: Removed in https://github.com/apple/swift-docc-render/pull/396/commits/62db4a637515e8fd67f83266a028ced48f825d01

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: I removed it to style the inner `Filter Input` component, but it doesn't seem necessary anymore with the latest changes, thanks! Added in https://github.com/apple/swift-docc-render/pull/396/commits/62db4a637515e8fd67f83266a028ced48f825d01

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: I ended up removing this constant used for custom logic inside `handleKeyEnter` and added an enter event modifier directly to `FilterInput` to handle the case when the clear x button is selected and the event is propagated to the parent`QuickNavigationModal` https://github.com/apple/swift-docc-render/pull/396/commits/25e08fea8973a787e881a5f7399d2e4e2be2dfff

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: I'm using `GenericModal` as a wrapper for `QuickNavigationModal` and using most of its logic, I made some refactor here [3aaf55f](https://github.com/apple/swift-docc-render/pull/396/commits/3aaf55f1359ecdcd482ad389caa06ae36ed19c88) to enclose everything in one component and make it fit my use case better, there's anything specific that is not being reused properly? 🤔

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: I agree that having the shortcut as part of the input text could be confusing, I'm going to remove it from the input text and try to find a better place for it, if any, in another iteration. Regarding the `cmd` key, AFAIK using `metaKey` works for both Mac and Windows. The idea behind having both shortcuts is that `/` is very common for opening similar tools on websites, and `cmd+shift+o` mimics Xcode's Open Quickly shortcut. Thanks for pointing this out, Marcus! It's been a bit tricky to figure out how to make the feature visible to the user.

dobromir-hristov: The `metaKey` aka `Win` key on Windows is almost never used in browser shortcuts, as its OS specific. Its more common to use `ctrl`, same for Linux I think.

src/components/Navigator/NavigatorCard.vue
```diff
@@ -245,17 +236,14 @@ export default {
...
-      default: false,
-    },
     allowHiding: {
       type: Boolean,
       default: true,
     },
   },
   mixins: [
     keyboardNavigation,
+    symbolTreeNavigator,
```
dobromir-hristov, line 246: I understand why you extracted those into a mixin, but now this component looks like its missing an important functionality. Mixins are hard to reason with, as they can hold hidden functionality, which is good and bad at the same time.

Since the navigator dataset is now used in other places, I think it makes more sense to extract all of these helper functions like `getParent`, `getSiblings` and so on into their own file, so they are widely available and easy to inject into other components, as long as they have access to the dataset.

src/components/Filter/FilterInput.vue
dobromir-hristov: Why doesnt it do that in other places we use the input though 🤔

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,368 @@
...
+          >
+            <p>
+              No results found.
+            </p>
+          </div>
+          <div
+            v-for="(symbol, index) in filteredSymbols"
+            :class="{ 'selected' : index == focusedIndex }"
+            :key="symbol.uid"
+            tabindex="0"
```
dobromir-hristov, line 69: Why this complication of adding tabindex to a div and not use the Reference's natural focusability, since its a link?

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,368 @@
...
+                role="list"
+              >
+                <div class="symbol-info">
+                  <div class="symbol-name">
+                    <NavigatorLeafIcon
+                      class="navigator-icon"
+                      :type="symbol.type"
+                    />
+                    <div class="symbol-title">
+                      <span v-text="symbol.title.slice(0, symbol.start)"></span>
```
dobromir-hristov, line 90: Calling native methods like this directly in the template can lead to some weird rendering issues sometimes, just wanted to point it out. Its a Vue thing, I have noticed over the years.

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,368 @@
...
+                      <QuickNavigationHighlighter
+                        :text="symbol.substring"
+                        :matcherText="debouncedInput"
+                      />
+                      <span v-text="symbol.title.slice(symbol.start + symbol.matchLength)">
```
dobromir-hristov, line 95: This whole thing feels like a great candidate for a component of its own.

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,368 @@
...
+    totalItemsToNavigate: ({ filteredSymbols }) => filteredSymbols.length,
+  },
+  watch: {
+    userInput: 'debounceInput',
+    focusedIndex: 'scrollIntoView',
   },
   methods: {
     closeQuickNavigationModal() {
-      this.quickNavigationStore.toggleShowQuickNavigationModal();
+      this.$emit('closeQuickNavigationModal', false);
```
dobromir-hristov, line 198: Maybe just leverage the `:showQuickNavigationModal.sync` syntax?

```suggestion
      this.$emit('update:showQuickNavigationModal', false);
```

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,368 @@
...
         return 0;
       });
     },
-  },
-  watch: {
-    userInput: 'debounceInput',
+    scrollIntoView() {
+      this.$refs.match[this.focusedIndex].scrollIntoView({
+        block: 'nearest',
+        inline: 'start',
```
dobromir-hristov, line 267: These dont work in safari I think, just so you are aware.

src/views/DocumentationTopic.vue
```diff
@@ -9,28 +9,42 @@
...
-            <div class="doc-topic-aside">
-              <NavigatorDataProvider
-                :interface-language="topicProps.interfaceLanguage"
-                :technology="technology"
-                :api-changes-version="store.state.selectedAPIChangesVersion"
-              >
-                <template #default="slotProps">
+  <CodeTheme class="doc-topic-view">
+    <template v-if="topicData">
+      <NavigatorDataProvider
```
dobromir-hristov, line 14: I could be wrong, but having the `NavigatorDataProvider` in two places, would initiate its logic twice, fetching the dataset, storing that huge piece of data twice, all the calculation and so on with it.

src/views/DocumentationTopic.vue
```diff
@@ -42,39 +56,50 @@
...
+          :isSymbolDeprecated="isSymbolDeprecated"
+          :isSymbolBeta="isSymbolBeta"
+          :currentTopicTags="topicProps.tags"
+          :references="topicProps.references"
+          :isWideFormat="enableNavigator"
+          :sidenavHiddenOnLarge="sidenavHiddenOnLarge"
+          @toggle-sidenav="handleToggleSidenav"
+        >
+          <template #menu-items>
+            <div
```
dobromir-hristov, line 80: Why use a div if you want a clickable element? Just use a `<button>` ?

src/views/DocumentationTopic.vue
```diff
@@ -319,13 +350,22 @@ export default {
         this.toggleMobileSidenav();
       }
     },
+    openQuickNavigationModal() {
```
dobromir-hristov, line 353: I feel like all of the logic related to `showQuickNavigationModal`, like listeners and all, can be inside the `<QuickNavigationModal>`, so its contained and does not spill into the main `DocumentationTopic.vue` component

dobromir-hristov (commented):
Did a quick review of the code. I havent tried the feature yet, will do that a bit later.

src/views/DocumentationTopic.vue
marinaaisa: Mmm, that could be true.  If we are going to share the same data and we don't want to re-use the component because performance issues, what is the point of having a NavigatorDataProvider component?

We could move all that logic to the DocumentationTopic component and send that data to the child `QuickNavigationModal` and `Navigator` components

marinaaisa: We could just add `NavigatorDataProvider` on the root of `DocumentationTopic` so the data is available for `QuickNavigationModal` and `Navigator` components. Does it sound right, @sofiaromorales ?

src/components/Filter/FilterInput.vue
```diff
@@ -93,11 +93,13 @@
             aria-label="Reset Filter"
             class="filter__delete-button"
             @click="resetFilters(true)"
+            @keydown.enter.exact.stop="resetFilters(true)"
```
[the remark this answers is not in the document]
sofiaromorales: `@click` does get fired on `keydown.enter`, but adding this will explicitly prevent collisions with `enter` events from parent components. 

I had the problem that `@keydown.enter.exact="handleKeyEnter"` on `QuickNavigationModal` was getting called instead of `resetFilters`. When `FilterInput` was used only by `NavigatorCard` this wasn't a problem, but now is being reused, and having this line allows to keep the intended behavior no matter what the parent component `enter` event modifier is, making the component more generic IMO.

src/components/GenericModal.vue
```diff
@@ -223,6 +232,7 @@ export default {
         this.selectContent();
       }
       if (key !== 'Escape') return;
+      event.preventDefault();
```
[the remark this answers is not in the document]
sofiaromorales: On Safari clicking the `Escape`key exits the full-screen mode, and I would like to avoid this when the modal is open, reason why I'm preventing the default behavior.

src/components/Navigator/NavigatorCard.vue
sofiaromorales: Thanks for pointing this out! I was extracting only the functions I needed on the other component, but it does make sense to extract all others (`getAllChildren` `getSiblings` `getChildren`) `getParents` is already on the file. Do you have any preference on how we should name this mixin?

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: I'm probably missing something here, but having the div tag wrapping `Reference` makes it necessary to use `tabindex="0"`. I tried to remove this external div and loop over `Reference` directly, but then the `@@focus.capture="focusIndex(index)"` logic doesn't get triggered.

src/components/Navigator/QuickNavigationModal.vue
mportiz08: Got it. I think I just missed that. Apologies.

src/components/Navigator/QuickNavigationModal.vue
dobromir-hristov: Because Reference is a Vue component, try with `@focus.native`

src/components/Navigator/NavigatorCard.vue
dobromir-hristov: I generally am not a fan of Mixins, as they are hard to reason with. You could use a mixin on just export a bunch of methods from a file, and could call these `NavigatorIndexUtils`.

sofiaromorales: I like the utils file approach, but since all the functions depend on a computed property, is this not an indicator that it should be converted into a mixin? I don't have a strong preference for either of them, just wondering since both seem to be very similar solutions

src/components/Filter/FilterInput.vue
sofiaromorales: Because `GenericModal` overwrites the behavior of cmd+a to select the content of the modal

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: I think that the MDN docs browser compatibility section for `Element.scrollIntoView()` is outdated (https://developer.mozilla.org/es/docs/Web/API/Element/scrollIntoView).  For me is working as expected in Safari 15.3, and there are multiple issues regarding this.
Ex:
https://github.com/mdn/browser-compat-data/issues/17542
https://github.com/mdn/browser-compat-data/issues/9055

src/views/DocumentationTopic.vue
sofiaromorales: `openQuickNavigationModal` and `showQuickNavigationModal` has to be accessed outside the `QuickNavigationModal` component to open the modal, like in the top navbar and the keydown event listener. Having everything contained in a single component would be ideal, but I think that is not possible with the current implementation.

src/components/Navigator/NavigatorCard.vue
dobromir-hristov: Hmmm I see what you mean... Its more annoying to refactor those as props, so yeah, I think a mixin should be fine in that case.

src/components/Navigator/QuickNavigationModal.vue
dobromir-hristov: Alright then :)

src/views/DocumentationTopic.vue
```diff
@@ -319,13 +350,22 @@ export default {
...
+      this.showQuickNavigationModal = true;
+    },
     toggleLargeSidenav(value = !this.sidenavHiddenOnLarge) {
       this.sidenavHiddenOnLarge = value;
       storage.set(NAVIGATOR_HIDDEN_ON_LARGE_KEY, value);
     },
     toggleMobileSidenav(value = !this.sidenavVisibleOnMobile) {
       this.sidenavVisibleOnMobile = value;
     },
+    onKeydown(event) {
```
dobromir-hristov, line 363: Maybe make this name relate to quickNavigation a bit more?

src/views/DocumentationTopic.vue
dobromir-hristov: Oh right right....

src/components/GenericModal.vue
```diff
@@ -294,6 +305,11 @@ $modal-close-color: light-color(figure-gray-tertiary) !default;
       @include prefers-dark {
         background: rgb(29, 29, 31);
       }
+      &.pin-content {
+        padding: 0;
+        top: $pinned-modal-top;
+        position: fixed;
```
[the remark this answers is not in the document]
sofiaromorales: The styling of the generic modal wasn't flexible enough to fit the Quick navigation modal design. That is why I added that a custom class at first, but a way to make it even more generic would be passing a styling prop as I did here 2e5a920bde8d099ba241252297ea6cdc585b2755. Not allowing only a custom width but also more CSS props. I haven't seen this pattern anywhere else on the codebase tho, so not sure if it's the best idea, @marinaaisa any thoughts on this?

src/components/Navigator/NavigatorCard.vue
sofiaromorales: The refactor was straightforward https://github.com/apple/swift-docc-render/pull/396/commits/0e1f5cff7caf0fe87ec9fdec752ad3b33c72f192. We're getting rid of the mixin, but now each method requires extra parameters that before would be coming from the computed properties, both ways seem fine to me

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: Thanks for pointing this out! Resolved here 4bd2fd13e893ecb2b3dee9617a6ba7e430882f48

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: Resolved 4bd2fd13e893ecb2b3dee9617a6ba7e430882f48

src/views/DocumentationTopic.vue
sofiaromorales: What Dobri mentioned is correct, I made the refactor here f905ee92e646c01349cd2f59abd67761204e68f2. Two tests were failing after moving things around, and I can't tell for sure why, but I resolved it by removing `await flushPromises()` from both. 

Thanks, Marina, for proposing the solution :)

src/views/DocumentationTopic.vue
sofiaromorales: Resolved 4bd2fd13e893ecb2b3dee9617a6ba7e430882f48

tests/unit/components/Navigator/__snapshots__/NavigatorDataProvider.spec.js.snap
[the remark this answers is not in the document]
sofiaromorales: They will be auto-generated every time we run the tests, just like `Navigator.spec.js.snap`

src/components/Icons/ClearRoundedIcon.vue
```diff
@@ -26,3 +26,9 @@ export default {
   components: { SVGIcon },
 };
 </script>
+
+<style scoped lang="scss">
+  path {
+    transform: translate(5%, 5%)
```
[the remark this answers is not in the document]
sofiaromorales: Thanks Marina! https://github.com/apple/swift-docc-render/pull/396/commits/2dfc53ebfb0b2238cf1faa6bb768e9fbd1ab5b7f

src/components/Filter/FilterInput.vue
```diff
@@ -79,6 +79,7 @@
               @keydown.meta.a.prevent.stop="selectInputAndTags"
               @keydown.ctrl.a.prevent="selectInputAndTags"
               @keydown.exact="inputKeydownHandler"
+              @keydown.191.exact.stop="inputKeydownHandler"
```
dobromir-hristov, line 82: I am not sure the FilterInput should handle specially the `/` key, if I am getting this correct.

src/utils/navigatorIndex.js
```diff
@@ -0,0 +1,96 @@
...
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import { INDEX_ROOT_KEY } from 'docc-render/constants/sidebar';
+
+// eslint-disable-next-line import/prefer-default-export
```
dobromir-hristov, line 13: lets remove this line and add proper JSDoc like the rest.

src/views/DocumentationTopic.vue
dobromir-hristov: One drawback is that now the `NavigatorDataProvider` is always rendered and tries to fetch the navigator data, but DocCRender may be served older `RenderJSON`, which may not have the navigator data.

I think we may need to introduce some conditionals based on `enableNavigator`.

src/components/Navigator/NavigatorCard.vue
dobromir-hristov: yeah those are fine i think.

marinaaisa (commented):
Hi Sofia! Thanks for the updates :)

I gave it another look and I found this UI issue:
Focus style for the FilterInput in the QuickNavigation should be the same as in the Navigator

QuickNavigaton:
Screenshot 2022-09-16 at 19 18 26

Navigator:
Screenshot 2022-09-16 at 19 09 41

src/components/Filter/FilterInput.vue
sofiaromorales: The reason behind this change was to stop the propagation of the event and prevent the `onKeydown` from `DocumentationTopic` from getting called. Without this, the modal gets opened when the user tries to type `/` on the filter input. But since you mention it, there's probably a better way to solve this.

sofiaromorales:
> One drawback is that now the NavigatorDataProvider is always rendered and tries to fetch the navigator data, but DocCRender may be served older RenderJSON, which may not have the navigator data.
I think we may need to introduce some conditionals based on enableNavigator.

@marinaaisa  @dobromir-hristov I see how this can be a drawback, but this would end up in an error caught by the `NavigatorDataProvider`, so the app functionality is not affected. I could open a new issue and work on this in a different pr, or do you think this is a blocker for merging this pr?

src/views/DocumentationTopic.vue
sofiaromorales: @marinaaisa @dobromir-hristov I see how this can be a drawback, but this would end up in an error caught by the NavigatorDataProvider, so the app functionality is not affected. I could open a new issue and work on this in a different pr, or do you think this is a blocker for merging this pr?

src/views/DocumentationTopic.vue
sofiaromorales: This method gets called on any key down, so it doesn't have to be related to Quick navigation necessarily, I've been trying to come up with a different (but generic) name but haven't come up with any yet 🤔

src/utils/navigatorIndex.js
sofiaromorales: Resolved 19f92322e60a5702a0c09780cdc14da1b22327b4

src/utils/navigatorIndex.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
```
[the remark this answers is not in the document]
sofiaromorales: Resolved here https://github.com/apple/swift-docc-render/commit/19f92322e60a5702a0c09780cdc14da1b22327b4

marinaaisa (changes requested):
The "X" button on the right is now too big in the FilterInput component.

I will ask Bea to provide feedback on what size and padding should have in both cases.

Screenshot 2022-09-21 at 13 13 35

Screenshot 2022-09-21 at 13 19 34

biamx3:
Hi! The X should be the same size as in the original filter component for consistency. The padding to the right of the X should be the same as to the left of the magnifying glass icon. The padding to the left of the X should be the same as the padding between the text input and magnifying glass icon.

sofiaromorales:
Thanks for the clarification Bea! the size of the X icon was changed from 7px to 16px in a different pr and is the size that is currently on the main branch, I'll restore it to its original size then @marinaaisa

src/components/Filter/FilterInput.vue
sofiaromorales: Resolved [dccfec7](https://github.com/apple/swift-docc-render/pull/396/commits/dccfec7956dd3d7b98fb0138097aaa05f31681e7)

src/views/DocumentationTopic.vue
sofiaromorales: Resolved [dccfec7](https://github.com/apple/swift-docc-render/pull/396/commits/dccfec7956dd3d7b98fb0138097aaa05f31681e7)

sofiaromorales:
Now that the feature flag is configured directly from docc I published a new URL with the latest changes of this pr https://sofiaromorales.github.io/SwiftDocC/documentation/swiftdocc

src/views/DocumentationTopic.vue
dobromir-hristov: But it will error in the user's browser. We can prevent that. 

Also your entire functionality depends on this data being there. Not sure what would happen if its missing.

src/components/Navigator/NavigatorDataProvider.vue
```diff
@@ -83,6 +97,81 @@ export default {
...
       }
     },
+    /**
+     * @param {{path: string, type: string, title: string, children?: [] }[]} childrenNodes
+     * @param {NavigatorFlatItem | null} parent
+     * @param {Number} depth
+     * @param {Boolean} parentBetaStatus
+     * @return {NavigatorFlatItem[]}
+     */
+    flattenNestedData(childrenNodes, parent = null, depth = 0, parentBetaStatus = false) {
```
dobromir-hristov, line 107: Lets move the `flattenNestedData` and `hashCode` into the utils. I think this will help thirdparty clients create their own data providers.

src/components/Navigator/NavigatorDataProvider.vue
```diff
@@ -45,6 +47,18 @@ export default {
...
   },
   computed: {
+    /**
+     * Recomputes the list of flat children.
+     * @return NavigatorFlatItem[]
+     */
+    flatChildren: ({
+      flattenNestedData, technologyWithChildren = {},
+    }) => {
+      const flatIndex = flattenNestedData(
```
dobromir-hristov, line 57: No need for a variable here, just return it.

marinaaisa:
I still see some regressions on the FilterInput

We should delete the padding on the `clear-rounded-icon` because it adds an extra 5px into the view.
Also the `clear-rounded-icon` elements looks visually much bigger than the `filter__filter-button` element so I would reduce its width to 12px.

Currently:
Screenshot 2022-09-27 at 17 55 28

Proposed changes:
Screenshot 2022-09-27 at 17 55 21

Also there is an extra left spacing on the `quick-navigation__magnifier-icon-container` that we may want to remove too.
Screenshot 2022-09-27 at 17 58 31

src/views/DocumentationTopic.vue
sofiaromorales: I did a refactor to prevent rendering `NavigatorDataProvider` when there's no available index data using the existing logic based on the same piece of logic that conditionally renders the `Navigator` +  I added logic for enabling opening quick navigation depending on `enableNavigator` too. Thx for the suggestion Dobri! [4038fea](https://github.com/apple/swift-docc-render/pull/396/commits/4038fead95b38d39762d9113dd7bc91279ded66a)

src/components/Navigator/NavigatorDataProvider.vue
sofiaromorales: Resolved [c17ff68](https://github.com/apple/swift-docc-render/pull/396/commits/c17ff680b73bf74237d8cf89054fcb7cd651b07a)

src/components/Navigator/NavigatorDataProvider.vue
sofiaromorales: Resolved [c17ff68](https://github.com/apple/swift-docc-render/pull/396/commits/c17ff680b73bf74237d8cf89054fcb7cd651b07a)

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,366 @@
...
 
 <template>
-  <div class="quick-navigation-modal">
-    <div class="container">
+  <GenericModal
+    :showClose="false"
+    :visible="showQuickNavigationModal"
+    :isFullscreen="true"
+    :styleProps="modalStyleProps"
+    @update:visible="$emit('update:showQuickNavigationModal', false)"
```
marinaaisa, line 17: You could replace these 2 lines using the syntax `.sync` as we do in other parts of the code.

So instead of writing

```js
:visible="showQuickNavigationModal"
@update:visible="$emit('update:showQuickNavigationModal', false)"
```

you would write `:visible.sync="isVisible"` in the template and then create a computed property of `isVisible` with a get and a set

```js
isVisible: {
  get: ({ showQuickNavigationModal }) => showQuickNavigationModal,
  set(value) {
    this.$emit('update:showQuickNavigationModal', value);
  },
},
```

src/views/DocumentationTopic.vue
```diff
@@ -9,74 +9,88 @@
...
+            :technology="technology"
+            :api-changes-version="store.state.selectedAPIChangesVersion"
+            ref="NavigatorDataProvider"
+          >
+            <template #default="slotProps">
+              <div class="doc-topic-aside">
+                <QuickNavigationModal
+                  :children="slotProps.flatChildren"
+                  :showQuickNavigationModal="showQuickNavigationModal"
+                  @update:showQuickNavigationModal="(show) => showQuickNavigationModal = show"
```
marinaaisa, line 32: Same in here, these 2 lines can be replaced by just:
```suggestion
                  :showQuickNavigationModal.sync="showQuickNavigationModal"
```

src/components/GenericModal.vue
```diff
@@ -87,6 +91,10 @@ export default {
       type: Boolean,
       default: true,
     },
+    styleProps: {
```
marinaaisa, line 94: I'm not a fan of adding styles through props, because that's what CSS classes are for instead.
I know that the width is a prop at the moment but overwriting more styles through props doesn't seem an efficient idea.
I suggest you overwrite those styles applying classes using the `/deep/` element

sofiaromorales: Thanks for the suggestion Marina 🙏 I was unaware of the `/deep/` selector. It doesn't feel right to have the style scoped and still mess with the child styling, but I agree that's a cleaner solution than passing it as a prop. Resolved [2fa4743](https://github.com/apple/swift-docc-render/pull/396/commits/2fa474366a821887f60369e3d6806267aac7abac).

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,366 @@
 -->
 
 <template>
-  <div class="quick-navigation-modal">
-    <div class="container">
+  <GenericModal
+    :showClose="false"
+    :visible="showQuickNavigationModal"
+    :isFullscreen="true"
```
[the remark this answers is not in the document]
sofiaromorales: Resolved [0b98f4c](https://github.com/apple/swift-docc-render/pull/396/commits/0b98f4c4a9201ff5868f778dc9a0770905b1eba2)

src/views/DocumentationTopic.vue
sofiaromorales: Resolved [0b98f4c](https://github.com/apple/swift-docc-render/pull/396/commits/0b98f4c4a9201ff5868f778dc9a0770905b1eba2)

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,349 @@
...
+    > * {
+      width: 100%;
+    }
+    &.blue .magnifier-icon {
+      fill: var(--color-figure-blue);
+      color: var(--color-figure-blue);
+    }
+  }
+  &__match-list {
     overflow: scroll;
```
marinaaisa, line 306: In Chrome, with an macOS system config of "Show scroll bar... Automatically"

This overflow scroll is creating a jumping effect and also an extra padding in the bottom when there is no results.

https://user-images.githubusercontent.com/8567677/197261667-d3c58eef-3c4e-4393-ad88-818285b6f020.mov

tests/unit/utils/indexData.spec.js
```diff
@@ -0,0 +1,319 @@
...
+
+import {
+  convertChildrenArrayToObject,
+  flattenNestedData,
+  getAllChildren,
+  getChildren,
+  getParents,
+  getSiblings,
+  hashCode,
+} from 'docc-render/utils/indexData';
```
marinaaisa, line 19: I was thinking that maybe we could consider replacing the name of "indexData" with "NavigatorData". What do you think?

marinaaisa (changes requested):
Hey @sofiaromorales , I left couple of more comments.
I'll continue reviewing next week, hopefully we get to merge it soon, sorry for the delay. It's taking a long time but the progress and feature is big and amazing! Thanks again! :)

tests/unit/components/Navigator/QuickNavigationModal.spec.js
```diff
@@ -9,73 +9,301 @@
...
+        type: '',
+      },
+    ];
+    wrapper = shallowMount(QuickNavigationModal, {
+      propsData: {
+        children: customSymbols,
+        showQuickNavigationModal: true,
+      },
+      provide: {
+        quickNavigationStore: {},
```
marinaaisa, line 262: We should delete this one, right?

marinaaisa:
Just a quick UI detail:

Whenever we have the focus on the Quick Navigation's filter, we may want to delete the border line on top of the match list. To achieve this you can add this line of CSS

```
.quick-navigation__filter.focus + .quick-navigation__match-list {
  border-top: 0;
}
```

tests/unit/utils/indexData.spec.js
sofiaromorales: It makes sense since most of these methods are used in the `NavigatorCard` component

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,349 @@
...
       ), '');
     },
     debounceInput: debounce(function debounceInput(value) {
-      this.debouncedInput = value;
-    }, 500),
-    fuzzyMatch: ({ debouncedInput, symbols, processedInputRegex }) => (
-      symbols.map((symbol) => {
-        const match = processedInputRegex.exec(symbol.title.toLowerCase());
+      // Remove space-character
+      this.debouncedInput = value.replace(/\s/g, '');
```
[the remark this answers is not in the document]
sofiaromorales: I made this refactor to extract that logic into another property 39f81c6db0f2f163ca067e5156e8c458af3f521e

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -9,180 +9,349 @@
...
         });
-      }).filter(Boolean)
-    ),
+      }).filter(Boolean);
+    },
+    handleKeyEnter() {
+      if (
+        this.noResultsWereFound
+        || !this.userInput.length
+      ) return;
```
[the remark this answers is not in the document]
sofiaromorales: Resolved here e39a36f46e04ffc43c726b27d5dda242e891dc21

tests/unit/utils/indexData.spec.js
sofiaromorales: Resolved e39a36f46e04ffc43c726b27d5dda242e891dc21

tests/unit/components/Navigator/QuickNavigationModal.spec.js
sofiaromorales: yess e39a36f46e04ffc43c726b27d5dda242e891dc21

sofiaromorales:
Thanks a lot for the review and patience @marinaaisa! Do you think it's safe to merge this already? I want to give people the opportunity to test quick navigation in their documentation, this way I can receive feedback and discover new bugs if any, and since it's under a feature flag, it will not affect the current functionality of the project

marinaaisa:
@swift-ci test

marinaaisa (approved):
Congratulations on your hard work, Sofia, it's ready to be merged :)

PR #398, opened 2022-07-21
Preventing storing quick navigator if the feature flag is off

marinaaisa:
Bug/issue #97377452, if applicable: 

## Summary

Preventing storing quick navigator if the feature flag is off.
This fix is to prevent performance issues when rendering big frameworks in the Navigator with the quick navigator flag off.

## Dependencies

NA

## Testing

Steps:
1. Use a large framework that has navigator
2. Assert that everything works correctly and there isn't performance issues

## Checklist

- [x] Added tests

### DIFF
--- a/src/stores/QuickNavigationStore.js
+++ b/src/stores/QuickNavigationStore.js
@@ -10,7 +10,7 @@
 
 export default {
   state: {
-    enableQuickNavigation: false,
+    showQuickNavigation: false,
     flattenIndex: [],
   },
   reset() {
@@ -20,7 +20,7 @@ export default {
     this.state.flattenIndex = flattenIndex;
   },
   toggleShowQuickNavigationModal() {
-    this.state.enableQuickNavigation = !this.state.enableQuickNavigation;
+    this.state.showQuickNavigation = !this.state.showQuickNavigation;
   },
 
 };
--- a/src/components/Navigator.vue
+++ b/src/components/Navigator.vue
@@ -25,6 +25,7 @@
       :error-fetching="errorFetching"
       :breakpoint="breakpoint"
       :api-changes="apiChanges"
+      :enableQuickNavigation="enableQuickNavigation"
       @close="$emit('close')"
     />
     <NavigatorCardInner v-else class="loading-placeholder">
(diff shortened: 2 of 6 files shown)


src/stores/QuickNavigationStore.js
```diff
@@ -10,7 +10,7 @@
 
 export default {
   state: {
-    enableQuickNavigation: false,
+    showQuickNavigation: false,
```
dobromir-hristov, line 13: wont this change break the current functionality?

dobromir-hristov (commented):
I dont know if reverting is an option, but this will temporarily prevent the the issue. We should make sure to go back to the code and check whats going on

src/stores/QuickNavigationStore.js
sofiaromorales: yes, @marinaaisa this has to stay as `enableQuickNavigation ` for the modal to appear, or refactor the appearances of `enableQuickNavigation` and change it to `showQuickNavigation `. Rn with this implementation when the magnifier icon is clicked nothing will happen bc the `modal-open` class is applied when `enableQuickNavigation` is set to true in `DocumentationTopicView` component.

marinaaisa: I just fixed it at the same time :D c3ebcd9c2a1a3e0dd20780f944092241f6f2128e

sofiaromorales:
Thanks for catching this bug!

src/stores/QuickNavigationStore.js
marinaaisa: Is working properly @sofiaromorales ? I'm trying out and it does for me. Let me know if I'm missing something else

sofiaromorales: @marinaaisa yeap the code looks good to me, how big was the index file that was causing the app to break?

marinaaisa: Very big :)
We should test with big amount of data too so we make sure this is a solid feature for all kinds of documentation.
I will try to provide you with a big OSS data example soon!

marinaaisa:
@swift-ci test

dobromir-hristov (approved):

PR #402, opened 2022-08-04
Improved theme settings support

mportiz08:
Bug/issue #, if applicable: 97040358

## Summary

This PR updates the existing support for the `theme-settings.json` file to be more fully featured and capable. This is the Swift-DocC-Render implementation support for my recent forum proposal titled [
Customizing the look and feel of Swift-DocC-Render](https://forums.swift.org/t/customizing-the-look-and-feel-of-swift-docc-render/58858).

### Major Changes:
* removed default `theme-settings.json` file since it only existed for pseudo documentation purposes (full OpenAPI spec will be created for Swift-DocC as well as conceptual docs)
* fixed bugs with overwritten colors not working right when switching between Light/Dark/Auto modes
* added support for global border-radius override
* added support for overriding content and code font stacks
* added support for overriding builtin SVG icons
* added support for overriding component specific borders of the following elements
  + asides
  + badges
  + buttons
  + code listings
* added some new colors to make it easier to customize the follow elements:
  + buttons
  + documentation intro
  + tutorials overview page
* added support for customizing outer border radius of step elements
* all supported properties of the `theme-settings.json` file should be specified [here](https://github.com/mportiz08/swift-docc/blob/document-theme-settings-api/Sources/SwiftDocC/SwiftDocC.docc/Resources/ThemeSettings.spec.json) (will open PR on Swift-DocC for this soon)

More documentation PRs will be created for the swift-docc repo next to guide users on how to utilize this file.

## Dependencies

https://github.com/apple/swift-docc/pull/339 (already merged to Swift-DocC main)

## Testing

Steps:
1. From this branch, run `npm run build` to build this branch of swift-docc-render
2. Checkout [swift-docc](https://github.com/apple/swift-docc) and pull the latest main branch
3. Extract the example `theme-settings.json` file and custom `article.svg` icon from [files.zip](https://github.com/apple/swift-docc-render/files/9255665/files.zip) and copy them into the top level of the documentation catalog for SwiftDocC at `Sources/SwiftDocC/SwiftDocC.docc`
4. From the swift-docc repo, run `env DOCC_HTML_DIR=/path/to/swift-docc-render/dist bin/preview-docs SwiftDocC`
5. Open http://localhost:8000/documentation/swiftdocc and check that the theme settings are applied as expected
6. Play around with your own settings and check that there are no regressions if there are no theme settings provided

## Checklist

- [X] Added tests


mportiz08:
@swift-ci test

src/components/NavTitleContainer.vue
```diff
@@ -36,6 +36,7 @@ export default {
 }
 
 .title {
+  color: var(--color-nav-root-title, currentColor);
```
dobromir-hristov, line 39: I think if you dont specify a default, and there is no css var for it, it would just work as if the rule was never defined, right?

src/components/TutorialsOverview.vue
```diff
@@ -122,7 +122,8 @@ export default {
       padding-top: $nav-height-small;
     }
 
-    background: var(--color-tutorials-overview-background);
+    background: var(--color-tutorials-overview-fill-secondary,
```
dobromir-hristov, line 125: Is `--color-tutorials-overview-fill-secondary` like a more broad color to style a set of things at once?

src/styles/core/_vars.scss
```diff
@@ -14,7 +14,7 @@ $globalnav-height: 44px;
 $form-focus-size: 4px;
 
 // Border radius
-$border-radius: 4px !default;
+$border-radius: var(--border-radius, 4px) !default;
```
dobromir-hristov, line 17: This is cool and probably safe for the radius, but we need to careful in the future if we plan to do this to the other vars, as some are used in some SASS math 🤣

src/styles/core/typography/_font-attribute-utils.scss
```diff
@@ -94,7 +94,10 @@
 
   // generate font family
   $font-family: map-get($attributes, font-family);
-  $value: if($font-family, $font-family, -get-font-family-for-size($size));
+  $value: if($font-family,
+    $font-family,
+    var(--typography-html-font, -get-font-family-for-size($size))
```
dobromir-hristov, line 99: Do we plan to allow overriding the pre-defined `font-styles` in the future, using the `theme-settings.json` file?

src/components/SVGIcon.vue
```diff
@@ -13,13 +13,32 @@
     aria-hidden="true"
     class="svg-icon"
     xmlns="http://www.w3.org/2000/svg"
+    xmlns:xlink="http://www.w3.org/1999/xlink"
   >
-    <slot />
+    <use v-if="themeOverrideURL" :href="`${themeOverrideURL}#${themeId}`" />
```
dobromir-hristov, line 18: 👏 very good

dobromir-hristov (approved):
Awesome work Marcus, just a few questions:

1. Would we keep the old `--colors` vs `--color` naming convention? I think `--colors` used to come from the initial version of the theme-settings? Or are we just deprecating that for now, without removing support?

src/App.vue
```diff
@@ -159,15 +170,15 @@ export default {
       AppStore.setSystemColorScheme(scheme.value);
     },
     attachStylesToRoot(CSSCustomProperties) {
-      const root = document.documentElement;
+      const root = document.body;
```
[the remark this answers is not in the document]
mportiz08: I believe so, and I think this was one of the causes of the bugs with switching between modes. I agree that it would be more ideal to apply to the root instead of the body, but I think there were issues with implementing it that way in the original implementation to support dark mode, which is why it was done on the body. I think the main definitions happen in `src/styles/base/_colors.scss`.

src/components/NavTitleContainer.vue
mportiz08: I think so, but this one has a default (`currentColor`). Or am I misunderstanding what you're asking maybe?

src/components/TutorialsOverview.vue
mportiz08: Yeah, there are two main background fill colors used in this page for the default style, so I needed to add a new var to support that, and I tried to make the variable names more semantic to better make that clear (background by itself implies that there is only one main background color, which isn't totally accurate).

src/styles/core/_vars.scss
mportiz08: For sure. Please let me know if you find any potential issues with this—I was thinking the same thing when making this change.

src/styles/core/typography/_font-attribute-utils.scss
mportiz08: I think that might be a beneficial improvement to make in the future. Right now, I expect the main use case for typography is that users may want to adjust the two main font stacks used in the various pages, but we could always allow for more fine-grained typography overrides if we want to.

src/components/SVGIcon.vue
mportiz08: This still seems pretty magical to me. The documentation will need to clearly point out the importance of using the right `id` attribute in the overriding SVG file, because this magic doesn't work without that 😢

It is pretty cool though, because it allows us to both reference and embed the SVGs, allowing for the app to still control fill colors and things.

mportiz08:
> Awesome work Marcus, just a few questions:
> 
> 1. Would we keep the old `--colors` vs `--color` naming convention? I think `--colors` used to come from the initial version of the theme-settings? Or are we just deprecating that for now, without removing support?

From an API point of view, I think it should be considered unofficially deprecated. Since none of the implementation for it has been changed, things should continue to work if the plural form is used in the cases where it was previously being used, although ideally those clients using the old form could very simply adopt to the new version. Going forward, I think the singular form would be documented as the correct API. (I went with that version only because it has a 1-1 correlation with the actual CSS property names.)

src/styles/core/_vars.scss
dobromir-hristov: Quick search shows these places:

1. https://github.com/apple/swift-docc-render/pull/402/files#diff-0d3d5e64c379c67e30afd3b923825a5f077aa0a1d0bf91b9a3e5c53cd6654a07R46
2. https://github.com/mportiz08/swift-docc-render/blob/improve-theme-settings-support/src/components/Filter/FilterInput.vue#L429
3. https://github.com/mportiz08/swift-docc-render/blob/improve-theme-settings-support/src/components/Filter/FilterInput.vue#L487-L488

src/components/SVGIcon.vue
dobromir-hristov: I think if there are no drawbacks of this, it might make sense in the future to actually extract the icons back into SVGs, as its much easier to work with and preview them. Just a thought.

mportiz08: Hm, I like that thought since it would be nice to have the raw SVG assets isolated from the actual Vue components.

mportiz08:
I've opened a PR on DocC that adds an API reference for `theme-settings.json` and a conceptual article to walkthrough examples of using the file: https://github.com/apple/swift-docc/pull/359

src/styles/core/_vars.scss
mportiz08: Thanks for spotting those! The first link isn't working for me, but I'm guessing it was an instance in the badge component where there was a Sass calculation derived from the `$border-radius` var.

I think I've fixed these with minimal changes to use `calc`, which should help the CSS and Sass vars play more nicely together: 2ca4be847d1c06dcdabb4cb9bf00e1139ba46428

In the future, I would love to entirely move away from Sass vars in favor of CSS vars due to issues like this, but that would require a much larger refactor. I also initially tried the reverse of having the CSS var get defined by the SASS var, but that seemed like it would require lots of changes to a huge number of files.

Let me know if you notice any other issues related to this please, thanks!

src/components/SVGIcon.vue
mportiz08: Added docs for this to the DocC PR.

dobromir-hristov (approved):
LGTM

marinaaisa (approved):
Thanks Marcus, this is a great improvement in the way we customise our UI 👏

src/components/ContentNode/Aside.vue
```diff
@@ -74,10 +74,14 @@ $aside-kinds: deprecated, experiment, important, note, tip, warning;
...
 // the aside in question
 .doc-topic aside {
-  border-radius: $border-radius;
+  border-radius: var(--aside-border-radius, $border-radius);
+  border-style: var(--aside-border-style, solid);
+  border-width: var(--aside-border-width,
+    $aside-width-border
+    $aside-width-border
+    $aside-width-border
+    $aside-width-left-border);
```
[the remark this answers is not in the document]
mportiz08: It could be changed from the theme file, but if it isn't, then the default will still look the same as it did before, where there is a distinct variable for the left border, allowing for these to be using different values.

mportiz08:
@swift-ci test

mportiz08:
@swift-ci test

mportiz08:
@swift-ci test

PR #403, opened 2022-08-09
Add support for DocC `@tabNavigator`

dobromir-hristov:
Bug/issue #, if applicable: 97714987

## Summary

This PR adds support for DocC `@tabNavigator` directive.

image
image

The generic `<TabNavigator />` Vue component supports `start/center/end` alignment of horizontal tabs.

Tabs become vertical when more than 5. In that case, the tabs are 20% wide(% of the available horizontal space), leaving 80% for the content.

On mobile they go full-width and below the tab content

## Dependencies

DocC PR - https://github.com/apple/swift-docc/pull/379

## Testing

You can paste this JSON into any `primaryContentSections` JSON or use the attached file below. 

```json
        {
          "type": "tabNavigator",
          "tabs": [
            {
              "title": "Long tab title",
              "content": [
                {
                  "type": "text",
                  "text": "Vertical Content 1"
                }
              ]
            }
          ]
        }
```

[example.json.zip](https://github.com/apple/swift-docc-render/files/9289615/example.json.zip)

Steps:
1. Assert tabNavigator is rendered
2. Assert tabNavigator goes vertical when more than 5 items

## Checklist

- [x] Added tests


src/components/ContentNode/TabNavigator.vue
```diff
@@ -0,0 +1,90 @@
...
+ * with ability to align the tabs horizontally and vertically to the start/center/end of the line.
+ * It can also flip the navigation, so it renders content first, navigation second,
+ * in both horizontal and vertical mode.
+ */
+export default {
+  name: 'TabNavigator',
+  components: {
+    TabnavItem,
+    Tabnav,
+    ContentNode: () => import('docc-render/components/ContentNode.vue'),
```
mportiz08, line 42: I recall you asking me about different ways to solve this problem with recursive use of the `ContentNode` component, and I remember now how we have done this in the past in similar situations:

In `ContentNode`, where there are places where certain block elements can render other block elements, we have a recursive function that utilizes the slots to recursively render the nested children of a given block, and that avoids the need of using dynamic imports like this.

This situation might be _slightly_ different, because you have an array of blocks as opposed to a single block, but I _think_ it could possibly be solved the same way depending on the API of `$slots`.

src/components/ContentNode.vue
```diff
@@ -32,6 +33,7 @@ const BlockType = {
   termList: 'termList',
   unorderedList: 'unorderedList',
   dictionaryExample: 'dictionaryExample',
+  tabNavigator: 'tabNavigator',
```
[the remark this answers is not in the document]
dobromir-hristov: I just saw a draft from Ethan. Expecting to see PRs by end of the week :)

src/components/ContentNode/TabNavigator.vue
```diff
@@ -0,0 +1,90 @@
...
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <div
+    class="TabNavigator"
+    :class="[{ 'tabs--vertical': vertical }]"
+  >
+    <Tabnav v-model="current" v-bind="{ position, vertical }">
+      <TabnavItem v-for="(item, index) in items" :key="index" :value="index">
```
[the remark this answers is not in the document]
dobromir-hristov: I will change that. Should I continue using the index as an active tab tracker? It makes it easier than filtering for a title every time...

src/components/ContentNode/TabNavigator.vue
dobromir-hristov: You mean the `renderChildren` function?

How would I know which child is active and should be rendered in that case? I would have to move/expose that logic out of the TabNavigator and into to the ContentNode, which feels unnecessary.

src/components/ContentNode/TabNavigator.vue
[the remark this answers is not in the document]
mportiz08: Sure. I meant specifically for the `key` prop since Vue usually warns against using indices when possible.

src/components/ContentNode.vue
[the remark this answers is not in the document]
mportiz08: Cool, thanks. I might suggest holding off on merging these PRs until they have formal specs to make it easier to review without the DocC implementation working yet.

src/components/ContentNode/TabNavigator.vue
mportiz08: I think so, yeah. I see what you mean—to do it that way, you would have to pre-render all the children first and then decide what to hide/show.

Anyways, that was the only similar thing I could think of. I don't think there's necessarily an issue with the dynamic import, as long as it works the way it should.

mportiz08:
Which page in your example fixture has the tab nav? I'm curious what the mobile experience looks like.

dobromir-hristov:
its just one JSON, put it anywhere in an archive.

dobromir-hristov:
@SamLanier or @willimholte might have to look at the mobile exp, because I was given very little specs. We might want to double check that.

mportiz08:
> its just one JSON, put it anywhere in an archive.

Whoops. I had confused it with a random `example` folder on my filesystem :P

> @SamLanier or @willimholte might have to look at the mobile exp, because I was given very little specs. We might want to double check that.

Yeah, we may want to consider letting the tab part scroll if necessary at least for the horizontal version, since the experience isn't that great when it has to wrap.

Screenshot 2022-08-10 at 10 32 57 AM

dobromir-hristov:
Did a quick fix on the overflows. We can iron it out in the next few days with the designers.

image

dobromir-hristov:
@swift-ci test

src/components/ContentNode/TabNavigator.vue
```diff
@@ -0,0 +1,111 @@
...
+ * with ability to align the tabs horizontally and vertically to the start/center/end of the line.
+ * It can also flip the navigation, so it renders content first, navigation second,
+ * in both horizontal and vertical mode.
+ */
+export default {
+  name: 'TabNavigator',
+  components: {
+    TabnavItem,
+    Tabnav,
+    ContentNode: () => import('docc-render/components/ContentNode.vue'),
```
mportiz08, line 42: I experimented a bit with this and came up with a way of doing this without the recursive import. The main benefit is that it removes the explicit `ContentNode` dependency and allows the content to be anything. It does utilize the `renderChildren` and slots in `ContentNode` as we described.

The commit/diff can be seen here: https://github.com/dobromir-hristov/swift-docc-render/compare/dhristov/97714987/add-tabnavigator-support...mportiz08:swift-docc-render:experiment/use-slots-for-tabnavigator

 You're welcome to try it out and merge it if you like it better (would need to update tests). Also, no worries if you don't think that method is better and prefer how it is already working here. I mostly just wanted to try and see if it would work.

mportiz08 (approved):
Seems to be working well with the DocC PR.

I had one alternative approach to suggest for the dynamic import issue that we discussed previously, but you don't need to go that route if you don't prefer it.

src/components/ContentNode/TabNavigator.vue
dobromir-hristov: I looked at it and honestly I would not have thought of inverting it like that and having a slot for every tab title 🤣 That is deff an interesting approach.

Only potential downside I can think of now is that we are making Vue create all those slots but conditionally render only some.

Upside is I guess we can reuse the component in the future, without having to provide a `contentNode` compatible slot content.

I am OK with using it, your call :)

mportiz08: Your call, it's your PR 😛

src/components/ContentNode/TabNavigator.vue
```diff
@@ -0,0 +1,111 @@
+<!--
+  This source file is part of the Swift.org open source project
+
+  Copyright (c) 2021 Apple Inc. and the Swift project authors
```
marinaaisa, line 4:
```suggestion
  Copyright (c) 2022 Apple Inc. and the Swift project authors
```

src/components/Tabnav.vue
```diff
@@ -9,7 +9,10 @@
 -->
 
 <template>
-  <nav class="tabnav">
+  <nav
+    class="tabnav"
+    :class="[{ [`tabnav--${position}`]: position, 'tabnav--vertical': vertical }]"
```
marinaaisa, line 14: I believe it should work without wrapping the object in an array

marinaaisa (approved):
Tested and left minor comments :) Works great!

src/components/Tabnav.vue
dobromir-hristov: yes you are right.

dobromir-hristov:
@swift-ci test

PR #404, opened 2022-08-09
Add support for the `@Image` directive

dobromir-hristov:
Bug/issue #, if applicable: 97715120

## Summary

Extends support for the `@Image` directive, by allowing users to add captions to images. 

image

It uses the already available `InlineImage` component, and just extends the `Figure` logic to wrap, if either an `anchor` or `abstract` is present.

## Dependencies

DocC PR - 

## Testing

You can add this RenderJSON piece to the `content` array of your render json `primaryContentSections`:

```json
                {
                  "type": "image",
                  "identifier": "reference-for-image",
                  "metadata": {
                    "abstract": [
                      {
                        "type": "text",
                        "text": "Image Caption"
                      }
                    ]
                  }
                }
```

Or use the attached JSON:

[example_image.json.zip](https://github.com/apple/swift-docc-render/files/9289833/example_image.json.zip)

Steps:
1. Assert the caption is rendered below the image

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -185,19 +185,20 @@ function renderNode(createElement, references) {
 
   const renderFigure = ({
     metadata: {
-      abstract,
+      abstract = [],
       anchor,
       title,
     },
     ...node
-  }) => createElement(Figure, { props: { anchor } }, [
-    ...(title && abstract && abstract.length ? [
-      createElement(FigureCaption, { props: { title } }, (
-        renderChildren(abstract)
-      )),
-    ] : []),
-    renderChildren([node]),
-  ]);
+  }) => {
+    const figureContent = [renderChildren([node])];
+    if ((title && abstract.length) || abstract.length) {
+      // if there is a `title`, it should be above, otherwise below
+      figureContent.splice(title ? 0 : 1, 0,
+        createElement(FigureCaption, { props: { title } }, renderChildren(abstract)));
+    }
+    return createElement(Figure, { props: { anchor } }, figureContent);
+  };
 
   return function render(node) {
     switch (node.type) {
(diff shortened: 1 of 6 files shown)


src/components/ContentNode/FigureCaption.vue
```diff
@@ -10,7 +10,7 @@
 
 <template>
   <figcaption class="caption">
-    <strong>{{ title }}</strong>&nbsp;<slot />
+    <strong v-if="title">{{ title }}</strong>&nbsp;<slot />
```
mportiz08, line 13: Should this component and its `<figcaption>` wrapper ever even render if it has no text content to show?

src/components/ContentNode.vue
```diff
@@ -284,7 +285,7 @@ function renderNode(createElement, references) {
         renderChildren(node.inlineContent)
       ));
     case InlineType.image: {
-      if (node.metadata && node.metadata.anchor) {
+      if (node.metadata && (node.metadata.anchor || node.metadata.abstract)) {
```
mportiz08, line 288: Has the DocC RenderNode spec been updated to reflect this? (Just checking)

src/components/ContentNode/FigureCaption.vue
dobromir-hristov: My idea was to still render a `<figcaption>` and populate it via the `<slot>`, I just `v-if` the bolded `title`, which this feature will not utilise.

src/components/ContentNode.vue
dobromir-hristov: Ethan showed me a draft spec. I hope we get some actual DocC PRs in the next few days.

src/components/ContentNode/FigureCaption.vue
mportiz08: Oh, got it. I thought the title was all the text and missed the slot—I'm remembering that those are 2 separate things (title and content).

mportiz08 (approved):
Tested with the DocC branch and looks good (no actual spec changes needed for the image specifically it seems).

The only potential issue I noticed is that the proposal for this new directive showed examples where the caption text is centered underneath the image as opposed to left aligned as it is now. If we were to support that, we may need to update the directive itself to allow for both alignments (because we already support the left aligned ones for older content).

dobromir-hristov:
@swift-ci test

PR #405, opened 2022-08-09
Add support for `@Small` directive

dobromir-hristov:
Bug/issue #, if applicable: 97717135

## Summary

Adds support for the new `@Small` directive from DocC. 

image

## Dependencies

DocC PR - https://github.com/apple/swift-docc/pull/380

## Testing

Add the RenderJSON below to a `content` field of a renderJSON of your choice

```json
        {
          "type": "small",
          "content": [{
            "type": "text",
            "text": "Text you want to render"
          }]
        }
```

Steps:
1. Assert the `small` text is rendered in a smaller font than the rest of the content

## Checklist

- [x] Added tests


src/components/ContentNode.vue
```diff
@@ -32,6 +33,7 @@ const BlockType = {
   termList: 'termList',
   unorderedList: 'unorderedList',
   dictionaryExample: 'dictionaryExample',
+  small: 'small',
```
mportiz08, line 36: Shouldn't this be an inline element?

src/components/ContentNode/Small.vue
```diff
@@ -0,0 +1,30 @@
...
+export default {
+  name: 'Small',
+};
+</script>
+
+<style scoped lang='scss'>
+@import 'docc-render/styles/_core.scss';
+
+small {
+ @include font-styles(small-text);
```
mportiz08, line 27: Is the indentation off here or is GitHub just rendering this weird for me?

src/components/ContentNode/Small.vue
```diff
@@ -0,0 +1,30 @@
+<!--
+  This source file is part of the Swift.org open source project
+
+  Copyright (c) 2021 Apple Inc. and the Swift project authors
```
hqhhuang, line 4: 2022？

dobromir-hristov, line 4:
```suggestion
  Copyright (c) 2022 Apple Inc. and the Swift project authors
```

src/components/ContentNode.vue
dobromir-hristov: I think it should, but in the RenderJSON draft I got it was specified as a block-level element, which is weird right?

@ethan-kusters can you look into this?

mportiz08: I think we should consider changing it to be inline, to better match similar inline elements like emphasis/bold/etc.

dobromir-hristov: Right, but I think the node itself is emitted as a block one, so that might be a change DocC has to do too?

hqhhuang (approved):

src/components/ContentNode.vue
ethan-kusters: Unfortunately I think this needs to be a block element because we don't support authoring inline directives.

Bold/emphasis/etc can be inline because we have built in syntaxes for them but directive-authored syntax will always be at the block-level.

mportiz08: Got it. Sounds like it was a purposeful decision due to that limitation. Thanks @ethan-kusters!

mportiz08 (approved):
Looks good. We may want to hold off on merging until the spec is complete for this just to make sure the API expectations are synced properly.

dobromir-hristov:
This PR is now synced with the latest changes on the DocC side. Each `small` tag is wrapped in a paragraph, so to keep semantics in order.

src/styles/core/typography/_font-styles.scss
```diff
@@ -194,5 +194,8 @@ $font-styles: (
   ),
   nav-menu-collapsible: (
     large: (nav_14_14, nav)
+  ),
+  small-text: (
```
mportiz08, line 198: I think we could probably just use `body-reduced-tight` instead of adding a new duplicate style, right?

mportiz08 (approved):
Cool. I found one minor inconsistency with the spec (Ethan fixed it already), and there's one tiny comment I had regarding the new font style.

Works and looks ready to go once you have a chance to look at that.

src/styles/core/typography/_font-styles.scss
dobromir-hristov: yep, done.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #406, opened 2022-08-09
Refactor Video assets to allow sound, poster support to replayable videos, text fixes.

dobromir-hristov:
Bug/issue #, if applicable: 97715398

## Summary

1. Refactors the Video component to allow rendering videos with sound.
2. Properly pass `posterVariants` to `VideoAsset.vue` from `ReplayableVideoAsset.vue`. This was a bug left from before.
3. Make sure `ReplayableVideoAsset.vue` shows the proper text in it's "Play" butotn, if its not autoplayed yet.

## Dependencies

NA

## Testing

This PR can be best tested with 

Steps:
1. Assert Videos can be rendered with sound ON.
2. Assert that videos with Sound and Autoplay, show a "Play" button.
3. Assert that the "Play" button changes to "Replay", upon video completion. 
4. Assert that in Safari, it shows a "Play" button, even if autoplay is ON with sound.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #407, opened 2022-08-09
Add support for `@Video` directive

dobromir-hristov:
Bug/issue #, if applicable: 97715316

## Summary

Add support for DocC `@Video` directive in Documentation pages.

image

## Dependencies

This PR extends the work done the PRs below, and should be rebased after they are merged into `main`.

https://github.com/apple/swift-docc-render/pull/404
https://github.com/apple/swift-docc-render/pull/406

DocC PR - https://github.com/apple/swift-docc/pull/381

## Testing

We need proper DocC support to be able to test this out, but you can add this JSON to the `content` of a random RenderJSON file from a doccarchive, and it *should* work:

Video JSON
```json
                {
                  "type": "video",
                  "identifier": "video-reference.mp4",
                  "metadata": {
                    "abstract": [
                      {
                        "type": "text",
                        "text": "Video Caption"
                      }
                    ]
                  }
                }
```
Reference JSON inside `references`
```json
    "video-reference.mp4": {
      "poster": "reference-to-poster-image",
      "type": "video",
      "identifier": "video-reference.mp4",
      "alt": null,
      "variants": [
        {
          "traits": [
            "1x",
            "light"
          ],
          "url": "/path/to/video.mp4"
        },
        {
          "traits": [
            "1x",
            "dark"
          ],
          "url": "/path/to/video~dark.mp4"
        }
      ]
    }
```

Steps:
1. Assert a replayable video is rendered with a "Play" button beneath it.
2. Assert Video can be rendered with a caption below it, but also without.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -11,6 +11,7 @@
 <script>
 import Aside from './ContentNode/Aside.vue';
 import CodeListing from './ContentNode/CodeListing.vue';
+import LinkableHeading from './ContentNode/LinkableHeading.vue';
 import CodeVoice from './ContentNode/CodeVoice.vue';
 import DictionaryExample from './ContentNode/DictionaryExample.vue';
 import EndpointExample from './ContentNode/EndpointExample.vue';
@@ -234,14 +235,13 @@ function renderNode(createElement, references) {
       };
       return createElement(EndpointExample, { props }, renderChildren(node.summary || []));
     }
-    case BlockType.heading:
-      return createElement(`h${node.level}`, {
-        attrs: {
-          id: node.anchor,
-        },
-      }, (
-        node.text
-      ));
+    case BlockType.heading: {
+      const props = {
+        anchor: node.anchor,
+        level: node.level,
+      };
+      return createElement(LinkableHeading, { props }, node.text);
+    }
     case BlockType.orderedList:
       return createElement('ol', {
         attrs: {
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -197,7 +197,9 @@ function renderNode(createElement, references) {
     if ((title && abstract.length) || abstract.length) {
       // if there is a `title`, it should be above, otherwise below
       figureContent.splice(title ? 0 : 1, 0,
-        createElement(FigureCaption, { props: { title } }, renderChildren(abstract)));
+        createElement(FigureCaption, {
+          props: { title, centered: !title },
+        }, renderChildren(abstract)));
     }
     return createElement(Figure, { props: { anchor } }, figureContent);
   };
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -20,6 +20,7 @@ import InlineImage from './ContentNode/InlineImage.vue';
 import Reference from './ContentNode/Reference.vue';
 import Table from './ContentNode/Table.vue';
 import StrikeThrough from './ContentNode/StrikeThrough.vue';
+import Small from './ContentNode/Small.vue';
 import BlockVideo from './ContentNode/BlockVideo.vue';
 
 const BlockType = {
@@ -33,6 +34,7 @@ const BlockType = {
   termList: 'termList',
   unorderedList: 'unorderedList',
   dictionaryExample: 'dictionaryExample',
+  small: 'small',
   video: 'video',
 };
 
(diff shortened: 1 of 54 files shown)


src/components/ContentNode.vue
```diff
@@ -38,6 +39,7 @@ const InlineType = {
   codeVoice: 'codeVoice',
   emphasis: 'emphasis',
   image: 'image',
+  video: 'video',
```
mportiz08, line 42: Is a video really an "inline" element? I imagine we would be rendering all videos as distinct blocks of content, but I could be mistaken.

src/components/ContentNode.vue
```diff
@@ -185,19 +187,20 @@ function renderNode(createElement, references) {
...
     ...node
-  }) => createElement(Figure, { props: { anchor } }, [
-    ...(title && abstract && abstract.length ? [
-      createElement(FigureCaption, { props: { title } }, (
-        renderChildren(abstract)
-      )),
-    ] : []),
-    renderChildren([node]),
-  ]);
+  }) => {
```
mportiz08, line 195: These changes are from another PR, right? I think in the future it would be easier to open these PRs with a dependency after the dependent PR has been merged, so that they're easier to review in isolation.

mportiz08 (commented):
I think we should mark this PR as a draft until the two dependent PRs have been merged, and this has been rebased.

dobromir-hristov:
> I think we should mark this PR as a draft until the two dependent PRs have been merged, and this has been rebased.

Yes, I just wanted to get things rolling.

src/components/ContentNode.vue
dobromir-hristov: You do have a point. I just followed the InlineImage.vue component, but that probably is not correct here.

dobromir-hristov: @ethan-kusters what do you think? Should the video be `inline` like the image, even though our images are not really inline either... :D I wont hurry to change anything for now.

ethan-kusters: Hm... I think we should leave it as inline even though I agree we're kind of misusing that term. We want video supported anywhere image is and we currently consider those inline.

We could consider refactoring all of those in the future but I think it's a separate task.

Screenshot 2022-09-09 at 12 22 54 PM

ethan-kusters: I chatted with @franklinsch about this and came to the opposite conclusion. 😁 

Commonmark markdown actually supports inline images which is why we use an inline element for it even though, today, DocC always wrap images in paragraphs which effectively makes them block elements. In the future we might want to actually properly support inline images so it makes sense to leave this as-is.

However, since DocC videos can only be authored with directives and we don't support inline directives (and we don't see a use-case for inline videos) it doesn't make sense for videos to be inline elements. I'll put up a PR momentarily but I'm updating the spec to make Video a block element based on this- let me know if ya'll disagree with this general line of logic though. Happy to discuss further.

ethan-kusters: Spec update for this is here: https://github.com/apple/swift-docc/pull/381/files#diff-fc7c8431152d3d494899ec4b551f1a5d1249ed44c853e05da1cc26374d9d4c54.

mportiz08: Sounds reasonable to me. The block vs inline image thing will be a pain in the future in terms of backwards compatibility but that's entirely separate from this.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

mportiz08 (approved):
Looks good and is working with the DocC branch.

My only minor concern is that the "Play" icon is the same icon as the "Replay" icon, which is a little bit odd. I would probably actually prefer that we just use the normal video controls and remove the play/replay link entirely, but you may want to discuss with the design team.

Also, make sure to rebase this against the main branch once the 3 dependent PRs have been merged (the replay one, the image one, and the figure caption one) before merging this one.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
> My only minor concern is that the "Play" icon is the same icon as the "Replay" icon

You have a point Marcus, I changed that because it's a good point anyway.

image

I will check with Design if they want me to just show full native controls all the time, instead of just on mobile devices. I think we discussed that we may want just the minimal UI of the tutorials.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #190, opened 2022-04-19
Add anchors to ContentTableSection

dobromir-hristov:
Bug/issue #, if applicable: 79432376

## Summary

This PR adds anchors to topic link sections, so users can link to them from content.

## Dependencies

Docc needs to implement this  and update the RenderJSON spec.

```diff
{
            "TasksGroupSection": {
                "required": [
                    "title",
                    "identifiers"
                ],
                "type": "object",
                "properties": {
                    "title": {
                        "type": "string"
                    },
                    "abstract": {
                        "type": "array",
                        "items": {
                            "$ref": "#/components/schemas/RenderInlineContent"
                        }
                    },
                    "discussion": {
                        "$ref": "#/components/schemas/ContentRenderSection"
                    },
                    "identifiers": {
                        "type": "array",
                        "items": {
                            "type": "string"
                        }
                    },
                    "generated": {
                        "type": "boolean"
                    },
+                 "anchor":{
+                     "type": "string"
+                  }
                }
            },
}
```

## Testing

User the provided archive - [TinyMixedLanguageLibrary.doccarchive.zip](https://github.com/apple/swift-docc-render/files/8514630/TinyMixedLanguageLibrary.doccarchive.zip)

Steps:
1. Go to http://localhost:8080/documentation/tinymixedlanguagelibrary/mixedlangclass
2. Click on `Initializers` link
3. Assert you are scrolled to that part of the page.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #408, opened 2022-08-09
Add anchor for section titles for quick sharing of sections

marinaaisa:
Bug/issue #96943502, if applicable: 

## Summary

Add anchor for section titles for quick sharing of sections.
This anchor shows when user hovers the title section.

Issue: https://github.com/apple/swift-docc-render/issues/273

## Dependencies

NA

## Testing

Steps:
1. Open any docc archive
2. Assert that all h2 and h3 titles in documentation pages show an anchor when hovering the title

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -11,7 +11,7 @@
 <script>
 import Aside from './ContentNode/Aside.vue';
 import CodeListing from './ContentNode/CodeListing.vue';
-import SectionTitle from './ContentNode/SectionTitle.vue';
+import LinkableHeading from './ContentNode/LinkableHeading.vue';
 import CodeVoice from './ContentNode/CodeVoice.vue';
 import DictionaryExample from './ContentNode/DictionaryExample.vue';
 import EndpointExample from './ContentNode/EndpointExample.vue';
@@ -25,7 +25,6 @@ import StrikeThrough from './ContentNode/StrikeThrough.vue';
 const BlockType = {
   aside: 'aside',
   codeListing: 'codeListing',
-  sectionTitle: 'sectionTitle',
   endpointExample: 'endpointExample',
   heading: 'heading',
   orderedList: 'orderedList',
@@ -232,9 +231,9 @@ function renderNode(createElement, references) {
     case BlockType.heading: {
       const props = {
         anchor: node.anchor,
-        tag: `h${node.level}`,
+        level: Number(node.level),
       };
-      return createElement(SectionTitle, { props }, node.text);
+      return createElement(LinkableHeading, { props }, node.text);
     }
     case BlockType.orderedList:
       return createElement('ol', {
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -11,6 +11,7 @@
 <script>
 import Aside from './ContentNode/Aside.vue';
 import CodeListing from './ContentNode/CodeListing.vue';
+import SectionTitle from './ContentNode/SectionTitle.vue';
 import CodeVoice from './ContentNode/CodeVoice.vue';
 import DictionaryExample from './ContentNode/DictionaryExample.vue';
 import EndpointExample from './ContentNode/EndpointExample.vue';
@@ -24,6 +25,7 @@ import StrikeThrough from './ContentNode/StrikeThrough.vue';
 const BlockType = {
   aside: 'aside',
   codeListing: 'codeListing',
+  sectionTitle: 'sectionTitle',
   endpointExample: 'endpointExample',
   heading: 'heading',
   orderedList: 'orderedList',
@@ -227,24 +229,13 @@ function renderNode(createElement, references) {
       };
       return createElement(EndpointExample, { props }, renderChildren(node.summary || []));
     }
-    case BlockType.heading:
-      return createElement(`h${node.level}`, {
-        attrs: {
-          id: node.anchor,
-        },
-      }, [
-        createElement('a', {
-          attrs: {
-            id: node.anchor,
-            class: 'header-anchor',
-            href: `#${node.anchor}`,
-            'aria-hidden': 'true',
-          },
-        }, (
-          '#'
-        )),
-        node.text,
-      ]);
+    case BlockType.heading: {
+      const props = {
+        anchor: node.anchor,
+        tag: `h${node.level}`,
+      };
+      return createElement(SectionTitle, { props }, node.text);
+    }
     case BlockType.orderedList:
       return createElement('ol', {
         attrs: {
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -231,7 +231,7 @@ function renderNode(createElement, references) {
     case BlockType.heading: {
       const props = {
         anchor: node.anchor,
-        level: Number(node.level),
+        level: node.level,
       };
       return createElement(LinkableHeading, { props }, node.text);
     }
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -232,9 +232,19 @@ function renderNode(createElement, references) {
         attrs: {
           id: node.anchor,
         },
-      }, (
-        node.text
-      ));
+      }, [
+        createElement('a', {
+          attrs: {
+            id: node.anchor,
+            class: 'header-anchor',
+            href: `#${node.anchor}`,
+            'aria-hidden': 'true',
+          },
+        }, (
+          '#'
+        )),
+        node.text,
+      ]);
     case BlockType.orderedList:
       return createElement('ol', {
         attrs: {
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -10,8 +10,8 @@
 
 <template>
   <component
-   :id="!isTargetIDE ? id : null"
-   :is="tag"
+   :id="id"
+   :is="`h${level}`"
    class="section-title"
   >
     <a
@@ -28,8 +28,10 @@
 <script>
 import scrollToElement from 'docc-render/mixins/scrollToElement';
 
+const HeadingLevels = [1, 2, 3, 4, 5, 6];
+
 export default {
-  name: 'SectionTitle',
+  name: 'LinkableHeading',
   mixins: [scrollToElement],
   data() {
     return {
@@ -41,9 +43,10 @@ export default {
       type: String,
       required: false,
     },
-    tag: {
-      type: String,
-      default: () => 'h2',
+    level: {
+      type: Number,
+      default: () => 2,
+      validator: v => v in HeadingLevels,
     },
   },
   inject: {
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -20,6 +20,7 @@
       @click="handleFocusAndScroll(anchor)"
     >
       <slot />
+      <LinkIcon class="icon" aria-hidden="true"/>
     </router-link>
     <template v-else>
       <slot />
@@ -29,10 +30,14 @@
 
 <script>
 import scrollToElement from 'docc-render/mixins/scrollToElement';
+import LinkIcon from 'theme/components/Icons/LinkIcon.vue';
 
 export default {
   name: 'LinkableHeading',
   mixins: [scrollToElement],
+  components: {
+    LinkIcon,
+  },
   props: {
     anchor: {
       type: String,
@@ -58,22 +63,28 @@ export default {
 .header-anchor {
   color: var(--colors-text, var(--color-text));
   text-decoration: none;
-  position: relative;
 
-  &:after {
-    content: "";
-    display: block;
-    position: absolute;
-    width: 0px;
-    bottom: -3px;
-    height: 1px;
-    background-color: var(--colors-text, var(--color-text));
-    // transition: width 0.2s cubic-bezier(0.82, 0.085, 0.395, 0.895) 0s;
-    // transition-delay: .5s;
+  // &:after {
+  //   content: "";
+  //   display: block;
+  //   position: absolute;
+  //   width: 0px;
+  //   bottom: -3px;
+  //   height: 1px;
+  //   background-color: var(--colors-text, var(--color-text));
+  //   // transition: width 0.2s cubic-bezier(0.82, 0.085, 0.395, 0.895) 0s;
+  //   // transition-delay: .5s;
+  // }
+
+  .icon {
+    display: none;
+    height: .6em;
+    margin-left: .3em;
+    fill: currentColor;
   }
 
-  &:hover::after {
-    width: 100%;
+  &:hover .icon {
+    display: inline;
   }
 }
 </style>
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -28,8 +28,6 @@
 <script>
 import scrollToElement from 'docc-render/mixins/scrollToElement';
 
-const HeadingLevels = [1, 2, 3, 4, 5, 6];
-
 export default {
   name: 'LinkableHeading',
   mixins: [scrollToElement],
@@ -46,7 +44,7 @@ export default {
     level: {
       type: Number,
       default: () => 2,
-      validator: v => v in HeadingLevels,
+      validator: v => v >= 1 && v <= 6,
     },
   },
   inject: {
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -17,6 +17,7 @@
       v-if="anchor && !isTargetIDE"
       :to="{ hash: `#${anchor}` }"
       class="header-anchor"
+      aria-label="Scroll to section"
       @click="handleFocusAndScroll(anchor)"
     >
       <slot />
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -10,7 +10,7 @@
 
 <template>
   <component
-   :id="id"
+   :id="anchor"
    :is="`h${level}`"
    class="section-title"
   >
@@ -31,11 +31,6 @@ import scrollToElement from 'docc-render/mixins/scrollToElement';
 export default {
   name: 'LinkableHeading',
   mixins: [scrollToElement],
-  data() {
-    return {
-      id: null,
-    };
-  },
   props: {
     anchor: {
       type: String,
@@ -51,15 +46,6 @@ export default {
     isTargetIDE: {
       default: () => false,
     },
-  },
-  mounted() {
-    if (!this.anchor) return;
-    const element = document.getElementById(`${this.anchor}`);
-    // if there is no element in the document that has an id,
-    // add it to this component
-    if (!element) {
-      this.id = this.anchor;
-    }
   },
 };
 </script>
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -63,17 +63,17 @@ export default {
 $icon-margin: 7px;
 
 .header-anchor {
-  color: var(--colors-text, var(--color-text));
+  color: inherit;
   text-decoration: none;
   position: relative;
-  padding-right: $icon-size + $icon-margin;
+  padding-right: $icon-size-default + $icon-margin;
   display: inline-block;
 
   .icon {
     position: absolute;
     bottom: .2em;
     display: none;
-    height: $icon-size;
+    height: $icon-size-default;
     margin-left: $icon-margin;
   }
 
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -60,21 +60,21 @@ export default {
 
 <style scoped lang="scss">
 @import 'docc-render/styles/_core.scss';
+$icon-margin: 7px;
 
 .header-anchor {
   color: var(--colors-text, var(--color-text));
   text-decoration: none;
   position: relative;
-  padding-right: 1em;
+  padding-right: $icon-size + $icon-margin;
   display: inline-block;
 
   .icon {
     position: absolute;
     bottom: .2em;
     display: none;
-    height: .6em;
-    margin-left: .3em;
-    fill: currentColor;
+    height: $icon-size;
+    margin-left: $icon-margin;
   }
 
   &:hover .icon {
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -63,20 +63,11 @@ export default {
 .header-anchor {
   color: var(--colors-text, var(--color-text));
   text-decoration: none;
-
-  // &:after {
-  //   content: "";
-  //   display: block;
-  //   position: absolute;
-  //   width: 0px;
-  //   bottom: -3px;
-  //   height: 1px;
-  //   background-color: var(--colors-text, var(--color-text));
-  //   // transition: width 0.2s cubic-bezier(0.82, 0.085, 0.395, 0.895) 0s;
-  //   // transition-delay: .5s;
-  // }
+  position: relative;
 
   .icon {
+    position: absolute;
+    bottom: .2em;
     display: none;
     height: .6em;
     margin-left: .3em;
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -12,16 +12,18 @@
   <component
    :id="anchor"
    :is="`h${level}`"
-   class="section-title"
   >
     <router-link
       v-if="anchor && !isTargetIDE"
       :to="{ hash: `#${anchor}` }"
       class="header-anchor"
-      aria-label="hidden"
       @click="handleFocusAndScroll(anchor)"
-    >#</router-link>
-    <slot />
+    >
+      <slot />
+    </router-link>
+    <template v-else>
+      <slot />
+    </template>
   </component>
 </template>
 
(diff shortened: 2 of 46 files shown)


marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

src/components/ContentNode/SectionTitle.vue
```diff
@@ -0,0 +1,75 @@
...
+  Copyright (c) 2022 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <component
+   :id="anchor ? anchor : null"
```
dobromir-hristov, line 13:
```suggestion
   :id="anchor || null"
```

src/components/ContentNode/SectionTitle.vue
```diff
@@ -0,0 +1,75 @@
...
+   :is="tag"
+   class="section-title"
+  >
+    <a
+      v-if="anchor || href"
+      :href="`#${anchor || href}`"
```
dobromir-hristov, line 19: I would suggest you look at how we have done this in the past, like in the TutorialsNavigationLink component.

We use a `scrollToElement` function that actually adjusts the scrolling by a few pixels, so the navigation does not overlap the title. I saw you added `scroll-margin-top` but I am getting cutoffs on the Topic section

We also used to use `<router-link>` components instead of `a`, not sure if there is a major diff.

image

dobromir-hristov: So if we pass `anchor`, it adds an ID, and if we pass just `href` it does not, and it uses both to create a `href`?

I had to go back midway review when I finally got whats going on, felt confusing. I will leave it the the rest of the team to give more feedback.

src/components/DocumentationTopic/ContentTableSection.vue
```diff
@@ -24,14 +24,20 @@
...
   name: 'ContentTableSection',
+  components: { SectionTitle },
   props: {
     title: {
       type: String,
       required: true,
     },
   },
+  computed: {
+    anchor: ({ title }) => title.replaceAll(' ', '-'),
```
dobromir-hristov, line 39: we have an `anchorize` function, use that instead.

dobromir-hristov (commented):
This looks great, very nice! 🚀 

One thing I noticed is if I tab through, it does not show the anchors, deff something we should fix.

Need to check if we should disable this for IDE environments.

Also we may check with the DocC team if they plan to provide anchors for those Topic subsections, instead of us generating them - #190

Other than that, looks great :)

src/components/ContentNode/SectionTitle.vue
marinaaisa: I addressed the feedback and I fixed the confusion with the anchor and href prop. Now we only have one. Check it out again for another review, thanks!

marinaaisa:
Thanks @dobromir-hristov !

> One thing I noticed is if I tab through, it does not show the anchors, deff something we should fix.

Fixed! Good catch!

> Need to check if we should disable this for IDE environments.

Why would we? it should scroll in the same way, right?
 
> Also we may check with the DocC team if they plan to provide anchors for those Topic subsections, instead of us generating them - #190

Yeah, that would be much better than the generation I'm doing in the frontend!

Give it another review. I still need to add final tests for the `handleFocusAndScroll` function I added in the scrollToElement mixin.

dobromir-hristov:
I don't see a point in having a clickable link in the ide, since there is no URL bar

dobromir-hristov (commented):
I think it looks good now. How does VO handle the hidden anchor item? Do we need anything special for it?

marinaaisa:
> I think it looks good now. How does VO handle the hidden anchor item? Do we need anything special for it?

We are hiding it from VO because I think it's not providing any value to VO users. https://vuejs.org/guide/introduction.html#what-is-vue does it in the same way

src/components/ContentNode/SectionTitle.vue
```diff
@@ -10,51 +10,57 @@
...
-      required: false,
-    },
     tag: {
       type: String,
       default: () => 'h2',
     },
   },
+  mounted() {
+    if (!this.anchor) return;
+    const element = document.getElementById(`${this.anchor}`);
```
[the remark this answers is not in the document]
marinaaisa: Unpredictable? We use dynamic ids in many more components across the app... 🤔 
But if you have any better approach, happy to look into it :)

marinaaisa:
> I don't see a point in having a clickable link in the ide, since there is no URL bar

True, I don't render it anymore on IDEs https://github.com/apple/swift-docc-render/pull/408/commits/f6bd656caaab0e1c853e5f416cdceca93f6d273c

marinaaisa:
@swift-ci test

src/components/ContentNode/SectionTitle.vue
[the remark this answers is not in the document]
dobromir-hristov: Well you may be right. I would have probably just made this logic user controllable, via some boolean whether to apply an `id` to the element or not.

dobromir-hristov (commented):
Looks good. @mportiz08 can you look into the anchor logic if you are fine with it and we can merge this one I think

src/components/ContentNode.vue
```diff
@@ -24,6 +25,7 @@ import StrikeThrough from './ContentNode/StrikeThrough.vue';
 const BlockType = {
   aside: 'aside',
   codeListing: 'codeListing',
+  sectionTitle: 'sectionTitle',
```
mportiz08, line 28: I don't think this line should needed since these enum values refer to content node block types from the Render Node spec (and this is not one).

src/components/ContentNode.vue
```diff
@@ -227,14 +229,13 @@ function renderNode(createElement, references) {
...
-        attrs: {
-          id: node.anchor,
-        },
-      }, (
-        node.text
-      ));
+    case BlockType.heading: {
+      const props = {
+        anchor: node.anchor,
+        tag: `h${node.level}`,
```
mportiz08, line 235: I think maybe we should pass the `level` value as a prop and then map those to the right header tag in the new component—otherwise we allow for any arbitrary string to be used as the HTML tag for this component when we probably only want to allow h1-h6.

src/components/ContentNode/SectionTitle.vue
```diff
@@ -0,0 +1,86 @@
...
+    >#</a>
+    <slot />
+  </component>
+</template>
+
+<script>
+import scrollToElement from 'docc-render/mixins/scrollToElement';
+
+export default {
+  name: 'SectionTitle',
```
mportiz08, line 32: I think we might want to rename this to better indicate that this component is used for headings. Maybe just `Heading` or `LinkableHeading`? That would also more closely align with the node kind from the Render Node spec (not necessary but maybe nice).

src/components/ContentNode/SectionTitle.vue
```diff
@@ -0,0 +1,86 @@
...
+    },
+  },
+  inject: {
+    isTargetIDE: {
+      default: () => false,
+    },
+  },
+  mounted() {
+    if (!this.anchor) return;
+    const element = document.getElementById(`${this.anchor}`);
```
mportiz08, line 56:
```suggestion
    const element = document.getElementById(this.anchor);
```

As a more general comment though, I don't totally understand the need for why this hook is needed. Why are there separate `id` and `anchor` values and how are they used differently?

src/components/ContentNode/SectionTitle.vue
```diff
@@ -0,0 +1,86 @@
...
+  Copyright (c) 2022 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <component
+   :id="!isTargetIDE ? id : null"
```
mportiz08, line 13: Can this just be `anchor` or am I missing something?

Also, I don't think we need to strip out the `id` attribute even if the actual link to that anchor isn't presented in certain situations (like in IDE views).

src/components/ContentNode/SectionTitle.vue
```diff
@@ -0,0 +1,86 @@
...
+    };
+  },
+  props: {
+    anchor: {
+      type: String,
+      required: false,
+    },
+    tag: {
+      type: String,
+      default: () => 'h2',
```
mportiz08, line 46: Whether this stays as a `tag` or `level` prop, we should probably add a validator to ensure that you can only ever use a h1-h6 (or only a 1-6 level).

src/components/ContentNode/SectionTitle.vue
```diff
@@ -0,0 +1,86 @@
...
+.section-title:hover {
+  .header-anchor {
+    opacity: 1;
+  }
+}
+
+.header-anchor {
+  margin-left: -0.73em;
+  padding-right: 0.23em;
+  transition: opacity .25s;
```
mportiz08, line 78: You may want to have someone from the design team review this new feature and test the experience in different browsers. The scrolling animation specifically (not this one) feels a little different in Safari but that may be a limitation of the JS scroll function itself.

src/components/DocumentationTopic/ContentTable.vue
```diff
@@ -15,18 +15,19 @@
     :title="title"
   >
     <div class="container">
-      <h2 class="title">{{title}}</h2>
+      <SectionTitle class="title" :anchor="anchor">{{ title }}</SectionTitle>
```
mportiz08, line 18: We might need to do some additional investigation on this, but I suspect that we are now adding multiple elements with the same `id` value to the page since the `OnThisPageSection` is also using this `anchor` as an `id` for another HTML element, and there should only be one on the page since these are meant to be unique on a given page.

It might be as simple as removing it from the other component since it was previously used for the old "On This Page" affordance, which was removed with the introduction of the sidebar.

src/components/DocumentationTopic/ContentTableSection.vue
```diff
@@ -24,14 +24,21 @@
...
   name: 'ContentTableSection',
+  components: { SectionTitle },
   props: {
     title: {
       type: String,
       required: true,
     },
   },
+  computed: {
+    anchor: ({ title }) => anchorize(title),
```
mportiz08, line 40: I know that DocC is not currently providing `anchor` values for these headings, but I think that work is being tracked since we ideally want it to be in control of the id as opposed to having it being generated it here—I think we could utilize `anchorize` as a fallback in the meantime, but we may run into future issues where old links might not work as one system takes over.

src/mixins/scrollToElement.js
```diff
@@ -27,5 +27,19 @@ export default {
       window.scrollBy(-offset.x, -offset.y);
       return element;
     },
+    /**
+     * Always scroll to the element and focus it.
+     * This ensures that the next tab target is inside
+     * and that it is in view, even if the hash is in the url
+     * @returns {Promise<void>}
```
mportiz08, line 34:
```suggestion
     * @returns {Promise<Element>}
```

mportiz08 (commented):
Very cool to see this in action, and I love that it just works as expected in places where no anchor data is provided.

I have a minor concern about the anchor generation, but I think the majority of my other comments are pretty small things that will be easy to fix.

mportiz08:
Just noting that we may want to close #190 when this gets merged.

src/components/ContentNode/SectionTitle.vue
dobromir-hristov: There are cases where the ID is applied on a section that wraps the heading (like in TopicsSection), but in other cases, the heading has to hold the ID.

src/components/DocumentationTopic/ContentTable.vue
dobromir-hristov: #416 :) Not quite :D

dobromir-hristov:
> Just noting that we may want to close https://github.com/apple/swift-docc-render/pull/190 when this gets merged.

Even though this PR adds most of what #190 does, that one was assuming the anchors come from DocC. I think if Marina does that too, we can safely drop it :)

src/components/ContentNode/SectionTitle.vue
mportiz08: I'm not totally following how that requires the mounted hook here (probably just missing something).

src/components/DocumentationTopic/ContentTable.vue
mportiz08: I know that component is coming back in a forthcoming PR, but we will need to ensure that we are only assigning a given unique ID to a single HTML element and not adding it in multiple places, regardless of whether or not there is an "On This Page" component being utilized.

src/components/ContentNode/SectionTitle.vue
dobromir-hristov: I think she wanted it to be more magical. Thats one of the points I raised that I wanted your input on :D 

The component tries to find an already existing element with that `id` in the DOM, and if it doesnt, it adds that `id` to itself.

src/components/ContentNode/SectionTitle.vue
dobromir-hristov: We should strip out the UI element on hover/focus, but the ID here is fine to stay.

src/components/DocumentationTopic/ContentTable.vue
dobromir-hristov: Right, I think you can check my explanation on how this component works and that might make things more clear. 
As long as we dont touch the OnThisPage registration logic, I am good 🤣

src/components/ContentNode/SectionTitle.vue
mportiz08: I guess I just don't understand the magical part, to be more clear. Why would this logic have to happen after the initial render?

dobromir-hristov: So all possible items that may have that ID are rendered? As I said I would probably control this through a prop instead.

marinaaisa: @mportiz08 I had this logic to check if the document has already an `id` on it where we want to link to. This happens in components that have the id in the beginning of the section itself instead of having it in the title.
Also in this way, we avoid duplication of `id`s in the HTML, since ID attributes must be unique in the document. 

Controlling through a prop, as @dobromir-hristov suggested, would allow us more flexibility but would open the component for possible HTML's errors because we wouldn't prevent the document from duplication of HTML `id`s.

src/components/DocumentationTopic/ContentTable.vue
marinaaisa: Current logic should be enough to prevent duplication: https://github.com/apple/swift-docc-render/pull/408#discussion_r958671127

src/mixins/scrollToElement.js
marinaaisa: fixed!

src/components/DocumentationTopic/ContentTableSection.vue
marinaaisa: I added most of the logic from https://github.com/apple/swift-docc-render/pull/190 into this PR to address this

src/components/ContentNode/SectionTitle.vue
marinaaisa: Added!

src/components/ContentNode/SectionTitle.vue
marinaaisa: I left the id without changing it

src/components/ContentNode/SectionTitle.vue
marinaaisa: I liked it! I renamed it to `LinkableHeading` :)

src/components/ContentNode.vue
marinaaisa: done!

marinaaisa:
> Even though this PR adds most of what https://github.com/apple/swift-docc-render/pull/190 does, that one was assuming the anchors come from DocC. I think if Marina does that too, we can safely drop it :)

I moved most of the changes from https://github.com/apple/swift-docc-render/pull/190 PR into this one, so we can merge it in here.

marinaaisa:
@swift-ci test

src/components/ContentNode/SectionTitle.vue
mportiz08: I understand why we might choose to sometimes assign the ID attribute to a non-heading element instead of a heading, but I don't understand why we couldn't prevent the ID from being assigned in multiple places in the components themselves—it seems like that is controlled by our app and not outside forces. If that's the case, then we should be able to programmatically decide where the anchor uniquely goes before any `mounted` hook gets called after the initial render.

I think we both agree on the goal of only having unique IDs but maybe there's a little confusion about how to ensure it within the components themselves. Let me know if you want to talk about this offline—I may be missing a subtle thing here that might make this approach necessary.

src/components/ContentNode/LinkableHeading.vue
```diff
@@ -41,9 +43,10 @@ export default {
       type: String,
       required: false,
     },
-    tag: {
-      type: String,
-      default: () => 'h2',
+    level: {
+      type: Number,
+      default: () => 2,
+      validator: v => v in HeadingLevels,
```
mportiz08, line 49: I don't think the `in` operator works like you're expecting it to here.

Example:
```javascript
> 0 in [1, 2, 3, 4, 5, 6]
true
> 6 in [1, 2, 3, 4, 5, 6]
false
```

It might be better to just check if the number is in bounds. Something like:
```javascript
v => v >= 1 && v <= 6
```

src/components/ContentNode.vue
```diff
@@ -232,9 +231,9 @@ function renderNode(createElement, references) {
     case BlockType.heading: {
       const props = {
         anchor: node.anchor,
-        tag: `h${node.level}`,
+        level: Number(node.level),
```
mportiz08, line 234: I believe that `node.level` values are guaranteed to be integers in the RenderNode spec, so you can just pass `node.level` here without the `Number()` call.

src/components/DocumentationTopic/ContentTableSection.vue
mportiz08: Good idea, thanks. There could be future issues when the switch happens where the anchors currently being generated in the app may conflict with futures ones provided by DocC, but I think that on-page link URLs are generally pretty unstable with titles changing anyways, so it probably makes sense to rely on DocC and only generate them when they aren't for now.

src/components/ContentNode.vue
marinaaisa: fixed!

src/components/ContentNode/LinkableHeading.vue
marinaaisa: Oh, thank you! it seemed I was using it wrong, yeah :D fixed it!

src/components/ContentNode/SectionTitle.vue
marinaaisa: @mportiz08 I set a meeting to continue discussing this. Interesting debate! Thanks :)

marinaaisa:
@swift-ci test

mportiz08 (approved):

hqhhuang:
Hey Marina!
This is brilliant! I still have questions about the `id`, but I will reach out offline for that.
I tested this and noticed 3 small things:
1. The URL seems to be case sensitive. Is this intentional?
2. This issue appears in Chrome only:

https://user-images.githubusercontent.com/87735557/188030876-b5ce5a28-39f1-4cf0-b5dd-3ad611ae10f5.mov

3. Do we want to dynamically update the URL with the anchors based on where the user is on the page as they scroll? I think the Vue Doc site does this.

marinaaisa:
Hey @hqhhuang ! thanks a lot for your comments!

> 1. The URL seems to be case sensitive. Is this intentional?

URL are always case sensitive. You will see that some anchors are in lowercase and others aren't. This is because some come directly from docc (some have uppercase initials) and other we add them depending on the title (these ones are in lowercase). The anchor's names don't matter if they are in lowercase or not, as long as they point to an existing id with the same name. 

> 2. This issue appears in Chrome only:
>  Screen.Recording.2022-09-01.at.4.24.57.PM.mov

This is because of the way we scroll to the section. We first scroll into the view where the title is, and then we scrollBy to overcome the navigation height. You can see it here: https://github.com/apple/swift-docc-render/blob/f5c1b7fd271518503e85f7a5d2dd5f993ded17be/src/mixins/scrollToElement.js#L26-L27
I'm not sure yet how to prevent that jumping at the beginning, in theory, it shouldn't be noticeable... we could look into improving it in the future.

> 3. Do we want to dynamically update the URL with the anchors based on where the user is on the page as they scroll? I think the Vue Doc site does this.

Very good catch! Maybe this could a separate feature since it uses the ids but it doesn't depend on the new LinkableHeadinds, could you fill a radar or Github issue for it? Thanks!

mportiz08:
I know you're still iterating on the visual design of the link, but I wanted to say that we might want to consider some sort of absolute positioning of it to avoid situations like the following where the inline content needs to reflow sometimes:

Screen Recording 2022-09-02 at 1 33 46 PM

hqhhuang:
**URL capitalization:**
> This is because some come directly from docc (some have uppercase initials) and other we add them depending on the title (these ones are in lowercase). The anchor's names don't matter if they are in lowercase or not, as long as they point to an existing id with the same name.

What are some existing examples of ones coming directly from docc? I think if we opted for all lowercase for Docc-render generated anchors, docc should also provide all lower case anchors. Is there a particular reason some of the docc ones have upper case initials?

**Scroll jump:**
> This is because of the way we scroll to the section. We first scroll into the view where the title is, and then we scrollBy to overcome the navigation height. You can see it here:
> 
> https://github.com/apple/swift-docc-render/blob/f5c1b7fd271518503e85f7a5d2dd5f993ded17be/src/mixins/scrollToElement.js#L26-L27

Is there a difference between how Chrome and Safari handles the scroll? If this is the issue, I would sort of expect it to happen in both places. 

**Updating URL based on anchors:**
> Could you fill a radar or Github issue for it? Thanks!

rdar://99584817

src/components/ContentNode/LinkableHeading.vue
```diff
@@ -0,0 +1,90 @@
...
+</script>
+
+<style scoped lang="scss">
+@import 'docc-render/styles/_core.scss';
+
+.header-anchor {
+  color: var(--colors-text, var(--color-text));
+  text-decoration: none;
+
+  // &:after {
```
dobromir-hristov, line 67: 👀

dobromir-hristov (commented):
I tested and it works OK. I think you havent missed any places where we use headings. 

Just make sure you fix what Marcus pointed out, with the jumping text on hover.

marinaaisa:
> I know you're still iterating on the visual design of the link, but I wanted to say that we might want to consider some sort of absolute positioning of it to avoid situations like the following where the inline content needs to reflow sometimes:

@mportiz08 oh true, fixed: https://github.com/apple/swift-docc-render/pull/408/commits/23fd3c74b803cf0a4909674154cb2b2280a090e6

but yes, the design specs aren't closed yet

marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

src/components/ContentNode/LinkableHeading.vue
```diff
@@ -0,0 +1,84 @@
...
+  },
+};
+</script>
+
+<style scoped lang="scss">
+@import 'docc-render/styles/_core.scss';
+$icon-margin: 7px;
+
+.header-anchor {
+  color: var(--colors-text, var(--color-text));
```
mportiz08, line 66: I think we should potentially remove this since these headings are currently being assigned the `--colors-header-text` value in a global/base h1-h6 rule, and this could potentially change the API, even though it's not noticeable at the moment since these colors are probably the same by default.

If removing it isn't sufficient, we should consider changing it to use `--colors-header-text`, but I would recommend just removing this line if possible to make it as close as possible to what it was before.

src/styles/core/_vars.scss
```diff
@@ -34,6 +34,9 @@ $article-stacked-margin-small: 20px;
 $topic-link-icon-width: 1.294rem;
 $topic-link-icon-spacing: 1rem;
 
+// Icons
+$icon-size: 16px;
```
mportiz08, line 38: Feel free to ignore this, but I think a name like `$icon-size-default` might be a little safer, because this one kind of implies that all icons will always be this size. That might be the case now (honestly not 100% sure), but I see that there could be a future situation where we have some icons of different sizes in different places in the UI and this variable name might be confusing then.

mportiz08 (commented):
Awesome work @marinaaisa! This is working great, and I think it's almost ready to merge.

I just noticed one minor potential issue to look at with the color of the new linkable heading element.

src/components/ContentNode/LinkableHeading.vue
dobromir-hristov: This should do it :) It will override the `color` for anchors and inherit the heading color.

```suggestion
  color: inherit;
```

dobromir-hristov (commented):
This looks good to me, just check the things that Marcus pointed out.

src/styles/core/_vars.scss
marinaaisa: done!

src/components/ContentNode/LinkableHeading.vue
marinaaisa: I applied `color: inherit;`, good idea @dobromir-hristov !

SamLanier:
New design approved.

src/components/ContentNode/LinkableHeading.vue
marinaaisa: and thanks @mportiz08 for pointing it out!

marinaaisa:
@swift-ci test

PR #409, opened 2022-08-10
Support the `@Row` directive

dobromir-hristov:
Bug/issue #, if applicable: 97715742

## Summary

This PR adds support for the new `@Row` directive in DocC.

image

image

## Dependencies

DocC PR - https://github.com/apple/swift-docc/pull/378

## Testing

Use the provided JSON file or insert the snippet below:
```json
        {
          "type": "row",
          "numberOfColumns": 3,
          "columns": [
            {
              "size": 2,
              "content": [{
                "type": "text",
                "text": "Span 2"
              }]
            },
            {
              "content": [{
                "type": "text",
                "text": "Span 1 (auto)"
              }]
            }
          ]
        }
```

Steps:
1. Assert the grid works with specifying a column count and without.
2. Assert extra items overflow on a new line.
3. Assert on mobile, everything is equal width, vertical.

## Checklist

- [x] Added tests


dobromir-hristov:
One thing I am not sure I like is the naming. We already have `GridRow` and `GridColumn`... but these new ones use actual `css grid`... What is your opinions?

src/components/ContentNode/Column.vue
```diff
@@ -0,0 +1,41 @@
+<!--
+  This source file is part of the Swift.org open source project
+
+  Copyright (c) 2021 Apple Inc. and the Swift project authors
```
dobromir-hristov, line 4:
```suggestion
  Copyright (c) 2022 Apple Inc. and the Swift project authors
```

src/components/ContentNode/Grid.vue
```diff
@@ -0,0 +1,55 @@
+<!--
+  This source file is part of the Swift.org open source project
+
+  Copyright (c) 2021 Apple Inc. and the Swift project authors
```
dobromir-hristov, line 4:
```suggestion
  Copyright (c) 2022 Apple Inc. and the Swift project authors
```

tests/unit/components/ContentNode/Column.spec.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
```
dobromir-hristov, line 4:
```suggestion
 * Copyright (c) 2022 Apple Inc. and the Swift project authors
```

tests/unit/components/ContentNode/Grid.spec.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
```
dobromir-hristov, line 4:
```suggestion
 * Copyright (c) 2022 Apple Inc. and the Swift project authors
```

src/components/ContentNode/Grid.vue
```diff
@@ -0,0 +1,55 @@
...
+-->
+<template>
+  <div class="grid" :style="style" :class="{ 'with-columns': columns }">
+    <slot />
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'grid',
```
mportiz08, line 18:
```suggestion
  name: 'Grid',
```

Have you considered naming this `Row` to have it more closely map to the actual DocC markdown directive that it corresponds to? (Alternatively, maybe the directive should be called Grid to match this...)

src/components/ContentNode/Grid.vue
```diff
@@ -0,0 +1,55 @@
...
+  props: {
+    columns: {
+      type: Number,
+      default: null,
+      required: false,
+      validator: v => v > 0,
+    },
+  },
+  computed: {
+    style: ({ columns }) => ({ '--col-count': columns }),
```
mportiz08, line 28: What happens when `columns` is `null` here? Should there be a default value of `12` since that is what used in the `grid-template-columns`  when this var isn't defined anyways?

src/components/ContentNode/Column.vue
```diff
@@ -0,0 +1,41 @@
...
+-->
+<template>
+  <div class="column" :style="style">
+    <slot />
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'column',
```
mportiz08, line 18:
```suggestion
  name: 'Column',
```

mportiz08:
> One thing I am not sure I like is the naming. We already have `GridRow` and `GridColumn`... but these new ones use actual `css grid`... What is your opinions?

I agree it's a little annoying but also not a huge deal, especially since these are located within the `ContentNode` folder along with the rest of the content components—should be clear from the surrounding context that these are specific to the Render Node directives.

src/components/ContentNode/Grid.vue
```diff
@@ -0,0 +1,55 @@
...
+  },
+};
+</script>
+
+<style scoped lang='scss'>
+@import 'docc-render/styles/_core.scss';
+
+.grid {
```
mportiz08, line 36: I think you'll need to add some basic spacing rules for this new block element to more closely match the spacing that we have for paragraphs and other things.

Right now, this is what it looks like when I splice your example JSON into an existing doc—notice how the paragraph is squeezed against the grid block.

Screenshot 2022-08-25 at 3 40 41 PM

I think you can probably just match what we're doing for a similar thing like code listings—I imagine it uses the `$stacked-margin-large` variable or something similar using a generic sibling selector.

mportiz08 (changes requested):

src/components/ContentNode/Grid.vue
mportiz08: Like this I think: https://github.com/apple/swift-docc-render/blob/9a608c3d457cb6c7880773941ffac2a8c792b97f/src/components/DocumentationTopic/ContentNode.vue#L43

src/components/ContentNode/Grid.vue
dobromir-hristov: Yeah honestly I dont know which one is better 🤔 It feels like the Vue component should be `Grid`, but then do we call the `Column`s, `Cell`s?

Or do I just call it `Row` and call it a day?

src/components/ContentNode/Grid.vue
dobromir-hristov: Actually this is a good question. That 12 is never used when I think about it.

My idea was that a Grid by default has no column count, so every item you add in there, will take 1fr. It might make sense to add a limit to that though, so things wrap after a certain point 🤔 

Then if you specify a column count, it splits the content into those columns, so you can even have 2x2 and 1x1 on a 5 col grid.

I will look into it again, and see if I can add a limit.

dobromir-hristov: OK I did some changes. Let me know how that looks.

mportiz08: It looks very similar to before. So what happens when it's `null`? Should `12` just be the default since that's the value I still see being used in the CSS when this prop isn't provided?

src/components/ContentNode/Grid.vue
mportiz08: You may want to talk to Ethan about possibly renaming the directive to `Grid` if that's what makes the most sense—I think the directive name and component name would make sense to match each other unless there's a good reason otherwise.

ethan-kusters: It's called a `Row` in Swift-DocC because it defines an individual row. You can't stack multiple cells vertically in it. If the Vue component can contain multiple Swift-DocC rows then I can see the argument for it being called a grid.

```md
This is my grid:

@Row {
  @Column {
    Bla.
  }
  @Column {
    Bla.
  }
}
@Row {
  @Column {
    Bla.
  }
  @Column {
    Bla.
  }
}
@Row {
  @Column {
    Bla.
  }
  @Column {
    Bla.
  }
}
```

ethan-kusters: > but then do we call the Columns, Cells?

Technically the `@Column` directive is more defining a cell- I think for SwiftDocC it's much more intuitive to think about a column belonging to a row just from an authoring perspective. But maybe for the Vue component a `cell` makes more sense.

mportiz08: Sounds like maybe we should rename it to `Row` since this is intended to be used only in support of that directive.

src/components/ContentNode/Grid.vue
dobromir-hristov: I did a small refactor. We need a way to distinguish between a pre-defined N column grid, vs one that accepts arbitrary columns, up to a max of 12.

Let me know if it looks OK the way I have it now.

src/components/ContentNode/Grid.vue
dobromir-hristov: Renamed

src/components/ContentNode/Row.vue
```diff
@@ -0,0 +1,62 @@
...
+  props: {
+    columns: {
+      type: Number,
+      default: null,
+      required: false,
+      validator: v => v > 0 && v <= 12,
+    },
+  },
+  computed: {
+    style: ({ columns }) => ({ '--col-count': columns || 12 }),
```
mportiz08, line 34: Can we make a constant for the `12` value?

mportiz08 (changes requested):
After testing this with some slightly more realistic content (2 paragraphs), I think we'll also need to address some default gutter spacing between columns (both horizontal and vertical) so that the content has some room to breathe.

Here's screenshots of how it looks right now:

Screenshot 2022-08-29 at 2 10 04 PM

Screenshot 2022-08-29 at 2 10 16 PM

src/components/ContentNode/Row.vue
mportiz08: Or if you make 12 the default instead of allowing `null`, then this conditional wouldn't be needed.

dobromir-hristov: if I make it a default, then the grid will always be 12 unless specified. But in the DocC spec, we dont require users to add a column count, they can just add items and they will fit next to each other..

dobromir-hristov:
> After testing this with some slightly more realistic content (2 paragraphs), I think we'll also need to address some default gutter spacing between columns (both horizontal and vertical) so that the content has some room to breathe.

I was actually thinking about that the other day. We dont add any gutter to our GridRow and GridColumn, but those are building blocks, so it makes sense. Here however, I think as users wont have the possibility to add those, we should have nice defaults.

I will make it configurable via a prop, with some sensible defaults.

dobromir-hristov:
Hey @mportiz08 I added the gap and all, but I am not sure if I like how the grid works when no column count is provided, aka auto mode.

Should we ensure we always fit up to a max of 12 columns, no matter how small they are, or do we add some sensible default?

This is something I saw thrown around on comment sections that is pretty clever. We can specify a max number of columns, the gap and the minimum size a column should be, and it should fit as many columns as it can, up to that `--min-col-size`.

12 in general is probably too much but yeah...

```css
grid-template-columns: repeat(auto-fit, minmax(
     min(max(100% / var(--col-count) - var(--gap), var(--min-col-size)), 100%), 1fr
  ));
```

mportiz08:
> Hey @mportiz08 I added the gap and all, but I am not sure if I like how the grid works when no column count is provided, aka auto mode.
> 

The gap is looking good to me visually now, thanks!

> Should we ensure we always fit up to a max of 12 columns, no matter how small they are, or do we add some sensible default?

I may be misunderstanding, but I don't think so. For example, I would expect the following to result in a grid with only 2 columns which are evenly spaced half/half, taking up the full width of the content area:

```markdown
@Row {
  @Column {
    Foo
  }
  @Column {
    Bar
  }
}
```

I think the concept of an "N column" grid only applies when both a `numberOfColumns` argument is provided to the `@Row` directive and a child `@Column` directive contains a numeric `size` argument to enable authors with more control over how the columns are spaced.

For example, if someone wanted to create a row a very small first and last column and a big middle column:

```markdown
@Row(numberOfColumns: 12) {
  @Column(size: 1) {
    Foo
  }
  @Column {
    Bar
  }
  @Column(size: 1) {
    Baz
  }
}
```

That's just based on the way I interpreted the proposal at [Supporting more dynamic content in Swift-DocC reference documentation](https://forums.swift.org/t/supporting-more-dynamic-content-in-swift-docc-reference-documentation/59527) though, so @ethan-kusters please correct me if I misunderstand or I'm missing anything.

It's also just possible that you were asking if we should impose a maximum number of columns, but I'm also not sure about that. I don't _think_ so, because maybe it would be reasonable to specify a 100 numCols with only 2 columns and one of them specifying size 5 to basically allow for a 5% width type of thing...maybe that's too much flexibility though and we want to enforce a 12 max—I'll need @ethan-kusters to give his thoughts there.

> 
> This is something I saw thrown around on comment sections that is pretty clever. We can specify a max number of columns, the gap and the minimum size a column should be, and it should fit as many columns as it can, up to that `--min-col-size`.
> 
> 12 in general is probably too much but yeah...
> 
> ```css
> grid-template-columns: repeat(auto-fit, minmax(
>      min(
>        max(
>          100% / var(--col-count) - var(--gap), 
>          var(--min-col-size)
>        ), 
>        100%
>      ), 
>      1fr
> ));
> ```

Going to take your word for it that the syntax is right there—the grid templating styles are so complex, haha. Sounds like an interesting idea as long as it plays nicely with what is allowed in the new markdown directive and its allowed argument values.

mportiz08:
> > After testing this with some slightly more realistic content (2 paragraphs), I think we'll also need to address some default gutter spacing between columns (both horizontal and vertical) so that the content has some room to breathe.
> 
> I was actually thinking about that the other day. We dont add any gutter to our GridRow and GridColumn, but those are building blocks, so it makes sense. Here however, I think as users wont have the possibility to add those, we should have nice defaults.
> 
> I will make it configurable via a prop, with some sensible defaults.

How do you imagine the prop could be modified? Would that be intended as a possible future theme setting?

dobromir-hristov:
> How do you imagine the prop could be modified? Would that be intended as a possible future theme setting?

I thought that it might be pretty easy for both Content Authors and DoccRender Developers using the `Row` component to build further components to specify the spacing in the future. Its just as easy to provide the gap via css rules, but this just gives a straight forward API.

I actually needed this for #419 but I think the default we added now is good enough for that use case.

dobromir-hristov:
> I may be misunderstanding, but I don't think so. For example, I would expect the following to result in a grid with only 2 columns which are evenly spaced half/half, taking up the full width of the content area:

Almost...

What you explained is exactly how it works now. But what if you add more columns in your scenario? How should it behave? Should it limit you to a max of 12 and wrap? Or should we limit to a `min-width-per-column` scenario? Or do we not limit them at all 🤣 What if their content just does not fit the space that is available for 1/12 of a column.

@SamLanier and @ethan-kusters I need your input on this, so I can wrap up the styling.

```
@Row {
  @Column {
    Foo
  }
  @Column {
    Bar
  }
  ... etc to 15
}
```

SamLanier:
> > I may be misunderstanding, but I don't think so. For example, I would expect the following to result in a grid with only 2 columns which are evenly spaced half/half, taking up the full width of the content area:
> 
> Almost...
> 
> What you explained is exactly how it works now. But what if you add more columns in your scenario? How should it behave? Should it limit you to a max of 12 and wrap? Or should we limit to a `min-width-per-column` scenario? Or do we not limit them at all 🤣 What if their content just does not fit the space that is available for 1/12 of a column.
> 
> @SamLanier and @ethan-kusters I need your input on this, so I can wrap up the styling.
> 
> ```
> @Row {
>   @Column {
>     Foo
>   }
>   @Column {
>     Bar
>   }
>   ... etc to 15
> }
> ```

We do not want to limit the authoring of columns. Yes, this means authors can create some not-so-great layouts, but it should be up to the author (or authoring guidelines) to create the best experience for the content.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
OK I have refactored this to not have limits anymore.

dobromir-hristov:
@swift-ci test

src/components/ContentNode/Row.vue
```diff
@@ -0,0 +1,72 @@
...
+export default {
+  name: 'Row',
+  props: {
+    columns: {
+      type: Number,
+      default: null,
+      required: false,
+      validator: v => v > 0,
+    },
+    gap: {
```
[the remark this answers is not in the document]
dobromir-hristov: Not right now, but I am assuming we may start to, if we start using this as a building block for future components.

src/components/ContentNode/Column.vue
```diff
@@ -0,0 +1,41 @@
...
+  },
+};
+</script>
+
+<style scoped lang='scss'>
+@import 'docc-render/styles/_core.scss';
+
+.column {
+  grid-column: span var(--col-span);
+  min-width: 0;
```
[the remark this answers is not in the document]
dobromir-hristov: Tells the flex/column item to become smaller than it can right now, so it does not overflow. 

If we have images or text that are wider than the column, this will make the column outgrow what it should.

dobromir-hristov:
@swift-ci test

PR #414, opened 2022-08-19
Prefetching routes for load speed improvements.

sanjaiyan-dev:
[Prefetch  resources](https://web.dev/link-prefetch/) to speed up future navigations.

_Sorry if I made any mistakes :(_

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -17,21 +17,27 @@ export default [
     path: '/tutorials/:id',
     name: 'tutorials-overview',
     component: () => import(
-      /* webpackChunkName: "tutorials-overview" */ 'theme/views/TutorialsOverview.vue'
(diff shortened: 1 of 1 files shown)


mportiz08:
@dobromir-hristov could you maybe review this next week? I know you were looking into something in the same realm but focused on the json data and maybe there's some overlap.

dobromir-hristov:
I dont think we need this, as Vue CLI already adds prefetching. 

This is a cool feature that would be useful, if we utilise async loading of components and can tell the browser that after a button is clicked, a modal has to be shown for example.

mportiz08:
Based on the comment from @dobromir-hristov, I'm going to close this for now since it doesn't seem to be needed at present. We can always re-open it if you had additional thoughts. Thanks @sanjaiyan-dev!

PR #416, opened 2022-08-22
Add on this page navigation

dobromir-hristov:
Bug/issue #, if applicable: 97715869

## Summary

Adds a new right sidebar component to the Documentation pages, that displays the current page headings and sections, and tracks the user as he scrolls across them.

image

The sidebar is only visible above 1500px if sidebar is hidden, or when the content width is above `1270px`. This is done intentionally, to keep the content space untouched, so we utilise only the empty space on larger screens.

**Note**
In cases where a section is very small, aka there is not much content under the heading, upon navigating to that section, the section after it may get highlighted instead. 
This happens because the intersection point is around 30% from the top of the screen, hence if a section is small, then the next section after it would probably reach the intersection point and get highlighted.

## Dependencies

NA

## Testing

Steps:
1. Assert that any documentation page gets the navigator rendered.
2. Assert that clicking on anchors navigates to that page section, keeping query parameters.

## Checklist

- [x] Added tests


franklinsch:
When you click on a link in the component and it's not near the top of the page because there isn't enough content below it, the On This Page component highlights the wrong item:

https://user-images.githubusercontent.com/7278429/185942205-c1a5d4f2-4604-4fb5-9f45-981565edbf2c.mov

Is this something that can be resolved?

dobromir-hristov:
The intersection point is 30% of the screen. Unless I move that up, small sections would have that weird behaviour. I already pointed that out in the PR description as a note.

src/components/OnThisPageNav.vue
```diff
@@ -0,0 +1,158 @@
...
+    window.addEventListener('scroll', this.onScroll, false);
+    this.$once('hook:beforeDestroy', () => {
+      window.removeEventListener('scroll', this.onScroll);
+    });
+    // make sure everything is ready
+    await waitFrames(8);
+    this.onScroll();
+  },
+  methods: {
+    onScroll: throttle(function onScroll() {
```
hqhhuang, line 73: It seems like because the code block renders slower than the rest of the page, `onScroll` is then triggered, even when I'm first landing on a page with a URL like: `blahblah#see-also`. If the See Also section is short, then we would highlight the previous section.
Is there a way to always highlight the section in the anchor when the page first loads?

hqhhuang:
Dobri this looks awesome!!
I just have one comment in terms of the aesthetic:
On pages without the darkened hero section, the new "on this page" section looks like it's just "floating" there. In fact, before I navigated to a page with that header, I was gonna ask why we left such a huge gap between the top of the page and the start of the nav. Do you think we could extend the gray border line more to the right and add a header to the nav?

src/components/OnThisPageNav.vue
dobromir-hristov: Think the jump is fixed, take a look.

The scroll detection should work on mounted, so we know which items is the current highlighted one.

tests/unit/components/DocumentationTopic/OnThisPageStickyContainer.spec.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
```
mportiz08, line 9:
```suggestion
*/
```

tests/unit/views/DocumentationTopic.spec.js
```diff
@@ -6,7 +6,7 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
-*/
+ */
```
mportiz08, line 9:
```suggestion
*/
```

src/mixins/scrollToElement.js
```diff
@@ -6,7 +6,7 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
-*/
+ */
```
mportiz08, line 9:
```suggestion
*/
```

src/stores/OnThisPageSectionsStoreBase.js
```diff
@@ -0,0 +1,28 @@
...
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+export default {
+  state: {
+    onThisPageSections: [],
+    currentPageSection: null,
```
mportiz08, line 14: Do you need to keep track of the entire current section object or would just the anchor be enough?

src/styles/core/typography/_font-styles.scss
```diff
@@ -194,5 +194,8 @@ $font-styles: (
   ),
   nav-menu-collapsible: (
     large: (nav_14_14, nav)
+  ),
+  on-this-page: (
```
mportiz08, line 198: Can we just use `body-reduced-tight` instead of adding a new style since they seem to be the same?

src/components/DocumentationTopic.vue
```diff
@@ -45,49 +45,56 @@
           :platforms="platforms" :technologies="technologies"
         />
       </DocumentationHero>
-      <div v-if="showContainer" class="container">
-        <div class="description" :class="{ 'after-enhanced-hero': enhanceBackground }">
-          <RequirementMetadata
-            v-if="isRequirement"
-            :defaultImplementationsCount="defaultImplementationsCount"
+      <div class="doc-content-wrapper">
```
mportiz08, line 48: This diff is really big and hard to tell what actually changed. It seems like you added 2 new wrapper divs surrounding everything after the hero for layout purposes and added the new on this page component to the very end to the outer div, right? Anything else? Just checking since it's hard to tell at a glance.

src/components/OnThisPageNav.vue
```diff
@@ -0,0 +1,160 @@
...
+
+  Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+<template>
+  <div class="OnThisPageNav">
+    <slot
```
mportiz08, line 12: What's the slot used for here?

src/views/DocumentationTopic.vue
```diff
@@ -289,6 +290,9 @@ export default {
         combineVersions(topicDataDefault.schemaVersion), MIN_RENDER_JSON_VERSION_WITH_INDEX,
       ) >= 0
     ),
+    enableOnThisPageNav: ({ isTargetIDE, store }) => (
+      !isTargetIDE && store.state.onThisPageSections.length > 2
```
mportiz08, line 294: I think we might want to consider `> 1` instead of `> 2`. Browsing the documentation for DocC, I was confused at first when the nav didn't show up, because there were only two sections (Overview and Topics), but it seemed like it would have been useful there.

src/components/DocumentationTopic/OnThisPageStickyContainer.vue
```diff
@@ -0,0 +1,60 @@
...
+// 660px content + (170px aside + 22px padding-right)*2 + 28px*2 gutter
+export const ON_THIS_PAGE_CONTENT_BREAKPOINT = 1110;
+// 1080 content + (170px aside + 22px padding-right)
+export const ON_THIS_PAGE_CONTENT_BREAKPOINT_BIG = 1272;
+
+export default {
+  name: 'OnThisPageStickyContainer',
+  inject: ['store'],
+  computed: {
+    isHidden: ({ store }) => {
```
mportiz08, line 26: I'm guessing you already looked into this, but was it not possible to do this with pure CSS? I'm noticing that sometimes the transition between showing/hiding this element can be not the smoothest (but most of the time it looks fine).

src/components/DocumentationTopic/OnThisPageSection.vue
```diff
@@ -33,11 +33,17 @@ export default {
       type: String,
       required: true,
     },
+    level: {
+      type: Number,
+      default: 2,
+    },
   },
   created() {
```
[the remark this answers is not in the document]
dobromir-hristov: Right, I think using the `LinkableHeading` could be a fine candidate to register the sections in the beginning. 

In the future if we have live updating of the content, this may cause issues as things will probably get out of order with, duplication and so on.

src/components/OnThisPageNav.vue
```diff
@@ -0,0 +1,160 @@
...
+    window.addEventListener('scroll', this.onScroll, false);
+    this.$once('hook:beforeDestroy', () => {
+      window.removeEventListener('scroll', this.onScroll);
+    });
+    // make sure everything is ready
+    await waitFrames(8);
+    this.onScroll();
+  },
+  methods: {
+    onScroll: throttle(function onScroll() {
```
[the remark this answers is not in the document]
dobromir-hristov: It would have, but the IntersectionObserver was not getting fired as fast as I needed it. If you scroll fast, then it just gets throttled and the highlighting is often off.

src/components/OnThisPageNav.vue
```diff
@@ -0,0 +1,160 @@
...
+    }, 100),
+    /**
+     * Returns whether the current item or some of its children is active.
+     * @param item
+     * @returns {boolean}
+     */
+    checkIsActive(item) {
+      return item.anchor === this.currentPageSection.anchor;
+    },
+    async scrollTo({ anchor }) {
```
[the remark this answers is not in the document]
dobromir-hristov: Marina extracted this into an extra helper, I am aware.

src/stores/OnThisPageSectionsStoreBase.js
dobromir-hristov: I think we can keep only the anchor yes. It would be a bit more annoying to find the metadata for the current item, but not a big deal.

src/styles/core/typography/_font-styles.scss
dobromir-hristov: Sure can.

src/components/OnThisPageNav.vue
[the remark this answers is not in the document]
mportiz08: Interesting. I'm pretty sure that the tutorials overview page has a very similar system for highlighting the current section in a table of contents using the observer, but maybe there are differences here that make it not work as well.

src/stores/OnThisPageSectionsStoreBase.js
mportiz08: Ok, just curious

src/components/OnThisPageNav.vue
mportiz08: Interesting. Got it, thanks

src/components/DocumentationTopic/OnThisPageStickyContainer.vue
mportiz08: Ok, just checking. I was right that you already looked into it :P

src/stores/OnThisPageSectionsStoreBase.js
dobromir-hristov: I refactored that, so gonna resolve this convo.

src/components/OnThisPageNav.vue
dobromir-hristov: I also refactored this out as well. We can add this later if someone needs it.

src/components/OnThisPageNav.vue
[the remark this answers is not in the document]
dobromir-hristov: Yes, that was my initial thought too, how come it works fine there??? 🤔 

But tutorials are a bit different because they are very long sections that hold IDs, but here we have a mix of sections and headings that have the id on them (soon only headings https://github.com/apple/swift-docc-render/pull/408), and I found it would just entirely skip over the heading as it flies by 🤣

src/views/DocumentationTopic.vue
SamLanier: There are two debatable options here:
1. Always show the OTP (so it doesn't look broken when not present; assume it is always useful info)
2. Remove the OTP when there are >2 items; assume those items are not helpful and it is repetitive info

I can see arguments for both. @mportiz08 and @ethan-kusters I think you two have differing opinions.

src/components/DocumentationTopic/OnThisPageSection.vue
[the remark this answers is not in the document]
mportiz08: Since #408 has now been merged, this PR will need to be rebased and modified to take those changes into account.

src/components/DocumentationTopic/PrimaryContent/GenericContent.vue
```diff
@@ -31,19 +31,20 @@ export default {
   methods: {
     ...ContentNode.methods,
     addOnThisPageSections() {
```
dobromir-hristov, line 33: I assume this would also go away if we go with the Heading registration 🤔 @mportiz08

src/views/DocumentationTopic.vue
ethan-kusters: My main objection here is that it's easy to end up with 2 autogenerated sections (via the declaration and topics section for example) on a page that has little to no content: https://apple.github.io/swift-argument-parser/documentation/argumentparser/argumentvisibility.

We could resolve that by looking at page height or but ignoring autogenerated content but those are both problematic in their own way.

Essentially I'd like to avoid including an "On this Page" for a page that doesn't use the scroll bar to make all the headings visible. If I can already see what's on this page by looking at the page (without scrolling), the extra navigation element is (to me) a little confusing.

dobromir-hristov: 🤔 we could easily check whether there is a scrollbar or not...

marinaaisa (approved):
Tested and looks good! Left minor comments

src/views/DocumentationTopic.vue
mportiz08: It still provides a highlight of the subsections that doesn't require the user manually reading and parsing through what they are—at a glance, they can see what's on the page without the actual paragraph text getting in the way. I wouldn't necessarily say it's a huge benefit or anything, but I don't think it's very distracting on short pages, especially since you have to already have a very wide screen just to see the affordance. For me personally, it would be much more confusing when it sometimes disappears without an obvious reason why. That's just my personal opinion though.

mportiz08: And the scenario I originally pointed out could happen even on a page with lots of non-auto-generated content (like the DocC example). For example, if you have a large overview and a topics section, like this one: https://www.swift.org/documentation/docc/

src/components/DocumentationTopic/PrimaryContent/GenericContent.vue
mportiz08: It could, yeah—it would probably be better/simpler than this method. It will have the same issues with the live-editing scenario though I assume (as this one does).

src/components/DocumentationTopic/OnThisPageSection.vue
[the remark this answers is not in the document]
mportiz08: I think that makes sense, otherwise we kind of have some components like this which are just leftover from the old implementation and it's more confusing to follow how things work since the markup and registration is now separate. We could potentially do that later though if removing it now is too risky of a change. What do you think?

src/components/DocumentationTopic/PrimaryContent/GenericContent.vue
dobromir-hristov: Yeah...

src/components/DocumentationTopic/OnThisPageSection.vue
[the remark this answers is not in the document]
dobromir-hristov: Done, 63aa095c600efef747b029c6a5759214f1fe22df

src/components/DocumentationTopic/PrimaryContent/GenericContent.vue
dobromir-hristov: Done - 63aa095c600efef747b029c6a5759214f1fe22df

dobromir-hristov:
@swift-ci test

dobromir-hristov:
I have updated the code to register sections from within the actual LinkableHeadings.

This works great as long as the entire page gets re-rendered and not parts of it, which is the same issue we had before.

In the future, it might make more sense to refactor this behaviour but for now its fine.

marinaaisa:
> I have updated the code to register sections from within the actual LinkableHeadings.

Nice! Great cleaning!

I believe we are safe to delete this property and the store here, right? https://github.com/apple/swift-docc-render/blob/63aa095c600efef747b029c6a5759214f1fe22df/src/components/DocumentationTopic.vue#L317-L319

Another thing, did you discuss with @hqhhuang her comment about spacing in pages without hero? I would like to know what was the conclusion. I couldn't find your reply, sorry if it's somewhere there. This is Hanqing's comment I'm referring to:

> I just have one comment in terms of the aesthetic:
On pages without the darkened hero section, the new "on this page" section looks like it's just "floating" there. In fact, before I navigated to a page with that header, I was gonna ask why we left such a huge gap between the top of the page and the start of the nav. Do you think we could extend the gray border line more to the right and add a header to the nav?

SamLanier:
> 

The divider line is only for the body content. We could look into extending is to be full width (on the left and the right) so that it could possibly work better for this new feature. @hqhhuang please file separate bug, as this is not a blocker, but something we can look at separately.

src/components/DocumentationTopic/Summary/OnThisPageNav.vue
```diff
@@ -1,90 +0,0 @@
...
-        </ListItem>
-      </List>
-    </nav>
-  </Section>
-</template>
-
-<script>
-import InlineChevronDownCircleIcon from 'theme/components/Icons/InlineChevronDownCircleIcon.vue';
-import { buildUrl } from 'docc-render/utils/url-helper';
-import List from './List.vue';
```
mportiz08, line 36: It's possible that these section/title/list/list-item components can also be deleted now since this might have been the last component using them from the old sidebar (please double check me before you delete them).

mportiz08 (approved):
LGTM.

My only remaining concern is about when the UI gets enabled, but that's just a personal opinion and not a blocker either way. Nice work @dobromir-hristov!

src/components/DocumentationTopic/Summary/OnThisPageNav.vue
dobromir-hristov: Great call. Checked, they are no longer used.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #417, opened 2022-08-22
Support custom page icons and image overrides

dobromir-hristov:
Bug/issue #, if applicable: 97715925

## Summary

Allow overriding the icons and images related to a page. These can be navigator icon, DocumentationHero icon or the Topics links.

image

**Note 1**
This PR sort of builds upon the #402, and more particularly the icon overriding. There should be minimal changed needed for both to co-exist.

## Dependencies

#402
DocC PR - https://github.com/apple/swift-docc/pull/367

## Testing

Use the DocC PR to build an archive with image overrides or use the attached one below

[archive_with_icon_override.zip](https://github.com/apple/swift-docc-render/files/9564779/archive_with_icon_override.zip)

Steps:
1. Point docc-render's serve to the archive of your choice
2. Go to a page that has page overrides, in my case `http://localhost:8081/documentation/docc/formatting-your-documentation-content`
3. Assert the icon gets changed in both Navigator and DocumentationHero.
4. Assert that when visiting `http://localhost:8081/documentation/docc/` the page above also has a changed icon in the Topic links.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -16,6 +16,7 @@
         :enhanceBackground="enhanceBackground"
         :shortHero="shortHero"
         :shouldShowLanguageSwitcher="shouldShowLanguageSwitcher"
+        :iconOverride="references[pageIcon]"
       >
         <template #above-content>
           <slot name="above-hero-content" />
@@ -268,6 +269,10 @@ export default {
     },
     remoteSource: {
       type: Object,
+      required: false,
+    },
+    pageIcon: {
+      type: String,
       required: false,
     },
   },
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -271,8 +271,8 @@ export default {
       type: Object,
       required: false,
     },
-    pageIcon: {
-      type: String,
+    pageImages: {
+      type: Array,
       required: false,
     },
   },
@@ -355,6 +355,15 @@ export default {
       || (primaryContentSections && primaryContentSections.length)
     ),
     tagName: ({ isSymbolDeprecated }) => (isSymbolDeprecated ? 'Deprecated' : 'Beta'),
+    /**
+     * Finds the page icon in the `pageImages` array
+     * @param {Array} pageImages
+     * @returns {String|null}
+     */
+    pageIcon: ({ pageImages = [] }) => {
+      const icon = pageImages.find(({ type }) => type === 'icon');
+      return icon ? icon.reference : null;
+    },
   },
   methods: {
     normalizePath(path) {
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -362,7 +362,7 @@ export default {
      */
     pageIcon: ({ pageImages = [] }) => {
       const icon = pageImages.find(({ type }) => type === 'icon');
-      return icon ? icon.reference : null;
+      return icon ? icon.identifier : null;
     },
   },
   methods: {
--- a/src/components/DocumentationTopic/DocumentationHero.vue
+++ b/src/components/DocumentationTopic/DocumentationHero.vue
@@ -17,7 +17,7 @@
     :style="styles"
   >
     <div class="icon">
-      <NavigatorLeafIcon
+      <TopicTypeIcon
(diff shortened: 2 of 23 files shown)


ethan-kusters:
RenderJSON spec for this change is up here: https://github.com/apple/swift-docc/pull/367/files#diff-5a65fbdff488799d5ec3462fe3145b2fa8d19a8e029cd568d12dc9fd5d613b46.

I've added an `svgID` property that just pulls the first ID we find in the given SVG file to support what we need for the theming infrastructure.

dobromir-hristov:
> 

Awesome. Added support for it.

mportiz08:
I haven't dug into the code review too much yet, but I started playing with this in combination with the corresponding [DocC PR](https://github.com/apple/swift-docc/pull/367) and I'm noticing some potential issues. It seems like only the icon in the navigator is being overwritten, and I think there are a couple of other places where this icon is being shown like the topics section and documentation intro.

Examples:
Screenshot 2022-09-09 at 3 20 01 PM
Screenshot 2022-09-09 at 3 22 55 PM

For those screenshots, I added a new `square.svg` asset to the bundle and included the following directive in an article like so:

```markdown
# Formatting Your Documentation Content

@Metadata {
  @PageImage(purpose: icon, source: "square.svg", alt: "A square icon")
}

Enhance your content's presentation with special formatting and styling for text, links, and other page elements.
```

mportiz08:
And my new icon looks like this for reference:

```svg
<svg viewBox="0 0 14 14" xmlns="http://www.w3.org/2000/svg" id="square">
  <rect x="2" y="2" width="10" height="10" stroke="currentColor" fill="none"></rect>
</svg>
```

Let me know if I misunderstood anything about this is expected to work. Thanks!

dobromir-hristov:
Thanks for the review @mportiz08 I used Ethan's DocC branch to do a more real test, with actual data and I fixed the issues you pointed out.

I am not super happy about how I got similar overriding logic in the two icon components, but I was not sure how to extract that away. I dont want to use mixins, yet we need a way to optionally render the override instead of the actual icon, while still applying classes and styling to both.

mportiz08:
Cool. It looks like the icon is being used in all the right places now that I highlighted in those screenshots.

mportiz08:
What is the expectation for this feature when combined with the (global) icon overriding from the theme-settings feature?

For example, what if I add an overriding icon for a specific article and _also_ override the default icon for all articles using `theme-settings.json`? Personally, I would expect the icon specified in the markdown directive to still have priority and only be shown for that one particular article, but I just want to make sure that's how it's expected to work. I could probably test this out, but I'd need to do a test branch to combine this and the theme-settings PR I think since this isn't directly dependent on that branch as of now.

mportiz08:
> I am not super happy about how I got similar overriding logic in the two icon components, but I was not sure how to extract that away. I dont want to use mixins, yet we need a way to optionally render the override instead of the actual icon, while still applying classes and styling to both.

Would a mixin solve the problem and you just prefer not to use them or is there an issue making them tricky to use here? Is there a solution where introducing a new component would simplify things or is there logic that could be moved to the base `SVGIcon` component to eliminate the repetition? (I'm not sure of the right answer myself off the top of my head, just throwing out some ideas in case something helps)

dobromir-hristov:
> What is the expectation for this feature when combined with the (global) icon overriding from the theme-settings feature?

I think as you, that markdown specified icon should have weight over the globally replaced icon.

> Would a mixin solve the problem and you just prefer not to use them 

A mixin would probably help here, I am just not the biggest fan of them. I will keep you updated.

dobromir-hristov:
OK refactored the duplicated logic out into a provider component, no magic mixins 🤨 

Let me know if this is OK.

marinaaisa:
Code looks good. I've been trying to test it using the `SlothCreator_with_image_overrides.doccarchive` you attached but it seems that the data is not updated or overwritting.
Is the testing steps updated?

dobromir-hristov:
> Is the testing steps updated?

I just went over them and re-build an archive of docc docs for you. You can also try out using the docc PR.

marinaaisa (approved):
Tested and works great!

dobromir-hristov:
@swift-ci test

PR #418, opened 2022-08-22
Add new Card component

dobromir-hristov:
Bug/issue #, if applicable: 97714707

## Summary

This introduces a new Card component, that will be used as a building block for TopicSectionsStyle and Links.

imageimage

Most props are optional, and the cards self-adjust, so they can act as simple or detailed as need be.

## Dependencies

- 

## Testing

NA

## Checklist

- [x] Added tests


mportiz08:
> ## Dependencies
> *
> ## Testing
> NA

😅

mportiz08:
I can try to review the code on its own, but it will be super helpful to have at least some sort of idea for what the spec for this new data might be and ideally some example test data would be awesome.

src/styles/core/colors/_light.scss
```diff
@@ -190,4 +190,9 @@
   --color-tutorial-hero-text: #{dark-color(figure-gray)};
   --color-tutorial-hero-background: #{dark-color(fill)};
   --color-navigator-item-hover: #{change-color(light-color(fill-blue), $alpha: 0.05)};
+
+  --color-card-background: var(--color-fill);
```
mportiz08, line 194: This is a little hard to say without some example data, but is it an expected scenario to have a card without an image? Depending on how common that scenario is, we might want to consider eliminating special color values for them to avoid bloating the very large number of special colors we already have. Ignoring the shadow, the only benefit to including these new variables would be so that these could potentially be specialized on their own through theming, right?

Not necessarily advocating for one approach of the other yet—just making sure we've thought through the common use case before we add new global color variables.

mportiz08 (commented):
Tried to do an initial pass on just code review but will need some more info to start testing this new feature.

src/components/Card.vue
```diff
@@ -0,0 +1,323 @@
...
+      type: String,
+      required: false,
+    },
+    content: {
+      type: Array,
+      required: false,
+    },
+    url: {
+      type: String,
+      required: false,
```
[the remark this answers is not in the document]
dobromir-hristov: I think if you pass an empty url it should just render an inactive Reference. I will add a default state so we dont have undefined and potential errors from Vue.

src/components/Card.vue
```diff
@@ -0,0 +1,323 @@
...
+    ButtonLink,
+  },
+  constants: {
+    CardSize,
+  },
+  inject: {
+    references: { default: () => ({}) },
+  },
+  computed: {
+    titleId: ({ _uid }) => `card_title_${_uid}`,
```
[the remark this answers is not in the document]
dobromir-hristov: Its something we use all over to generate IDs per component instance. Its internal, but it has been there for quite a while and persists in Vue 3 as well.

src/components/Card.vue
```diff
@@ -0,0 +1,323 @@
...
+      >
+        {{ eyebrow }}
+      </div>
+      <div
+        :id="titleId"
+        class="title"
+      >
+        {{ title }}
+      </div>
+      <ContentNode
```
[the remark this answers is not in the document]
dobromir-hristov: I think you may have a point. That would remove the need for having the async import as well, because if you render a card inside a ContentNode you create an infinite loop.

src/components/Card.vue
```diff
@@ -0,0 +1,317 @@
...
+  }
+}
+
+.details {
+  background-color: var(--color-card-background);
+  padding: $details-padding;
+  position: relative;
+  height: var(--card-details-height);
+
+  .small &, & {
```
marinaaisa, line 220: What do you mean by adding the second `&` to it?

src/components/CardCover.vue
```diff
@@ -0,0 +1,79 @@
...
+      <div class="card-cover card-cover--dark" :style="styles.dark" />
+    </slot>
+  </div>
+</template>
+
+<script>
+import imageAsset from 'docc-render/mixins/imageAsset';
+import { normalizeAssetUrl } from 'docc-render/utils/assets';
+
+const toCSSUrl = url => (url ? `url('${normalizeAssetUrl(url)}')` : undefined);
```
marinaaisa, line 24: Maybe we could extract this function into an utility? we also use them in other places like
`bgStyle` computed property in `Tutorials/Hero.vue`

src/components/Card.vue
dobromir-hristov: Transpiles to:
```css
.details, .small .details {

}
```
It is basically a default state

src/styles/core/colors/_dark.scss
```diff
@@ -86,4 +86,6 @@
   --color-syntax-urls: rgb(102, 153, 255);
   --color-tutorial-background: var(--color-fill-tertiary);
   --color-navigator-item-hover: #{change-color(dark-color(fill-blue), $alpha: 0.5)};
+
+  --color-card-shadow: rgb(255, 255, 255, 0.04);
```
marinaaisa, line 90: Isn't the same as: `#{change-color(dark-color(figure-gray), $alpha: 0.4)};` ?

src/components/CardCover.vue
dobromir-hristov: deff a thing we can do..

src/styles/core/colors/_dark.scss
dobromir-hristov: oh lol could be, I just did some of those in a hurry based on design specs

src/components/Card.vue
marinaaisa: if it's the default state, why do you need the `small` class at all since it's not overwriting it?

dobromir-hristov:
Hey I took into account both of your feedback. You can give it another look.

mportiz08:
Not totally sure how I should be testing this.

dobromir-hristov:
> Not totally sure how I should be testing this.

This is part of https://github.com/apple/swift-docc-render/pull/419, unfortunately as of now, there isnt an isolated way to test this specific PR, unless we build a Storybook-like page with example components.

dobromir-hristov:
Tested Xcode as well, all looks good. Updated to latest docc-render and SummitE

src/components/CardCover.vue
```diff
@@ -0,0 +1,77 @@
...
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <div class="card-cover-wrap" :class="{ rounded }">
+    <slot classes="card-cover">
+      <div class="card-cover card-cover--light" :style="styles.light" />
```
marinaaisa, line 14: Why not using the `ImageAsset` component instead? so we could load the images on demand based on the prefers-color-scheme instead of loading both as backgrounds

marinaaisa (changes requested):
I tested it and it looks good!

I added a blocking comment about AX and a suggesting about images.

Thanks Dobri!

src/components/CardCover.vue
dobromir-hristov: Good question. We are using `background-size: cover` to position the images, but I guess I could refactor that to `object-fit`. Stay tuned.

dobromir-hristov:
@swift-ci test

src/components/CardCover.vue
dobromir-hristov: Done.

src/components/Card.vue
```diff
@@ -0,0 +1,308 @@
...
+  inject: {
+    references: { default: () => ({}) },
+  },
+  computed: {
+    titleId: ({ _uid }) => `card_title_${_uid}`,
+    contentId: ({ _uid }) => `card_content_${_uid}`,
+    eyebrowId: ({ _uid }) => `card_eyebrow_${_uid}`,
+    linkAriaTags: ({ titleId, eyebrowId, contentId }) => ({
+      'aria-labelledby': `${titleId} ${eyebrowId}`,
+      'aria-describedby': `${contentId}`,
```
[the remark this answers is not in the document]
dobromir-hristov: Done

dobromir-hristov:
@marinaaisa please review again, I have updated the code to fit both your comments.

dobromir-hristov:
@swift-ci test

marinaaisa (approved):
Looks good!

PR #420, opened 2022-08-23
Add table spanning support

dobromir-hristov:
Bug/issue #, if applicable: 98017748

## Summary

Adds support for Table col and row spanning, based on the initial DocC draft proposal.

image

**Note:**
This is still experimental and needs to be synced with actual DocC RenderJSON. For now think of this as a POC.

## Dependencies

- DocC PR - https://github.com/apple/swift-docc/pull/385

## Testing

Use the provided JSON.

[example_tablespan.json.zip](https://github.com/apple/swift-docc-render/files/9585315/example_tablespan.json.zip)

Steps:
1. Paste it in your doccarchive of choice
2. Visit the page
3. Assert a table is rendered with row and col spans.

## Checklist

- [x] Added tests


src/components/ContentNode.vue
```diff
@@ -100,29 +100,58 @@ function renderNode(createElement, references) {
     ))
   ));
 
-  const renderTableChildren = (rows, headerStyle = TableHeaderStyle.none) => {
+  const renderTableCell = (
+    element, attrs, data, cellIndex, rowIndex, extendedData, cellsToRender,
```
mportiz08, line 104: Do you have any kind of informal spec that you're working off of? That would be really helpful even for reviewing a draft PR. My initial thoughts is that this seems more complicated than what we previously had discussed, but maybe that's necessary based on the latest expectation of the JSON data for this.

dobromir-hristov: I had this - https://github.com/apple/swift-docc/compare/main...QuietMisdreavus:swift-docc:table-span-spec2

This looks like it would send you:

```json
[
  [
   { "rowspan": 1 },
   {}
  ],
  [ 
   {},
   {}
  ]
]
```
where as my idea was:

```json
{
 "0_0": { "rowspan": 1}
}
```

But it felt like it would have to provide data for each cell, and you would still have to create the matrix in order to know which fields to render and which to skip. Unless docc drops extra columns, in which case we talked that would be a breaking change and decided not to do that :/

QuietMisdreavus: Dobri and i had talked offline about his idea, but i hadn't taken the time to draft it into DocC's RenderNode spec. The idea is that you would create the string key from the x/y coordinate of a cell, and check for it in the `extendedData` property. So borrowing from the test data in the PR, that would look like this:

```js
      const extendedData = {
         '0_0': { colspan: 2 },
         '1_2': { rowspan: 2 },
       };
```

The existing table cell data would still contain all `(rows * columns)` cells in the table, just with empty cells for the ones being spanned over, for backwards compatibility with versions of docc-render without this change.

...i might need help writing this version of `extendedData` into an OpenAPI spec, though 😅

mportiz08: Yeah, it may be hard to specify that stringly typed key format in a formal way using OpenAPI. We should make sure to at least document it in the spec though, especially for other renderers that may come along.

QuietMisdreavus: I've drafted an update to the spec for this format here: https://github.com/apple/swift-docc/compare/main...QuietMisdreavus:swift-docc:table-span-spec3

src/components/ContentNode.vue
```diff
@@ -100,29 +100,35 @@ function renderNode(createElement, references) {
     ))
   ));
 
-  const renderTableChildren = (rows, headerStyle = TableHeaderStyle.none) => {
+  const renderTableCell = (
+    element, attrs, data, cellIndex, rowIndex, extendedData,
+  ) => {
+    const { colspan, rowspan } = extendedData[`${rowIndex}_${cellIndex}`] || {};
+    // if either is `0`, then its spanned over and should not be rendered
+    if (colspan === 0 || rowspan === 0) return null;
```
mportiz08, line 108: I'm not entirely sure that we _need_ to add "0" data for anything in the JSON since we can calculate everything by using the indices of the actual cell that is doing the spanning, but this works and makes the code a little shorter I suppose.

src/components/ContentNode.vue
```diff
@@ -100,29 +100,35 @@ function renderNode(createElement, references) {
...
     // tbody with rows where first cell in each row is th, others are td
     case TableHeaderStyle.both: {
       const [firstRow, ...otherRows] = rows;
       return [
         createElement('thead', {}, [
-          createElement('tr', {}, firstRow.map(cell => (
-            createElement('th', { attrs: { scope: 'col' } }, (
-              renderChildren(cell)
-            ))
+          createElement('tr', {}, firstRow.map((cell, ci) => (
```
mportiz08, line 123:
```suggestion
          createElement('tr', {}, firstRow.map((cell, cellIndex) => (
```

For me at least, `ci` is pretty confusing to read on its own without the above context from the render function.

src/components/ContentNode.vue
```diff
@@ -100,29 +100,35 @@ function renderNode(createElement, references) {
...
+            renderTableCell('th', { scope: 'col' }, cell, ci, 0, extendedData)
           ))),
         ]),
-        createElement('tbody', {}, otherRows.map(([firstCell, ...otherCells]) => (
+        createElement('tbody', {}, otherRows.map(([firstCell, ...otherCells], ri) => (
```
mportiz08, line 127:
```suggestion
        createElement('tbody', {}, otherRows.map(([firstCell, ...otherCells], rowIndex) => (
```

mportiz08 (commented):
The JS code for creating the markup looks good, and I like that it's pretty minimal in terms of what actually changed. The empty cells for older versions of DocC without extended data look how I'd expect to as well—nice job!

I do have some concerns about the change in the visual style that is now happening only for tables that have cells that span. It wasn't obvious why there needs to be a difference between normal tables without any. It seems like the width and border of the table is changing, but it isn't obvious why.

src/components/ContentNode/Table.vue
```diff
@@ -35,6 +47,14 @@ export default { name: 'Table' };
   -webkit-overflow-scrolling: touch;
 }
 
+table {
+  border-style: hidden;
+
+  &.spanned {
```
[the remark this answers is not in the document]
dobromir-hristov: CC @SamLanier to explain the reasoning.

src/components/ContentNode/Table.vue
```diff
@@ -44,7 +64,7 @@ export default { name: 'Table' };
   th {
     border-color: var(--color-fill-gray-tertiary);
     border-style: solid;
-    border-width: 1px 0;
+    border-width: var(--table-border-width, 1px 1px);
```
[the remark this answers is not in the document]
dobromir-hristov: Sam wanted to add vertical borders by default to tables now, so I did that but thought it might make sense to expose this as a CSS var, so third party clients can modify it.

src/components/ContentNode.vue
dobromir-hristov: Yep, makes total sense.

dobromir-hristov:
@swift-ci test

marinaaisa (approved):
Code looks good! Very cool

Is this how it's intended to be? It looks a bit strange example but I understand that it is

Screenshot 2022-09-16 at 16 57 28

dobromir-hristov:
@marinaaisa no, I have not updated the example since @QuietMisdreavus changed the spec

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@marinaaisa test it again :)

QuietMisdreavus:
@marinaaisa @dobromir-hristov It should also be noted that the current implementation in Swift-Markdown and Swift-DocC uses `^` symbols as a row-span marker, not `"`. That's part of what's happening here.

src/components/ContentNode/Table.vue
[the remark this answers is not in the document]
SamLanier: @mportiz08 The vertical lines help with parsing more technical data (vs. marketing tables). In this example you showed, you wouldn't need to use any table spanning, you would just have a simple table with empty entries and the table would look right.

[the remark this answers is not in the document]
mportiz08: I was more curious about the difference in width. Previously tables were only as big as needed based on the content and with this change the tables would always be as wide as they could be, regardless of the content, which may work well for some content but not for others.

And more importantly, this change (as-is) would _only_ take place for tables that have cells that span more than a single cell, and not for normal tables, which seems kind of odd to me.

And just to be clear, the first screenshot was without this PR. With this PR and without the table width specific change, this is what that example would look like:
Screenshot 2022-09-16 at 10 50 16 AM

That is more aligned with what I was expecting, as it could still easily adapt the width to fit the given content while still maintaining the span data.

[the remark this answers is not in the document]
dobromir-hristov: I think the reasoning was to match Github's table rendering (which I may have goofed). 

If you want a cell to be 2x wide as the rest, you can `colspan` it with an empty cell and it becomes 2x. 

In our current existing design that will not happen. But by adding `width: 100%; table-layout: fixed;` to the table, it baheves just like that. Yet that is a pretty big breaking change with current tables, so I advised if we are doing this, to only do it to spanned tables.

dobromir-hristov:
> @marinaaisa @dobromir-hristov It should also be noted that the current implementation in Swift-Markdown and Swift-DocC uses ^ symbols as a row-span marker, not ". That's part of what's happening here.

Tested it, works like a charm. Thanks. We just need to clear up some styling confusion and we are good to go.

src/components/ContentNode/Table.vue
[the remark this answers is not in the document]
mportiz08: > In our current existing design that will not happen.

But it will—just like in the screenshot from my most recent comment above. I think that is where my main confusion is from. I may be misunderstanding the issue, sorry!

marinaaisa:
Okay, now it looks as expected :) thanks @dobromir-hristov @QuietMisdreavus !

Screenshot 2022-09-20 at 17 54 06

src/components/ContentNode/Table.vue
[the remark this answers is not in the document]
mportiz08: We discussed this in person and agreed that it would be best to leave this out for now as something that we can either look at wholesale applying to the default table styles or exploring as something that can be an opt-in by using some syntax feature.

src/components/ContentNode.vue
```diff
@@ -267,8 +263,12 @@ function renderNode(createElement, references) {
         return renderFigure(node);
       }
 
-      return createElement(Table, {}, (
-        renderTableChildren(node.rows, node.header)
+      return createElement(Table, {
+        props: {
+          spanned: !!node.extendedData,
```
mportiz08, line 268: Is this still needed since the class isn't being used anymore?

mportiz08 (approved):

src/components/ContentNode.vue
dobromir-hristov: Probably not, just thought it might help future devs, if they want to style these tables in some way with custom styling 🤔

dobromir-hristov:
@swift-ci test

PR #424, opened 2022-08-29
Fixes scrolling the navigator to the active item

dobromir-hristov:
Bug/issue #, if applicable: 99279409

## Summary

Fixes a bug where the navigator does not scroll to the active item, if deeply in to the tree.

## Dependencies

NA

## Testing

Steps:
1. Navigate to a deep item in your docs navigator
2. Assert that when you refresh, i scrolls to the active item

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -889,16 +889,17 @@ export default {
       await waitFrames(1);
       if (!this.$refs.scroller) return;
       // if we are filtering, it makes more sense to scroll to top of list
-      if (this.hasFilter) {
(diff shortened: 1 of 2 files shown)


mportiz08 (approved):
Code change looks good and the fix seems to work well.

Could you briefly explain what was actually going wrong in the code for a deeply nested item just for my own clarification?

dobromir-hristov:
> Could you briefly explain what was actually going wrong in the code for a deeply nested item just for my own clarification?

`this.getChildPositionInScroller(element)` returns `0` if `element` is falsy, because we cannot figure out the position if we dont have an actual element. Now I just added an extra `element && ` to make sure it is there.

PR #427, opened 2022-09-06
Prevent Nav from Trapping while tabbing 

hqhhuang:
Bug/issue #, if applicable: rdar://94044029 ([AX] Localnav should not FocusTrap for documentation pages)

## Summary
For accessibility considerations, when we tab to the end of the nav menu, we should tab to the browser menu bar next (instead of tabbing to the beginning of the nav menu again).

## Dependencies
NA

## Testing
Steps:
1. Medium or small viewport, tab open the nav menu bar
2. tab through the items in the menu, at the end, focus should be brought to the browser address bar. It should not take you to the first link within our nav menu bar. 

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/NavBase.vue
+++ b/src/components/NavBase.vue
@@ -311,11 +311,11 @@ export default {
         scrollLock.unlockScroll(this.$refs.tray);
       }
     },
-    async onExpand() {
+    onExpand() {
       this.$emit('open');
-      await this.$nextTick();
       // hide sibling elements from VO
       changeElementVOVisibility.hide(this.$refs.wrapper);
+      // focus on the toggle to prevent tabbing to links in the body
       this.$refs.axToggle.focus();
     },
     onClose() {
--- a/src/components/NavBase.vue
+++ b/src/components/NavBase.vue
@@ -315,9 +315,10 @@ export default {
         scrollLock.unlockScroll(this.$refs.tray);
       }
     },
-    onExpand() {
+    async onExpand() {
       this.$emit('open');
       // lock focus
+      await this.$nextTick();
       this.focusTrapInstance.start();
       // hide sibling elements from VO
       changeElementVOVisibility.hide(this.$refs.wrapper);
--- a/src/components/NavBase.vue
+++ b/src/components/NavBase.vue
@@ -1,7 +1,7 @@
 <!--
   This source file is part of the Swift.org open source project
 
-  Copyright (c) 2021 Apple Inc. and the Swift project authors
+  Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
   Licensed under Apache License v2.0 with Runtime Library Exception
 
   See https://swift.org/LICENSE.txt for license information
--- a/src/components/NavBase.vue
+++ b/src/components/NavBase.vue
@@ -75,7 +75,6 @@ import onIntersect from 'docc-render/mixins/onIntersect';
 import NavMenuItems from 'docc-render/components/NavMenuItems.vue';
 import BreakpointEmitter from 'docc-render/components/BreakpointEmitter.vue';
 
-import FocusTrap from 'docc-render/utils/FocusTrap';
 import scrollLock from 'docc-render/utils/scroll-lock';
 import { baseNavStickyAnchorId, MenuLinkModifierClasses } from 'docc-render/constants/nav';
 import { isBreakpointAbove } from 'docc-render/utils/breakpoints';
@@ -154,7 +153,6 @@ export default {
       isTransitioning: false,
       isSticking: false,
       noBackgroundTransition: true,
-      focusTrapInstance: null,
       currentBreakpoint: BreakpointName.large,
     };
   },
@@ -196,7 +194,6 @@ export default {
     document.addEventListener('click', this.handleClickOutside);
     this.handleFlashOnMount();
     await this.$nextTick();
-    this.focusTrapInstance = new FocusTrap(this.$refs.wrapper);
   },
   beforeDestroy() {
     window.removeEventListener('keydown', this.onEscape);
@@ -206,7 +203,6 @@ export default {
     if (this.isOpen) {
       this.toggleScrollLock(false);
     }
-    this.focusTrapInstance.destroy();
   },
   methods: {
     getIntersectionTargets() {
@@ -317,17 +313,15 @@ export default {
     },
     async onExpand() {
       this.$emit('open');
-      // lock focus
       await this.$nextTick();
-      this.focusTrapInstance.start();
       // hide sibling elements from VO
       changeElementVOVisibility.hide(this.$refs.wrapper);
+      this.$refs.axToggle.focus();
     },
     onClose() {
       this.$emit('close');
       // stop the scroll lock
       this.toggleScrollLock(false);
-      this.focusTrapInstance.stop();
       changeElementVOVisibility.show(this.$refs.wrapper);
     },
     async handleFlashOnMount() {
--- a/src/utils/changeElementVOVisibility.js
+++ b/src/utils/changeElementVOVisibility.js
@@ -1,7 +1,7 @@
 /**
  * This source file is part of the Swift.org open source project
  *
- * Copyright (c) 2022 Apple Inc. and the Swift project authors
+ * Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
  * Licensed under Apache License v2.0 with Runtime Library Exception
  *
  * See https://swift.org/LICENSE.txt for license information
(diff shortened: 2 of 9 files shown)


hqhhuang:
@swift-ci test

src/components/NavBase.vue
```diff
@@ -315,9 +315,10 @@ export default {
         scrollLock.unlockScroll(this.$refs.tray);
       }
     },
-    onExpand() {
+    async onExpand() {
       this.$emit('open');
       // lock focus
+      await this.$nextTick();
```
dobromir-hristov, line 321: I know why we added this, but do we need it for this fix?

tests/unit/components/Asset.spec.js
```diff
@@ -185,7 +185,7 @@ describe('Asset', () => {
       expect(asset.exists()).toBe(true);
       expect(asset.props()).toEqual({
         alt: image.alt,
-        loading: "lazy",
```
dobromir-hristov, line 188: Lets revert these as they are not related to the PR.

src/utils/changeElementVOVisibility.js
```diff
@@ -1,12 +1,44 @@
...
+import TabManager from 'docc-render/utils/TabManager';
+
+const PREFIX = 'data-original-';
+const ARIA = 'aria-hidden';
+const TABINDEX = 'tabindex';
+
+function setOriginalValue(element, prop) {
+  // make sure that prop isn't cached already
+  let originalValue = element.getAttribute(PREFIX + prop);
+  if (originalValue === undefined || originalValue === null || originalValue === '') {
```
dobromir-hristov, line 19: You can probably just check for `hasAttribute` instead?

src/utils/changeElementVOVisibility.js
```diff
@@ -1,12 +1,44 @@
...
- * Copyright (c) 2021 Apple Inc. and the Swift project authors
+ * Copyright (c) 2022 Apple Inc. and the Swift project authors
  * Licensed under Apache License v2.0 with Runtime Library Exception
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
+import TabManager from 'docc-render/utils/TabManager';
+
+const PREFIX = 'data-original-';
```
dobromir-hristov, line 12: maybe something like this, so its more clear what the prefix is?

```suggestion
const OG_PREFIX = 'data-original-';
```

src/utils/changeElementVOVisibility.js
```diff
@@ -1,12 +1,44 @@
...
+const ARIA = 'aria-hidden';
+const TABINDEX = 'tabindex';
+
+function setOriginalValue(element, prop) {
```
dobromir-hristov, line 16: Maybe call it something like `preserveAttribute` or `cacheAttribute`?

`setOriginalValue` is very vague.

src/utils/changeElementVOVisibility.js
```diff
@@ -1,12 +1,44 @@
...
+}
+
+function retrieveOriginalValue(element, prop) {
+  // get the cached property
+  const originalValue = element.getAttribute(PREFIX + prop);
+
+  // remove the prefixed attribute
+  element.removeAttribute(PREFIX + prop);
+
+  if (typeof originalValue === 'string') {
```
dobromir-hristov, line 32: I wonder why the check if its a `string` 🤔

src/utils/changeElementVOVisibility.js
```diff
@@ -27,43 +59,46 @@ function iterateOverSiblings(el, callback) {
...
-      element.setAttribute(prop, originalValue);
-    } else {
-      // otherwise remove the attribute entirely.
-      element.removeAttribute(prop);
-    }
+  retrieveOriginalValue(element, ARIA);
+  retrieveOriginalValue(element, TABINDEX);
+
+  // make sure element's tabbable children are restored as well
+  const tabbables = element.querySelectorAll(`[${PREFIX + TABINDEX}]`);
```
dobromir-hristov, line 96: `TabManager.getTabbableElements` would not work here now, would it? 🤔 Because they are no longer tabbable.

dobromir-hristov (commented):
I did a quick test and it looks to work good.

The code is OK, just left a few small comments.

src/utils/changeElementVOVisibility.js
hqhhuang: Yup. If we use `getTabbableElements`, the only ones that would get restored would be the ones that are focusable by nature.

tests/unit/components/Asset.spec.js
hqhhuang: Yea I figured. These 2 are generated when I ran test.

hqhhuang: Should we open a separate PR to fix these?

src/utils/changeElementVOVisibility.js
hqhhuang: I tested this. If we call `element.getAttribute` on an attribute that doesn't exist, the return value is `null` and the type is therefore an object. 
In our case, the `cacheOriginalAttribute` ensures that even where there's no original value, we still set that attribute, just with an empty string. So in our case, the type of `originalValue` here should always be a string. However, if we had a bug where we somehow didn't set the attribute correctly, we would want to check the type here.

hqhhuang:
@swift-ci test

src/utils/changeElementVOVisibility.js
```diff
@@ -1,12 +1,44 @@
...
+import TabManager from 'docc-render/utils/TabManager';
+
+const OG_PREFIX = 'data-original-';
+const ARIA = 'aria-hidden';
+const TABINDEX = 'tabindex';
+
+function cacheOriginalAttribute(element, prop) {
+  // make sure that prop isn't cached already
+  let originalValue = element.getAttribute(OG_PREFIX + prop);
+  if (!element.hasAttribute(OG_PREFIX + prop) || originalValue === '') {
```
dobromir-hristov, line 19: Shouldnt we just check  if there is a cached attribute and only cache if there is not?

```suggestion
  if (!element.hasAttribute(OG_PREFIX + prop)) {
```

dobromir-hristov (commented):
LGTM, left a small final question.

src/utils/changeElementVOVisibility.js
marinaaisa: Thanks for the explanation, @hqhhuang , maybe then it's easier to do something like this?

````js

if (originalValue === null) return;

if (originalValue.length) {
  element.setAttribute(prop, originalValue);
} else {
  // otherwise remove the attribute entirely.
  element.removeAttribute(prop);
}
````

File: src/utils/changeElementVOVisibility.js [hunk 9]
```diff
@@ -1,12 +1,44 @@
 /**
  * This source file is part of the Swift.org open source project
  *
- * Copyright (c) 2021 Apple Inc. and the Swift project authors
+ * Copyright (c) 2022 Apple Inc. and the Swift project authors
```
marinaaisa, line 4: I believe we should do `2021-2022` instead of `2022` if we edit a file. :)

src/utils/changeElementVOVisibility.js
```diff
@@ -1,12 +1,44 @@
...
+const TABINDEX = 'tabindex';
+
+function cacheOriginalAttribute(element, prop) {
+  // make sure that prop isn't cached already
+  let originalValue = element.getAttribute(OG_PREFIX + prop);
+  if (!element.hasAttribute(OG_PREFIX + prop) || originalValue === '') {
+    originalValue = element.getAttribute(prop) || '';
+    element.setAttribute(OG_PREFIX + prop, originalValue);
+  }
+}
```
dobromir-hristov, line 23: or maybe even:

```suggestion
function cacheOriginalAttribute(element, prop) {
  // make sure that prop isn't cached already
  if(element.hasAttribute(OG_PREFIX + prop)) return
  // cache the attribute
  const value = element.getAttribute(prop) || '';
  element.setAttribute(OG_PREFIX + prop, value);
}
```

src/utils/changeElementVOVisibility.js
dobromir-hristov: Cant you just check for `if(originalValue)` ? Its always a `string` or `null`, so a simple `truthy` check should work. 

We could also probably store `OG_PREFIX + prop` as a var in the function, so we dont concat it 3 times.

src/utils/changeElementVOVisibility.js
marinaaisa: Is there any reason why this function doesn't look like this?

```js
function cacheOriginalAttribute(element, prop) {
  // make sure that prop isn't cached already
  if (element.hasAttribute(OG_PREFIX + prop)) return;

  const originalValue = element.getAttribute(prop) || '';
  element.setAttribute(OG_PREFIX + prop, originalValue);
}
```
I don't understand why adding the originalValue as a let var before.

marinaaisa: so we wrote the same comment @dobromir-hristov 😆 https://github.com/apple/swift-docc-render/pull/427#discussion_r964889622
actually you wrote it first!

src/components/NavBase.vue
```diff
@@ -317,16 +313,15 @@ export default {
     },
     onExpand() {
       this.$emit('open');
-      // lock focus
-      this.focusTrapInstance.start();
       // hide sibling elements from VO
       changeElementVOVisibility.hide(this.$refs.wrapper);
+      // focus on the toggle to prevent tabbing to links in the body
+      this.$refs.axToggle.focus();
```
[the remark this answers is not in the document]
dobromir-hristov: You click on another element, the axToggle is only when you focus navigate around :/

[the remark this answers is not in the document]
hqhhuang: This is needed to retain focus on the toggle when user expands Nav with mouse, then starts tabbing.

src/utils/changeElementVOVisibility.js
hqhhuang: Because if the element has a cached attribute, but it's empty; we don't want to return here on this line:
`if (element.hasAttribute(OG_PREFIX + prop)) return;`

We would still want to set the attribute. Did I miss something?

src/utils/changeElementVOVisibility.js
hqhhuang: Replied above under Marina's comment: https://github.com/apple/swift-docc-render/pull/427/files#r965056810

src/utils/changeElementVOVisibility.js
hqhhuang: This is a good idea!
`if (!element.hasAttribute(OG_PREFIX + prop)) return;`

hqhhuang:
@swift-ci test

hqhhuang:
@swift-ci test

src/utils/changeElementVOVisibility.js
marinaaisa: then what about `if (!!element.getAttribute(OG_PREFIX + prop)) return;` ?

hqhhuang:
@swift-ci test

dobromir-hristov (approved):

src/components/NavBase.vue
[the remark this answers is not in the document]
dobromir-hristov: I think we introduced a regression here....

PR #429, opened 2022-09-12
Make sure toggling full trees takes groupMarkers into consideration and does not add extra space

dobromir-hristov:
Bug/issue #, if applicable: 99748634

## Summary

Fixes a bug where `alt` + clicking on a tree to deep toggle it, would add extra spaces under each groupMarker.

This is indirectly caused by #355, as `groupMarkers` now have references to items they label, which was not handled properly, when considering toggling.

## Dependencies

NA

## Testing

Steps:
1. Toggle items on/off while holding the `alt` key.
2. Assert items behave as expected, without adding extra space.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -650,7 +650,7 @@ export default {
      * @return {NavigatorFlatItem[]}
      */
     getAllChildren(uid) {
-      const arr = [];
+      const collection = new Set([]);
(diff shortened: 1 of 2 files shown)


marinaaisa (approved):
Tested and works great. Interesting solution using Set instead of arrays. Thanks!

dobromir-hristov:
@swift-ci test

PR #433, opened 2022-09-19
render task list items with a checkbox

QuietMisdreavus:
Bug/issue #, if applicable: rdar://100145156

## Summary

This PR uses new data from Swift-DocC to render GitHub-style task lists, i.e. list items with a check box.

## Dependencies

https://github.com/apple/swift-docc/pull/388

## Testing

See testing instructions on the linked Swift-DocC PR.

## Checklist

- [x] Added tests
- [ n/a ] Updated documentation if necessary


QuietMisdreavus:
@swift-ci Please test

src/components/ContentNode.vue
```diff
@@ -107,11 +107,29 @@ function renderNode(createElement, references) {
     renderNode(createElement, references),
   );
 
-  const renderListItems = items => items.map(item => (
-    createElement('li', {}, (
-      renderChildren(item.content)
-    ))
-  ));
+  const renderListItems = items => items.map((item) => {
```
mportiz08, line 110: This change works great, although it adds a bit of complexity to an already very large render function.

Not a blocker, but would you mind if I possibly take a pass at refactoring this to extract an explicit task-list component that we could also re-use for other things in the future if needed?

src/styles/base/_typography.scss
```diff
@@ -108,6 +108,12 @@ li li {
   font-size: 1em;
 }
 
+// make sure that task list checkboxes don't push the rest of the content to the next line
+li input[type="checkbox"] + p {
```
mportiz08, line 112: I think we should also remove the default bullet that comes from unordered list items in addition to adding the checkbox. Any objections?

src/components/ContentNode.vue
mportiz08: I worked on this a bit if you're curious (99.9% done) and can open a PR against your fork if you'd like.

[diff against your branch](https://github.com/QuietMisdreavus/swift-docc-render/compare/task-lists...mportiz08:swift-docc-render:task-list-component)
[diff against main](https://github.com/apple/swift-docc-render/compare/main...mportiz08:swift-docc-render:task-list-component)

src/styles/base/_typography.scss
QuietMisdreavus: That's fine by me, though i'm unfamiliar with the CSS needed to accomplish that.

src/components/ContentNode.vue
QuietMisdreavus: My one reservation about the component as you've written it (if i understand it correctly) is that it's possible to mix "task list items" with plain "list items", e.g. the task list in this PR description. What happens in your patch if someone writes a "mixed list" like that?

mportiz08: Ah, I didn't realize it was supposed to be possible to mix/match like that, but I think I realize now why there wasn't a separate list type in the JSON to represent this. I can make one small tweak to not show the checkbox for regular items in that scenario.

I think we would generally want to match the behavior of the GitHub rendering where that item just shows up without the checkbox in that scenario, right?

Example:
- [ ] a
- [x] b
- c
- [ ] d

mportiz08: I updated it to work that way with https://github.com/mportiz08/swift-docc-render/commit/c96a2419d6ef1664e990af59a21844c36a21a43d

QuietMisdreavus: I think that works, and likely matches what people expect. You can push your branch directly into mine if you'd like.

mportiz08:
@swift-ci test

mportiz08 (approved):

PR #434, opened 2022-09-20
Add extra offset when scrolling to a section in Documentati…

dobromir-hristov:
Bug/issue #, if applicable: 100028167

## Summary

Adds an extra offset to the scrollTo helper function, only for the documentation page.

## Dependencies

NA

## Testing

Steps:
1. Click on links in OnThisPage nav or other anchors on the page
2. Assert nav is accounted for + extra 10px for doc pages

## Checklist

- [x] Added tests


src/utils/scroll-offset.js
```diff
@@ -0,0 +1,23 @@
...
+
+import { documentationTopicName } from 'docc-render/constants/router';
+
+export const EXTRA_DOCUMENTATION_OFFSET = 10;
+
+/**
+ * @param {import('vue-router').Route} to
+ * @returns {Number}
+ */
+export default function getExtraScrollOffset(to) {
```
mportiz08, line 19: Would it make sense to move this to the existing `router-utils.js` file where we already have utilities related to scrolling (and also offsets for scrolling) as opposed to creating an entirely new `scroll-offset.js` file?

src/utils/scroll-offset.js
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import { documentationTopicName } from 'docc-render/constants/router';
+
+export const EXTRA_DOCUMENTATION_OFFSET = 10;
```
mportiz08, line 13: Can you add a small comment to point out that this is meant to represent pixel units?

src/utils/scroll-offset.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
```
mportiz08, line 4:
```suggestion
 * Copyright (c) 2022 Apple Inc. and the Swift project authors
```

src/utils/scroll-offset.js
```diff
@@ -0,0 +1,23 @@
...
+
+export const EXTRA_DOCUMENTATION_OFFSET = 10;
+
+/**
+ * @param {import('vue-router').Route} to
+ * @returns {Number}
+ */
+export default function getExtraScrollOffset(to) {
+  const { name } = to;
+  const isDocumentation = name.includes(documentationTopicName);
```
mportiz08, line 21: Is it safe to assume that `name` will never be undefined here? I think it is—just double checking.

dobromir-hristov: Yep, we always have a name to our routes.

src/utils/scroll-offset.js
dobromir-hristov: Thought it might be easier for future 3rd party users to extend this function 🤔

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

PR #435, opened 2022-09-20
Adjust syntax for some root relative imports

mportiz08:
The `'docc-render'` prefix should be preferred over `'@'` when relevant to prevent issues with other Vue apps integrating these components.


mportiz08:
@swift-ci test

hqhhuang (approved):

PR #436, opened 2022-09-21
Support `@Links` directive

dobromir-hristov:
Bug/issue #, if applicable: 98312171

## Summary

Supports the new `@Links` directive, in the body of pages.

image

## Dependencies

- https://github.com/apple/swift-docc/pull/396

## Testing

Until we get a DocC PR, use the archive below: [slothcreator.json.zip](https://github.com/apple/swift-docc-render/files/9613959/slothcreator.json.zip)

Use the example Slothcreator main page uploaded above.

Steps:
1. Go to http://localhost:8080/documentation/slothcreator
2. Assert you can see the `Link` tag rendered as Grid and List

## Checklist

- [x] Added tests


dobromir-hristov:
@swift-ci test

marinaaisa (approved):
Tested and code looks great! Thanks!

src/components/ContentNode.vue
```diff
@@ -332,6 +334,14 @@ function renderNode(createElement, references) {
         scopedSlots,
       });
     }
+    case BlockType.links: {
```
mportiz08, line 337: Is there a DocC PR for this somewhere?

This is more of a comment on the Render Node API than these code changes here, but it's starting to be concerning how this `ContentNode` component is beginning to be conflated with types of layout sections. In my opinion, this kind of directive is better expressed as a new type of primary content section rather than a new type of content itself, similar to how the existing Topics section is expressed. The surface area of this component has expanded very rapidly in the last week or two!

src/components/ContentNode/LinksBlock.vue
```diff
@@ -0,0 +1,63 @@
...
+  This source file is part of the Swift.org open source project
+
+  Copyright (c) 2022 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+<template>
+  <TopicsLinkCardGrid
```
mportiz08, line 11: Should we swap the order of this conditional so that it's a little easier to read? Something like `if is {} else {}` as opposed to `if not is {} else {}` like it is here?

src/components/ContentNode.vue
dobromir-hristov: No PR yet, this is just based on some draft specs I got.

src/components/ContentNode/LinksBlock.vue
dobromir-hristov: yep totally.

src/components/ContentNode.vue
mportiz08: We should make a note to test this as soon as one is ready in case there are any changes needed to make.

mportiz08 (approved):

ethan-kusters:
Noticing a few alignment issues when testing with https://github.com/apple/swift-docc/pull/396 but it's generally looking great!

Screenshot 2022-09-28 at 3 40 47 PM

Screenshot 2022-09-28 at 3 39 53 PM

dobromir-hristov:
Great catches @ethan-kusters . Fixed those :)

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #437, opened 2022-09-21
Allow disabling the sidebar resizing and setting a hard width

dobromir-hristov:
Bug/issue #, if applicable: 97716953

## Summary

Allows for disabling the resizing of the AdjustableSidebar and setting a pre-defined width.

## Dependencies

NA

## Testing

Steps:
1. Assert there are no regressions
2. Using VueDevtools, flip the `disableResizing` Boolean to true and assert you cannot resize anymore.

## Checklist

- [x] Added tests


dobromir-hristov:
@swift-ci test

marinaaisa (changes requested):
Hey @dobromir-hristov , thanks for working on this, very useful!

I've been playing with the UI and I noticed a jumping issue when setting the `hardWidth` value to 200 and trying to resize afterwards.

This happens because the `hardWidth` value is lower than the `minWidth`.
The same happens when the `hardWidth` is greater than the `maxWidth`.

I propose to update the `minWidth` and `maxWidth` depending on the `hardWidth` values as:

```js
maxWidth: ({ maxWidthPercent, windowWidth, hardWidth }) => (
  Math.max(hardWidth, calcWidthPercent(maxWidthPercent, windowWidth))
),
minWidth: ({ minWidthPercent, windowWidth, hardWidth }) => (
  Math.min(hardWidth, calcWidthPercent(minWidthPercent, windowWidth))
),
```

so we avoid the jumping issue.
Let me know what you think.

As an alternative, we could also limit the `hardWidth` value so it is within the minWidth and maxWidth always.

dobromir-hristov:
@marinaaisa refactored, thanks for the idea

src/components/AdjustableSidebarWidth.vue
```diff
@@ -113,6 +114,14 @@ export default {
       type: Boolean,
       default: false,
     },
+    disableResizing: {
+      type: Boolean,
+      default: false,
+    },
+    hardWidth: {
```
mportiz08, line 121: This might just be me but "hard" width is a slightly non-obvious term. How about `fixedWith`? Just an idea, not a blocker.

src/components/AdjustableSidebarWidth.vue
```diff
@@ -113,6 +114,14 @@ export default {
       type: Boolean,
       default: false,
     },
+    disableResizing: {
```
mportiz08, line 117: Does this need to be a separate prop? I was thinking that it should probably be implied that the component is not meant to be resizable if you're already setting an explicit width for it.

mportiz08, line 117: Also, I find it a little odd/funny that there's a component called `AdjustableSidebarWidth` that can be configured to _not_ be adjustable, but I completely understand how we got here, haha 😅

Not necessary now, but maybe we should rename the component now that it can be restricted in this way.

src/components/AdjustableSidebarWidth.vue
dobromir-hristov: I like it.

src/components/AdjustableSidebarWidth.vue
dobromir-hristov: hmm.. `DynamicSidebar` ? 🤣 `FlexibleSidebar` ??? 👀

dobromir-hristov: I thought about it more and probably no...

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

PR #438, opened 2022-09-21
Crawl RenderJSON for OnThisPage sections

dobromir-hristov:
Bug/issue #, if applicable: 58295413

## Summary

Crawls the `RenderJSON` on every change to the `topicData`, so we can reduce the possibility of stale items in the `OnThisPageNav`, from minor changes to the data, without a page navigation or language change.

## Dependencies

NA

## Testing

Steps:
1. Navigate across pages
2. Assert `OnThisPage` items are properly generated
3. You can try in Vue Devtools to replace the entire `topicData` with a slightly modified version of the one that is already there, like changing a title only. That way you can see how it will re-crawl the data and reflect that title change in the navigator.

## Checklist

- [x] Added tests


src/constants/ContentSectionAnchors.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
```
marinaaisa, line 4:
```suggestion
 * Copyright (c) 2022 Apple Inc. and the Swift project authors
```

src/components/DocumentationTopic/OnThisPageRegistrator.vue
```diff
@@ -0,0 +1,96 @@
...
+
+  Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <div />
```
marinaaisa, line 12: I'm not a big fan of an empty div... I feel the `OnThisPageRegistrator` could look more like a mixin than a component because it doesn't have any template reason to render or exist.

Thoughts?

dobromir-hristov: Mixins are the devil 😈  🤣

dobromir-hristov: Refactored to use a mixin instead. Let me know if its OK now.

dobromir-hristov:
@swift-ci test

marinaaisa (approved):

src/mixins/onThisPageRegistrator.js
```diff
@@ -0,0 +1,88 @@
...
+        topicSections,
+        defaultImplementationsSections,
+        relationshipsSections,
+        seeAlsoSections,
+      } = topicData;
+      if (primaryContentSections) {
+        primaryContentSections.forEach((section) => {
+          switch (section.kind) {
+          case SectionKind.content:
+            section.content.forEach((subSection) => {
```
mportiz08, line 48: Maybe it's not _strictly_ necessary since headings will generally tend to be at the top-most level of the "content tree" 95% of the time, but this could potentially miss items since it only loops over the top level array of nodes and doesn't do a recursive loop over the full tree.

If you want, you could potentially just import the `ContentNode.methods.forEach` function which will do a recursive loop through the tree nodes (just might to use `bind` to change the context to work with this content array—there may be an example of where this was done in the past).

src/mixins/onThisPageRegistrator.js
```diff
@@ -0,0 +1,88 @@
...
+    topicData: {
+      immediate: true,
+      handler: 'extractOnThisPageSections',
+    },
+  },
+  methods: {
+    shouldRegisterContentSection(subSection) {
+      return subSection.type === 'heading' && subSection.level < 4;
+    },
+    extractOnThisPageSections(topicData) {
```
mportiz08, line 32: Is there a reason why this is done with a watcher and a method as opposed to just a computed property that depends on `topicData`? I assume there is, but I'm just curious.

src/constants/ContentSectionAnchors.js
```diff
@@ -0,0 +1,57 @@
...
+ * Copyright (c) 2022 Apple Inc. and the Swift project authors
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import { SectionKind } from '@/constants/PrimaryContentSection';
+
+export const MainContentSectionAnchors = {
```
mportiz08, line 13: I'm not sure right now of a better name off the top of my head, but I think it's a little confusing to have distinguished groupings called "main" and "primary" since they both imply similar meanings.

src/mixins/onThisPageRegistrator.js
```diff
@@ -0,0 +1,88 @@
...
+export default {
+  watch: {
+    topicData: {
+      immediate: true,
+      handler: 'extractOnThisPageSections',
+    },
+  },
+  methods: {
+    shouldRegisterContentSection(subSection) {
+      return subSection.type === 'heading' && subSection.level < 4;
```
mportiz08, line 30: Small thing but you may want to export the `BlockType` constant from `ContentNode` so you can use `BlockType.heading` instead of the string literal here.

src/stores/DocumentationTopicStore.js
```diff
@@ -25,7 +25,6 @@ export default {
   reset() {
     this.state.preferredLanguage = Settings.preferredLanguage;
     this.resetApiChanges();
-    this.resetPageSections();
```
mportiz08, line 28: This _might_ have been the only place where this method was called. Can it be removed from the base store entirely?

src/views/DocumentationTopic.vue
```diff
@@ -121,7 +122,7 @@ export default {
     Nav: DocumentationNav,
     QuickNavigationModal,
   },
-  mixins: [performanceMetrics, onPageLoadScrollToFragment],
+  mixins: [performanceMetrics, onPageLoadScrollToFragment, OnThisPageRegistrator],
```
mportiz08, line 125: My only concern with making this a mixin is that it is currently only used in a single component (I think). Was the reason for extracting it to avoid making this component really big with all of the extracted logic?

tests/unit/mixins/onThisPageRegistrator.spec.js
```diff
@@ -0,0 +1,153 @@
...
+    wrapper.vm.topicData = {
+      primaryContentSections: [
+        {
+          kind: SectionKind.content,
+          content: contentSections,
+        },
+      ],
+    };
+    await flushPromises();
+    expect(onThisPageSectionsStoreBase.state.onThisPageSections).toHaveLength(2);
```
mportiz08, line 151: Should we also do a check to make sure that it's the right two sections after things are updated? (Maybe overkill, feel free to ignore)

mportiz08 (commented):
Cool. I still want to test this some, but I think this is probably the right approach even though it's not as declaratively controlled by the components anymore.

The main downside is that we'll have to remember to update this logic as new section kinds are introduced in some situations, but that's not all that different from having to remember to utilize a specific component anyways, so 🤷‍♂️

src/mixins/onThisPageRegistrator.js
dobromir-hristov: A computed property should not mutate state. In our case we want to extract the sections and save them to the store.

Unless we switch to a scoped slot and provide the data to the child, dropping the store, I think this is the way to go.

src/mixins/onThisPageRegistrator.js
dobromir-hristov: Yep, good catch.

src/views/DocumentationTopic.vue
dobromir-hristov: Yeah, I thought about keeping things separate. Plus it should help third party clients integrate the mixin into their own View pages.

src/constants/ContentSectionAnchors.js
dobromir-hristov: Right... I dont have a better suggestion either, but they are from different places in the RenderJSON, so I did not want to merge them.

dobromir-hristov:
@swift-ci test

src/mixins/onThisPageRegistrator.js
mportiz08: Yeah, I just meant for calculating the sections themselves, not adding them to the store. Anyways, this is fine as-is—I was just curious.

src/constants/ContentSectionAnchors.js
mportiz08: Keeping them split makes sense. Yeah, not coming up with a better alternative at the moment, but it may be confusing to people.

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #439, opened 2022-09-22
Fix flaky test for throttle

dobromir-hristov:
Bug/issue #, if applicable: 100137207

## Summary

Fixes a flaky test assertion in the `throttle.spec.js` file.

## Dependencies

NA

## Testing

No real way to test this apart from keeping an eye on it.

## Checklist

- [x] Added tests


dobromir-hristov:
@swift-ci test

mportiz08 (commented):
These additional assertions look like fine improvements, but I don't see how they would impact the stability of the test in any way (unless I'm misunderstanding their true purpose).

What do you think about possibly using the fake timer API from jest to replace the `setTimeout` implementation? I think that might be a more reliable way of testing this and might even speed up the tests in general since I don't think they would actually be waiting for the specified timeouts?

Just a thought...not sure if that's the best approach. It would be easier if I was able to reproduce the intermittent failure on my own machine, but I haven't found a reliable way to do that yet.

dobromir-hristov:
The extra assertions are just to be able to catch where fails, if it does it again. The fix itself is the longer at the end, to make sure the function is called.

I think there was some issue with the fake timeouts and promises in Jest, I will try though.

dobromir-hristov:
@mportiz08 i could not get the fake time api working. It was not advancing time as I expected, so I reduced the timeout to just 10ms, as 100 was unnecessary.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

mportiz08:
> @mportiz08 i could not get the fake time api working. It was not advancing time as I expected, so I reduced the timeout to just 10ms, as 100 was unnecessary.

I think I got it working—the tricky part was that the fake timers for this version of jest we're using don't stub `Date.now()` which our `throttle` function relies on.

I opened #442 as a possible alternative to this PR if you want to check it out and make sure I did it right.

mportiz08 (changes requested):
These changes don't address the underlying issue as-is.

PR #445, opened 2022-09-29
Fix issues with icons being too small/big when overriding.

dobromir-hristov:
Bug/issue #, if applicable: 100551981

## Summary

Fixes a minor issue with overridden icons, where they can render too big/small.

## Before
image

## After
image

## Dependencies

NA

## Testing

[docc-build.zip](https://github.com/apple/swift-docc-render/files/9673915/docc-build.zip)

Steps:
1. Mount the compiled docc docs above
2. Visit - http://localhost:8080/documentation/docc
3. Assert that the plus icon in the navigator and topic links look OK

## Checklist

- [x] Added tests


src/components/SVGIcon.vue
```diff
@@ -15,7 +15,7 @@
     xmlns="http://www.w3.org/2000/svg"
     xmlns:xlink="http://www.w3.org/1999/xlink"
   >
-    <use v-if="themeOverrideURL" :href="`${themeOverrideURL}#${themeId}`" />
+    <use v-if="themeOverrideURL" :href="`${themeOverrideURL}#${themeId}`" width="100%" height="100%" />
```
mportiz08, line 18: Am I understanding correctly that this is only needed in the scenario where a referenced `<svg>` specifies its own explicit `width` or `height` attribute? Just trying to make sure I understand the scenario where this is breaking at the moment.

mportiz08 (approved):
Changes seem fine to me 👍

Do you think we should add some basic unit tests for this component now that it is slightly less basic than it previous was (I think that's likely the reason why it doesn't have any now at least).

src/components/SVGIcon.vue
dobromir-hristov: Yep, exactly then.

dobromir-hristov:
> Do you think we should add some basic unit tests

I could have sworn I wrote tests for this component in one of my PRs .... I need to find it 🤣 if not I will add a few.

dobromir-hristov:
@swift-ci test

PR #448, opened 2022-10-05
Allow using all kinds of images for icon overriding. 

dobromir-hristov:
Bug/issue #, if applicable: 99205219

## Summary

Allows using any type of image as an icon override. This means SVGs can be hosted remotely, or locally, with or without an ID attribute, or it can be a PNG.

image

## Dependencies

NA

## Testing

Use the provided archive and assert the PNG icon renders properly - [docc-build.zip](https://github.com/apple/swift-docc-render/files/9717625/docc-build.zip)

Steps:
1. Assert its in the navigator
2. Assert the round icon is in the cards - http://localhost:8080/documentation/docc/formatting-your-documentation-content
3. Assert the round icon is in the Hero- http://localhost:8080/documentation/docc/customizing-the-appearance-of-your-documentation-pages
4. Assert the square icon is also rendered- http://localhost:8080/documentation/docc/formatting-your-documentation-content

## Checklist

- [x] Added tests


ethan-kusters:
> or without an ID attribute

Is there still value in DocC providing the ID if there is one?

dobromir-hristov:
> Is there still value in DocC providing the ID if there is one?

Yeah, because SVGs without an ID will use an inferior rendering mechanism, where we cant change colors and so on.

src/components/DocumentationTopic/TopicLinkBlockIcon.vue
```diff
@@ -9,19 +9,10 @@
...
-      <SVGIcon
-        v-if="shouldUseOverride"
-        v-bind="{ themeId, iconUrl }"
-        class="topic-icon"
-      />
-      <component v-else :is="icon" class="topic-icon" />
-    </div>
-  </IconOverrideProvider>
+  <div class="topic-icon-wrapper" v-if="imageOverride || icon">
+    <OverridableAsset v-if="imageOverride" :imageOverride="imageOverride" class="topic-icon" />
```
mportiz08, line 13: I'm noticing a potential issue when testing this specifically for external SVGs or SVGs without an ID—it looks like the icon is somehow getting sized based on its viewbox, which means that it will always have the same width wherever it is used on the page.

This is a problem for the documentation intro element where we show a bigger version of the same icon that is sized much smaller in the sidebar, for example.

SVG with ID:
Screenshot 2022-10-05 at 4 08 00 PM

SVG without ID:
Screenshot 2022-10-05 at 4 07 28 PM

I think we should try to see if there's a way to fix that with these changes if possible. (The issues with colors still stand but they can at least _partially_ be mitigated by providing dark mode versions)

src/components/OverridableAsset.vue
```diff
@@ -0,0 +1,47 @@
...
+      type: Object,
+      default: null,
+    },
+  },
+  computed: {
+    variants: ({ imageOverride }) => (imageOverride ? imageOverride.variants : []),
+    firstVariant: ({ variants }) => (variants[0]),
+    iconUrl: ({ firstVariant }) => firstVariant && firstVariant.url,
+    themeId: ({ firstVariant }) => firstVariant && firstVariant.svgID,
+    isSameDomain: ({ iconUrl, sameOrigin }) => sameOrigin(iconUrl),
```
mportiz08, line 35: Can we change this to `isSameOrigin` to avoid mixing terminology?

mportiz08 (commented):
Mostly seems to work as I would expect for non-SVG assets, but I found one potential issue with how the sizing works for the SVGs that are rendered using the `` tag as introduced here.

src/components/DocumentationTopic/DocumentationHero.vue
```diff
@@ -154,7 +154,7 @@ $doc-hero-icon-dimension: 250px;
 
   .background-icon {
     color: var(--color-documentation-intro-accent, $doc-hero-icon-color);
-    display: block;
+    display: flex;
```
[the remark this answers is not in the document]
dobromir-hristov: Because otherwise images were not obeying the size restrictions that the Hero was applying.

src/components/DocumentationTopic/TopicLinkBlockIcon.vue
dobromir-hristov: Hey Marcus, I saw what you mean. It was happening because we did not set a width/height to the `picture` element of the Asset. Should be good now.

src/components/DocumentationTopic/DocumentationHero.vue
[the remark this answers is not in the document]
dobromir-hristov: No longer necessary.

src/components/DocumentationTopic/TopicLinkBlockIcon.vue
mportiz08: I'm not sure that it's also needed on the `<picture>` element in addition to the ``.

src/components/OverridableAsset.vue
```diff
@@ -0,0 +1,53 @@
...
+    sameOrigin(url) {
+      if (!url) return false;
+      const urlA = new URL(url, window.location);
+      const urlB = new URL(window.location);
+      return urlA.origin === urlB.origin;
+    },
+  },
+};
+</script>
+<style scoped lang="scss">
```
mportiz08, line 48: I think this whole block can be dropped as the `/deep/ img` addition you made to the `DocumentationHero` styles should suffice.

dobromir-hristov: I added it so we dont have to define this everywhere. Overridden icons in the TopicLinks section were also off.

dobromir-hristov:
Here is an archive with all kinds of icons i could think of, large, small, with and without dimensions and ids
[docc-build.zip](https://github.com/apple/swift-docc-render/files/9732558/docc-build.zip)

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

PR #456, opened 2022-10-11
Selecting group titles on the navigator should take the user to that section on the page

marinaaisa:
Bug/issue #95617147, if applicable: 

## Summary

Selecting group titles on the navigator should take the user to that section on the page

## Dependencies

NA

## Testing

Steps:
1. Run project
2. Go to any group mark in the navigator
3. Click and assert that it takes you to the correct section

## Checklist

- [x] Added tests


src/components/Navigator.vue
```diff
@@ -195,14 +197,18 @@ export default {
...
+          path: parentPath,
+        } = parent;
         // generate a uid to track by
         node.uid = this.hashCode(`${parentUID}+${node.path}_${depth}_${index}`);
         // store the parent uid
         node.parent = parentUID;
         // store the current groupMarker reference
         if (node.type === TopicTypes.groupMarker) {
           node.deprecatedChildrenCount = 0;
+          node.path = `${parentPath}#${anchorize(node.title)}`;
```
mportiz08, line 211: I think this line could cause potential issues in the situation where DocC is already providing an `anchor` value that we are using for the ID of a given topic group since it's not guaranteed that the value will be the same as what gets generated here.

To avoid any potential broken links, I think we would have to check the `topicSections` page data for `anchor` values to copy here or only generate them using `anchorize` in _both_ places if it isn't already provided there.

Ideally we would also separately update DocC to generate this `path` with the same anchor to avoid the need for manually assigning one here also, although that could be a future improvement.

src/components/Navigator.vue
```diff
@@ -42,6 +42,7 @@ import NavigatorCard from 'theme/components/Navigator/NavigatorCard.vue';
 import { INDEX_ROOT_KEY } from 'docc-render/constants/sidebar';
 import { TopicTypes } from 'docc-render/constants/TopicTypes';
 import { getSetting } from 'docc-render/utils/theme-settings';
+import { anchorize } from '@/utils/strings';
```
mportiz08, line 45:
```suggestion
import { anchorize } from 'docc-render/utils/strings';
```

src/components/Navigator.vue
dobromir-hristov: My concerns too, though I dont think docc is emitting IDs for those sections at all. 

https://github.com/apple/swift-docc/blob/61100a442ff1021405e19de8b4b225e576b19cdd/Sources/SwiftDocC/SwiftDocC.docc/Resources/RenderNode.spec.json#L2586-L2614

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -251,13 +252,13 @@ $chevron-width: $nav-card-horizontal-spacing;
...
     font-weight: $font-weight-semibold;
-    // groups dont need the overlay
-    &:after {
-      display: none;
-    }
   }
 }
 
+.link {
+  line-height: 1;
```
dobromir-hristov, line 259:
```suggestion
  min-width: 0;
```

dobromir-hristov: Also here should be block - https://github.com/apple/swift-docc-render/pull/456/files#diff-0770d1c4ed499b9fc69b4a7e9e024e5bc5539eec417365e0364a394a5dc9d3b4R310

And I think we are good.

src/components/Navigator.vue
marinaaisa: Is this what you are meaning @mportiz08 ? https://github.com/apple/swift-docc-render/pull/456/commits/37ce9f9756b96604d1a7d9f62a83249f015d1ff1

(Test need to be fixed, but just to know your opinion)

mportiz08: That is what I was meaning, however I wasn't thinking clearly enough—this will _only_ work for the titles in the navigator that relate to the current page—it won't work for any child pages since we only have the data for the current page to copy the anchors from.

To do this right, I think we would need the navigator index data to include the anchors for all the titles—there isn't a way to do this today and have it work everywhere without that :( sorry

marinaaisa: Yep, true. I'll revert the latest commit and speak with the backend team to provide us the anchor in the navigator data.

marinaaisa:
> Titles are no longer being truncated. The rest is fine. I commented on what I think might fix it.

Fixed at https://github.com/apple/swift-docc-render/pull/456/commits/c9001fa09317f98db45f41eafff017ec52c8505e
Thanks!

marinaaisa:
@swift-ci test

src/components/Navigator.vue
mportiz08: Sounds good. To be clear, I don't think any of this will work as it should without that data though. Technically we could change this to assume the navigator will always be vending this data in anticipation of that happening in the future, although I would be a little wary of merging this if that won't be happening for some time. What do you think?

marinaaisa: I think we should wait for the data to come from the backend before merging this

PR #462, opened 2022-10-21
Make sure videos are as big as their poster image.

dobromir-hristov:
Bug/issue #, if applicable: 101373576

## Summary

Sets the width/height of video to match those of their poster image, if provided. This prevents jumping on the page, if the video and poster are different dimensions.

## Dependencies

NA

## Testing

Steps:
1. Use an archive that has a video with a cover image. The image should be smaller than the video.
2. Assert the video is rendered with the same dimensions as the poster image.
3. Assert that switching dark/light mode follows the dimensions of the poster image.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/VideoAsset.vue
+++ b/src/components/VideoAsset.vue
@@ -15,7 +15,6 @@
     :poster="normalisedPosterPath"
     :muted="muted"
     :width="optimalWidth"
-    :height="optimalHeight"
     playsinline
     @playing="$emit('playing')"
     @pause="$emit('pause')"
@@ -36,6 +35,7 @@ import {
   separateVariantsByAppearance,
   normalizeAssetUrl,
   getIntrinsicDimensions,
+  extractDensities,
 } from 'docc-render/utils/assets';
 import AppStore from 'docc-render/stores/AppStore';
 import ColorScheme from 'docc-render/constants/ColorScheme';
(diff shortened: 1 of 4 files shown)


src/components/VideoAsset.vue
```diff
@@ -121,8 +134,23 @@ export default {
...
   methods: {
     normalizeAssetUrl,
+    async getPosterDimensions(path) {
+      let width = null;
+      let height = null;
+      if (path) {
+        ({ width, height } = await getIntrinsicDimensions(path));
+      }
+      this.optimalWidth = width;
+      this.optimalHeight = height;
```
mportiz08, line 152: I think this `optimalHeight` var is causing some scaling issues in places where the width of the video is larger than the width of the element it is contained in—since you calculate and assign _both_ a with and height attribute to the video element, the height of the video can be way too big in this scenario (since it isn't constrained in the same way that the width is).

To address this, I suggest we simply match what we do for the `` element where we assign the height attribute as `"auto"` when we have a fixed width—that way, we can both assign an ideal width (whether or not it can be used as-is) and the height will scale automatically based on how wide the element actually ends up being (depending on where it ends up in the UI).

I can give you some examples offline of where I'm noticing this issue today.

mportiz08 (changes requested):

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

PR #466, opened 2022-10-26
Make compactGrid a 2 column grid on Medium sizes

dobromir-hristov:
Bug/issue #, if applicable: 101583390

## Summary

The compactGrid style for Topics should become a 2 col grid on Medium sizes, because it cramped atm on smaller sizes. On Mobile its still 1 column.

## Dependencies

NA

## Testing

Steps:
1. Set a TopicsSectionStyle to a page to compactGrid
2. Assert that the grid of card items has columns as follows: 3 on Large, 2 on medium, 1 on small.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic/TopicsLinkCardGrid.vue
+++ b/src/components/DocumentationTopic/TopicsLinkCardGrid.vue
@@ -10,7 +10,10 @@
 
 <template>
   <div class="TopicsLinkCardGrid">
-    <Row :columns="compactCards ? 3 : 2">
+    <Row :columns="{
+      large: compactCards ? 3 : 2,
+      medium: 2,
+    }">
       <Column
         v-for="item in items"
         :key="item.title"
--- a/tests/unit/components/DocumentationTopic/TopicsLinkCardGrid.spec.js
+++ b/tests/unit/components/DocumentationTopic/TopicsLinkCardGrid.spec.js
@@ -6,7 +6,7 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
-*/
+ */
 
 import TopicsLinkCardGrid from '@/components/DocumentationTopic/TopicsLinkCardGrid.vue';
 import { shallowMount } from '@vue/test-utils';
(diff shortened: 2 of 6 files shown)


src/components/DocumentationTopic/TopicsLinkCardGrid.vue
```diff
@@ -10,7 +10,7 @@
 
 <template>
   <div class="TopicsLinkCardGrid">
-    <Row :columns="compactCards ? 3 : 2">
+    <Row :columns="gridSize">
```
mportiz08, line 13: What you have here seems pretty clean but do you think this could be done without the use of `BreakpointEmitter`?

I'm thinking like utilizing one of our breakpoint Sass mixins for the right media query to override the `--col-count` variable for the row?

Maybe it's not possible or is just too messy, but I was just wondering out loud since adding the JS emitter might be a little heavyweight for things that can be solved just with CSS.

mportiz08: It's ok if the answer is no—just curious.

dobromir-hristov: We can deff do that, was just not sure if you would prefer to use the "official" component API, aka the prop, or to try and modify the css vars :) 

We use the breakpoint emitter in a few places, so I thought it might make sense to just have the current breakpoint globally available.

mportiz08: My initial thought was that the emitter component just adds additional JS complexity, and I think we're only currently using it in places where CSS alone won't suffice (but I could be wrong).

It could end up being the case that the CSS approach is actually worse when compared to this. I was mostly just wondering if you had considered it and your thoughts.

If we do add it globally, we may want to consider refactoring other places using it so that we don't have multiple emitters going on the same page that do the same thing.

mportiz08 (approved):

mportiz08:
@swift-ci test

tests/unit/components/DocumentationTopic/TopicsLinkCardGrid.spec.js
```diff
@@ -6,7 +6,7 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
-*/
+ */
```
mportiz08, line 9:
```suggestion
*/
```

mportiz08:
@swift-ci test

mportiz08 (approved):

PR #467, opened 2022-10-27
Make sure content updates, when switching TabNavigator tabs

dobromir-hristov:
Bug/issue #, if applicable: 101627936

## Summary

Makes sure content is updated when switching across tabs in a TabNavigator. If using a `@Video` tag, the video element is currently cached and never re-renders.

## Dependencies

NA

## Testing

Steps:
1. Create a TabNavigator with a video in each
2. Assert that switching tabs also switches the video sources.
3. Play one video and switch tabs. Assert the Replay button becomes Replay and video is chagned.

## Checklist

- [ ] Added tests - Cannot add tests, internal Vue issue.

### DIFF
--- a/src/components/ContentNode/TabNavigator.vue
+++ b/src/components/ContentNode/TabNavigator.vue
@@ -18,7 +18,7 @@
         {{ title }}
       </TabnavItem>
     </Tabnav>
-    <div class="tabs-content">
+    <div class="tabs-content" :key="currentTitle">
       <template v-for="title in titles">
         <slot v-if="title === currentTitle" :name="title" />
       </template>


src/components/ContentNode/TabNavigator.vue
```diff
@@ -18,7 +18,7 @@
         {{ title }}
       </TabnavItem>
     </Tabnav>
-    <div class="tabs-content">
+    <div class="tabs-content" :key="currentTitle">
```
mportiz08, line 21: Don't we want to maintain the state of a played video in a given tab, etc?

Also, I don't totally understand why the issue is only reproducible for videos—using something like text or images seems to work fine. Could you help explain the problem to me?

dobromir-hristov: Good questions. 

To maintain the video playback, we need to refactor this to use v-show, otherwise the v-if-ing will just destroy the video.

As for why it does this... I do not know. In devtools the components actually get the right props and everything, Vue just does not re-render the video component. Really weird honestly... Could be the internal diffing they use?

mportiz08 (approved):

mportiz08:
@swift-ci test

PR #473, opened 2022-11-01
Fix issue with shifting icons in Navigator

dobromir-hristov:
Bug/issue #, if applicable: 101679521

## Summary

This fixes issues with Navigator icons when loaded from remote servers. 

## Dependencies

NA

## Testing

Steps:
1. Assert that overloaded icons look OK in both Navigator and Cards from `@Links` directive

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCardItem.vue
+++ b/src/components/Navigator/NavigatorCardItem.vue
@@ -51,6 +51,7 @@
         :key="item.uid"
         :type="item.type"
         :image-override="item.icon ? navigatorReferences[item.icon] : null"
+        :shouldCalculateOptimalWidth="false"
         :class="className"
       />
       <span
--- a/src/components/Navigator/NavigatorCardItem.vue
+++ b/src/components/Navigator/NavigatorCardItem.vue
@@ -48,6 +48,7 @@
     <template #navigator-icon="{ className }">
       <TopicTypeIcon
         v-if="!apiChange"
+        :key="item.uid"
         :type="item.type"
         :image-override="item.icon ? navigatorReferences[item.icon] : null"
         :class="className"
--- a/src/components/OverridableAsset.vue
+++ b/src/components/OverridableAsset.vue
@@ -8,7 +8,7 @@
   See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
 <template>
-  <ImageAsset v-if="shouldUseAsset" :variants="variants" />
+  <ImageAsset v-if="shouldUseAsset" :variants="variants" :loading="null" />
   <SVGIcon v-else :icon-url="iconUrl" :themeId="themeId" />
 </template>
 <script>
--- a/src/components/OverridableAsset.vue
+++ b/src/components/OverridableAsset.vue
@@ -8,7 +8,10 @@
   See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
 <template>
-  <ImageAsset v-if="shouldUseAsset" :variants="variants" :loading="null" />
+  <ImageAsset
+    v-if="shouldUseAsset"
+    v-bind="{ variants, loading: null, shouldCalculateOptimalWidth }"
+  />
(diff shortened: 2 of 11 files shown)


src/components/OverridableAsset.vue
```diff
@@ -8,7 +8,7 @@
   See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
 <template>
-  <ImageAsset v-if="shouldUseAsset" :variants="variants" />
+  <ImageAsset v-if="shouldUseAsset" :variants="variants" :loading="null" />
```
mportiz08, line 11: This change didn't have a noticeable impact on the rendering for me—I'm still seeing the visual flicker when scrolling the sidebar quickly when there are many items with custom icons.

Looking closer at this, I think that the actual flicker/shift happens when the logic for calculating the optimal width for the `` tag happens here: https://github.com/apple/swift-docc-render/blob/881b0a551e544384f69690dfd26b9abcd457ecb8/src/components/ImageAsset.vue#L179

Since the design of the sidebar dictates the size of the icons for this particular part of the UI, I'm wondering if this could be fixed by simply enforcing a fixed width for overwritten icons (using the img tag) in the sidebar—that should prevent that logic from needing to happen asynchronously and fix the shifting issue maybe?

mportiz08 (changes requested):

src/components/OverridableAsset.vue
dobromir-hristov: I saw two issues, one is flickering, the other is shifting of size.

Once I added that `flex: 1` to the `picture` it fixed the size shifting and the flicker got removed from the `loading=null` thing.

I will do more testing today.

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -48,6 +48,7 @@
     <template #navigator-icon="{ className }">
       <TopicTypeIcon
         v-if="!apiChange"
+        :key="item.uid"
```
dobromir-hristov, line 51: @mportiz08 i think this was the "key" lol...

dobromir-hristov:
OK I think it works fine now. Images are loaded with a bit of a pause if remote, but once cached, they are instant.

src/components/OverridableAsset.vue
mportiz08: The shifting of size issue has been fixed with the latest change to add the `key`.

I'm still seeing an initial flicker as each distinct icon loads when its parent sidebar item is disclosed. Do you think there's anything else we can do to minimize that?

I'm still curious if maybe there's an easy way to avoid calculating the sizes of each image when we already know that they're going to be a certain size in this part of the UI. Could that calculation be causing the first load flicker or is it unavoidable without preloading the images? Looking closer, I'm actually seeing an error for that logic for each icon—maybe it doesn't even work well with SVG assets...

src/components/OverridableAsset.vue
```diff
@@ -8,7 +8,10 @@
   See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
 <template>
-  <ImageAsset v-if="shouldUseAsset" :variants="variants" />
+  <ImageAsset
+    v-if="shouldUseAsset"
+    v-bind="{ variants, loading: null, shouldCalculateOptimalWidth }"
```
mportiz08, line 13: I'm not 100% convinced that the `loading` override is needed since the assets are still being loaded asynchronously for the first time when the nav containing them is expanded anyway, but I haven't done enough testing to say for sure.

mportiz08 (approved):
Things look good to me.

There isn't any shifting anymore and there are no errors from attempting to asynchronously calculate the image sizes.

There is some minimal lag for the images to load for the first time as nav items are disclosed, however it isn't possible to eliminate this completely if the assets are remote—this should be able to be optimized through efficient asset requests with caching, etc.

dobromir-hristov:
@swift-ci test

PR #474, opened 2022-11-01
Fix issue with Navigator closing sibling trees

dobromir-hristov:
Bug/issue #, if applicable: 101803433

## Summary

Fixes an issue where the navigator is closing sibling trees. This is happening because of extra computed property re-calculations, from the recent PR - #396 

The exact cause of the problem is that computed properties re-calculate themselves and all depending computeds and render functions, when any of the dependencies in the chain change.

In our case, the `technology` passed to the `NavigatorDataProvider` changes on each page navigation, even though its the same object visually, it is not the same Object reference. 

This was not such a problem before, because we used to just find the Technology with the children from the already fetched navigator data set, and that object actually never changes. The `technology` on the DocumentationTopic however does change on each page nav, because we change the RenderJSON.

This was causing the flattenedChildren to get recreated on every single page navigation, causing headaches down the line. 

## Dependencies

NA

## Testing

Steps:
1. Mount any archive:
2. Open a few items
3. Start navigating across pages.
4. Assert no items are closed unintentionally

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorDataProvider.vue
+++ b/src/components/Navigator/NavigatorDataProvider.vue
@@ -24,8 +24,13 @@ export default {
       type: String,
       default: Language.swift.key.url,
     },
-    technology: {
-      type: Object,
+    /**
+     * The technology we need to fetch data for.
+     * Important - We are passing just the URL, as the technology object changes
+     * between page navigations, resulting in excess re-calculations on each page change.
+     */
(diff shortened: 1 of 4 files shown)


mportiz08 (approved):

mportiz08:
@swift-ci test

PR #475, opened 2022-11-02
Make sure image spacing inside `figure` is consistent

dobromir-hristov:
Bug/issue #, if applicable: 101632403

## Summary

Adjusts the bottom margin of an `img` when it has a caption, to be close to the caption itself. Moves that spacing to the caption text so spacing is consistent. 

image

## Dependencies

NA

## Testing

Steps:
1. Assert that spacing of an image with or without a caption is the same, considering post/prior items.
2. Assert the caption is closer to the image, than the vertical space between the next element.

## Checklist

- [ ] Added tests - no css only

### DIFF
--- a/src/components/DocumentationTopic/ContentNode.vue
+++ b/src/components/DocumentationTopic/ContentNode.vue
@@ -75,6 +75,17 @@ $docs-code-listing-border-width: 1px !default;
     max-width: 100%;
   }
 
+  // if the image is inside a figure, and caption is after it, remove the bottom margin
+  figure > picture:first-child {
+    img {
+      margin-bottom: 0;
+    }
+    // ensure the figcaption has the same bottom margin as the img top margin
+    + figcaption {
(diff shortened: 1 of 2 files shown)


SamLanier (approved):
Design approved.

src/components/DocumentationTopic/ContentNode.vue
```diff
@@ -75,6 +75,17 @@ $docs-code-listing-border-width: 1px !default;
     max-width: 100%;
   }
 
+  // if the image is inside a figure, and caption is after it, remove the bottom margin
```
mportiz08, line 78: I know you were thinking there might be a better way of writing these styles, but it works for now and we can always revisit it later to refactor. I don't have a better solution that I could come up with quickly. 😅

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

PR #478, opened 2022-11-04
Fix issues with spacing of sibling components 

dobromir-hristov:
Bug/issue #, if applicable: 101955529

## Summary

Fixes issues with spacing out components like `img` or `code-listing`, when in around siblings.

This should make `img` wrapped in a `figure` consistently spaced out.

Also fixes some issues from nesting elements inside `/deep/`, which we did not spot before.

## Dependencies

NA

## Testing

[build.zip](https://github.com/apple/swift-docc-render/files/9938077/build.zip)

Steps:
1. Ensure the refactored elements are now properly spaced apart from siblings.
2. Ensure images with/without figure are properly spaced out. Even if inside rows/columns/tabnavs. You can use the archive attached above.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -273,10 +273,15 @@ function renderNode(createElement, references) {
       }, (
         renderListItems(node.items)
       ));
-    case BlockType.paragraph:
-      return createElement('p', {}, (
+    case BlockType.paragraph: {
+      const hasSingleImage = node.inlineContent.length === 1
+        && node.inlineContent[0].type === InlineType.image;
+      const props = hasSingleImage ? { class: ['inline-image-container'] } : {};
+
+      return createElement('p', props, (
         renderChildren(node.inlineContent)
       ));
+    }
     case BlockType.table:
       if (node.metadata && node.metadata.anchor) {
         return renderFigure(node);
--- a/src/components/ContentNode/Table.vue
+++ b/src/components/ContentNode/Table.vue
@@ -32,10 +32,7 @@ export default {
 @import 'docc-render/styles/_core.scss';
 
 .table-wrapper {
-  & + *,
-  * + & {
-    margin-top: $stacked-margin-xlarge;
-  }
+  @include space-out-between-siblings($stacked-margin-xlarge);
 }
 </style>
 
--- a/src/components/DocumentationTopic/ContentNode.vue
+++ b/src/components/DocumentationTopic/ContentNode.vue
@@ -30,6 +30,10 @@ export default {
 
 $docs-code-listing-border-width: 1px !default;
 
+@function between-elements-selector($el) {
+  @return '* + #{$el}, #{$el} + *'
+}
+
 /deep/ {
   .code-listing {
     background: var(--background, var(--color-code-background));
@@ -37,11 +41,6 @@ $docs-code-listing-border-width: 1px !default;
     border-color: var(--colors-grid, var(--color-grid));
     border-width: var(--code-border-width, $docs-code-listing-border-width);
     border-style: var(--code-border-style, solid);
-
-    & + *,
-    * + & {
-      margin-top: $stacked-margin-xlarge;
-    }
 
     pre {
       padding: $code-block-style-elements-padding;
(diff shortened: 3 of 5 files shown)


src/components/ContentNode.vue
```diff
@@ -273,10 +273,15 @@ function renderNode(createElement, references) {
       }, (
         renderListItems(node.items)
       ));
-    case BlockType.paragraph:
-      return createElement('p', {}, (
+    case BlockType.paragraph: {
```
dobromir-hristov, line 276: Here is what I meant by poking inside. I think its a pretty simple thing we can do that helps us in the long run later.

src/components/ContentNode/Table.vue
```diff
@@ -32,10 +32,7 @@ export default {
 @import 'docc-render/styles/_core.scss';
 
 .table-wrapper {
-  & + *,
-  * + & {
-    margin-top: $stacked-margin-xlarge;
-  }
+  @include space-out-between-siblings($stacked-margin-xlarge);
```
dobromir-hristov, line 35: Refactored for consistency.

src/components/DocumentationTopic/ContentNode.vue
```diff
@@ -30,6 +30,10 @@ export default {
 
 $docs-code-listing-border-width: 1px !default;
 
+@function between-elements-selector($el) {
```
dobromir-hristov, line 33: I was not sure if to use a mixin or a function. Mixins would create multiple declarations, where as here we merged them into one...

Unfortunately using `space-out-between-siblings` does not work, because the resulting rule was:

`.content-node .element + *, * + *.content-node .element` - which is wrong.... I tried making that function more flexible, but using `&` is just not cutting it, when in combination with `/deep/`.

src/components/DocumentationTopic/ContentNode.vue
```diff
@@ -38,11 +42,6 @@ $docs-code-listing-border-width: 1px !default;
     border-width: var(--code-border-width, $docs-code-listing-border-width);
     border-style: var(--code-border-style, solid);
 
-    & + *,
```
dobromir-hristov, line 41: this was just not working, explained in the func above

src/components/DocumentationTopic/ContentNode.vue
```diff
@@ -99,13 +86,6 @@ $docs-code-listing-border-width: 1px !default;
     }
   }
 
-  dl {
```
dobromir-hristov, line 102: this was working, mostly because `p` tags have the same spacing 😆

src/components/Tabnav.vue
```diff
@@ -67,7 +67,7 @@ $-tabnav-top-margin: rem(15px);
 $-tabnav-bottom-margin: rem(25px);
 
 .tabnav {
-  margin: $-tabnav-top-margin 0 $-tabnav-bottom-margin 0;
+  margin: 0 0 $-tabnav-bottom-margin 0;
```
dobromir-hristov, line 70: `TabNavigator` and `EndpointExample` both handle the top spacing. No need to have that here.

src/components/ContentNode.vue
mportiz08: It seems really specific and only needed because we need a certain style to be applied in a certain situation. Are you sure there isn't a cleaner way of achieving this with CSS only? Maybe something that `:has` or `:only-of-type` could help with?

src/components/DocumentationTopic/ContentNode.vue
mportiz08: Is this something we even need to create a function/mixin for? It doesn't seem like it's even saving much typing.

dobromir-hristov: Just felt a bit easier to read and structure.

src/components/ContentNode.vue
dobromir-hristov: I could not think of a way to add optional spacing to the paragraph or the next element, when that paragraph has an img somewhere inside it.

The file structure we have to work with is:

Just an img

```html
<maybe-something />

<p>
  <picture>
     <img />

<something-else />
```

With a figure

```html
<maybe-something />

<p>
  <figure>
    <picture>
       <img />

<something-else />
```

when we have an error in the img

```html
<maybe-something />

<p>
   <img />

```

WIth that in mind, `:has` could work, though I was not sure about browser support at the time - check https://caniuse.com/?search=has, its very new still.

I tried the selector below and it works, though its too broad and I need to clean it up a bit:

```css
p:has(img) + *, * + p:has(img)
```

marinaaisa: Because `has` doesn't have support for Firefox yet, I think the safest option is modifying `BlockType.paragraph` as you did, @dobromir-hristov

src/components/DocumentationTopic/ContentNode.vue
marinaaisa: Even tho it's not a mixin, it may make sense to move this function to a more generic file like _helpers.scss so we can use it outside this component too

dobromir-hristov: The problem is we already have something similar. Its just that one has `/deep/` and uses `&` and this does not, because this is nested already in a component. So in a sense, this makes sense to be used only here :/

marinaaisa: oh, I see... @dobromir-hristov 
Maybe adding this context with a comment in the code would help.

marinaaisa (approved):

dobromir-hristov:
@swift-ci test

PR #485, opened 2022-11-17
Fix performance issues with quick searching on very large collections

dobromir-hristov:
Bug/issue #, if applicable: 102424873

## Summary

This fixes an issue where quick search struggles with searching on very large frameworks. This is due to some Vue internals, like accessing a computed property that returns a none-primitive on each iteration for the collections.

## Dependencies

NA

## Testing

Steps:
1. Load the app with a very large archive
2. Assert that searching for a very common word is as fast as searching for a less common one.

## Checklist

- [ ] ~Added tests~ - this is Vue internals, and we dont do perf tests in unit tests.

### DIFF
--- a/src/components/Navigator/QuickNavigationModal.vue
+++ b/src/components/Navigator/QuickNavigationModal.vue
@@ -158,6 +158,7 @@ export default {
       children,
       fuzzyMatch,
       processedUserInput,
+      childrenMap,
       orderSymbolsByPriority,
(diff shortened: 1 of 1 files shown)


marinaaisa (approved):
Quite impressive how the experience change from this small refactor, thanks a lot, very interesting

dobromir-hristov:
@swift-ci test

PR #486, opened 2022-11-21
Optimize the Navigator searching for large indexes

dobromir-hristov:
Bug/issue #, if applicable: 102430474

## Summary

Fixes performance degradations with large navigator trees, that have tens of thousands of nested items.

## Approach

There are 4 main changes, that improved the speed:

### 1. Access to Computed/Reactive State as little as possible

Especially from within loops, accessing computed props often would cause Vue to track allot of internal dependencies.
This meant storing/providing computed props/state as local vars in functions, but drastically reduces the internal Vue.js dependency tracking.

### 2. Store large arrays/objects that dont change deeply via `Object.freeze`. 

This further reduces the reacting dependency tracking, as frozen objects are not tracked internally.

### 3. Use plain `for` loops where necessary

In order to improve pure JS looping speed, I had to use `for` loops. `reduce` `filter` and `map` offer more readability but are generally multiple times slower. That combined with `Object.entries` on large objects and we end up with unnecessary computations.

### 4. Improved algorithm to determine which items to show, based on reverse looping.

In order to determine whether an item should be renderable or not, we had to do lots of checks up/down the tree, whether any of it's children are already matching and so on. 

I reversed the logic to search from last to first, that way if a deep child is to be rendered, we know for sure all of it's parents will. This means any other matches up the same tree chain can skip any further computations, as all the work is already done by the already looped deep child.

## Dependencies

NA

## Testing

Steps:
1. Make sure opening/closing items is not regressing
5. Make sure searching is not regressing, items are shown properly
6. Make sure Tag generation works properly

## Checklist

- [ ] ~Added tests~ - no need, JS improvements only. We dont perf test those.

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -98,6 +98,7 @@
 </template>
 
 <script>
+/* eslint-disable prefer-destructuring,no-continue,no-param-reassign,no-restricted-syntax */
 import { DynamicScroller, DynamicScrollerItem } from 'vue-virtual-scroller';
 import { clone } from 'docc-render/utils/data';
 import { waitFrames, waitFor } from 'docc-render/utils/loading';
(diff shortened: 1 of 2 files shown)


marinaaisa (approved):
Tested and looks great! thanks for working on it, Dobri!

dobromir-hristov:
@swift-ci test

PR #491, opened 2022-12-06
Fix issue where linking to an anchor on another page doesn't work as expected

mportiz08:
Bug/issue #, if applicable: 102985056

## Summary

This fixes a problematic scenario where a user clicks a link to an anchor on another page and the location isn't updated to the right place (is usually a little too early in the page than where the actual anchor should be).

I believe this is a timing issue where the logic to scroll is happening before the page has fully rendered. To fix this, I've added a line to wait for the next Vue tick _only_ when clicking a link to an anchor on another page (and not other scenarios like if the URL was directly loaded). I'm honestly not sure why adding this wait unconditionally doesn't work in the scenario where the URL is directly loaded—it seems to have the inverse problem.

## Testing

Steps:
1. Run `env VUE_APP_DEV_SERVER_PROXY=https://krilnon.github.io/swift-book npm run serve` to start the dev server using data from the [swift-book](https://krilnon.github.io/swift-book/documentation/tspl/) content.
2. Open http://localhost:8080/documentation/the-swift-programming-language/generics
3. Find the first link to "In-Out Parameters" on the page and click it
4. Verify that the "In-Out Parameters" section heading is visible at the top of the screen, as expected
5. Load http://localhost:8080/documentation/the-swift-programming-language/functions#In-Out-Parameters directly in a new tab/window and verify that the "In-Out Parameters" section heading is visible at the top of the screen, as expected
6. Load http://localhost:8080/documentation/the-swift-programming-language/functions#In-Out-Parameters directly or from a link and refresh the page, then verify that the "In-Out Parameters" section heading is visible at the top of the screen
7. Try other linking scenarios and/or other content to verify that there are no regressions introduced by this change

## Checklist

- [X] Added tests

### DIFF
--- a/src/utils/router-utils.js
+++ b/src/utils/router-utils.js
@@ -59,6 +59,12 @@ export async function scrollBehavior(to, from, savedPosition) {
     const offset = baseNavOffset + apiChangesNavHeight + getExtraScrollOffset(to);
 
     const y = process.env.VUE_APP_TARGET === 'ide' ? 0 : offset;
+
+    // Wait until the page has been rendered if the navigation to a hash is
+    // triggered from a previous route.
+    if (from) {
+      await this.app.$nextTick();
+    }
     return { selector: cssEscapeTopicIdHash(hash), offset: { x: 0, y } };
   }
   if (areEquivalentLocations(to, from)) {
--- a/tests/unit/utils/router-utils.spec.js
+++ b/tests/unit/utils/router-utils.spec.js
@@ -76,6 +76,7 @@ describe('router-utils', () => {
       const routeDocsNoChanges = createRoute(documentationTopicName, {}, 'bar');
 
       const resolved = await scrollBehavior(routeDocsNoChanges, routeBar);
+      expect(mockApp.app.$nextTick).toHaveBeenCalled();
       expect(resolved).toEqual({
         selector: routeDocsNoChanges.hash,
         offset: { x: 0, y: baseNavHeight + EXTRA_DOCUMENTATION_OFFSET },


src/utils/router-utils.js
```diff
@@ -59,6 +59,12 @@ export async function scrollBehavior(to, from, savedPosition) {
     const offset = baseNavOffset + apiChangesNavHeight + getExtraScrollOffset(to);
 
     const y = process.env.VUE_APP_TARGET === 'ide' ? 0 : offset;
+
+    // Wait until the page has been rendered if the navigation to a hash is
+    // triggered from a previous route.
+    if (from) {
```
mportiz08, line 65: @dobromir-hristov, assuming this change looks good to you...do you know _why_ I need this conditional? I would think that adding the `await ...` call would work in all scenarios but it seems to break loading the URL directly for some reason (basically inverting the original problem)

dobromir-hristov: We use `scrollBehavior` in basically two places, the router setup and the `mixins/scrollToElement`.

The problem is the router has no idea an async fetching process is running, so it can await it before scrolling to the element. Which is probably why, we added another call via the `mixins/OnPageScrollToFragment`, which once the `DocumentationTopic` is mounted, calls the `scrollToElement` function.

This on its own is wrong, because we now try to scroll to the right part twice.

Then, I saw we have this call inside `mixins/scrollToElement`

```js
      } = await this.$router.options.scrollBehavior(resolvedRoute.route);
```

This probably changes the `this` context of that `scrollBehavior` method, which is why you may not be able to call `await this.$app.nextTick()` inside.

```diff
Index: src/mixins/scrollToElement.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/mixins/scrollToElement.js b/src/mixins/scrollToElement.js
--- a/src/mixins/scrollToElement.js	(revision 48ec05a4c7969aeaef28479a27c9633311a8f90f)
+++ b/src/mixins/scrollToElement.js	(date 1670315962975)
@@ -19,7 +19,7 @@
       const {
         selector,
         offset,
-      } = await this.$router.options.scrollBehavior(resolvedRoute.route);
+      } = await this.$router.options.scrollBehavior.call(this.$router, resolvedRoute.route);
       const element = document.querySelector(selector);
       if (!element) return null;
 
Index: src/utils/router-utils.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/utils/router-utils.js b/src/utils/router-utils.js
--- a/src/utils/router-utils.js	(revision 48ec05a4c7969aeaef28479a27c9633311a8f90f)
+++ b/src/utils/router-utils.js	(date 1670316291651)
@@ -62,9 +62,7 @@
 
     // Wait until the page has been rendered if the navigation to a hash is
     // triggered from a previous route.
-    if (from) {
-      await this.app.$nextTick();
-    }
+    await this.app.$nextTick();
     return { selector: cssEscapeTopicIdHash(hash), offset: { x: 0, y } };
   }
   if (areEquivalentLocations(to, from)) {
```

dobromir-hristov: We may do something like this at some time, where we await the data fetch promise - https://stackoverflow.com/questions/74008705/how-to-use-vue-router-scroll-behavior-with-page-transition

dobromir-hristov: yep exactly what I said :) Its barely noticeable atm though, so deff something to refine if someone is not super busy.

mportiz08: Hm. I spent a lot of time playing with this today and didn't make any meaningful progress :(

I was able to set up [something](http://github.com/apple/swift-docc-render/compare/main...mportiz08:improve-scrolling-to-anchors) quick/dirty to wait on the fetch requests and also the next tick—however, I still had to wait for an arbitrary number of frames for it to work in all scenarios, possibly because of things like images still taking additional time to load.

dobromir-hristov: I think after it downloads the JSON, it needs a few moments to renders all the images AND the code listings, as those come in a few moments later.

mportiz08:
Closing in favor of #493

PR #493, opened 2022-12-08
Ensure user always scrolls to the right place when linking to subsections of pages

mportiz08:
Bug/issue #, if applicable: 102985056

## Summary

There may be times when a user clicks a link to a subsection of another page and isn't scrolled to the appropriate place—this issue is most likely happening more often on uncached, content heavy pages, especially with many and/or large images since the height of the images may be impacting the scroll location before they have fully loaded.

These 2 explicit changes were made to make sure that scrolling consistently goes to the right place in every scenario:

* Updated the `onpageLoadScrollToFragment` mixin so that the logic is also called in the `updated` hook and not just the `mounted` one. This is needed since following a link to a subsection of one documentation page from another actually updates the main component instead of remounting a new one
* Before scrolling, all images are set to eagerly load so that we can wait for them to be inserted on the page before attempting to scroll—otherwise, we may scroll to the wrong place since the loaded images will change the position on the page we need to scroll to.

## Testing

Steps:
1. Run the dev server with `env VUE_APP_DEV_SERVER_PROXY=https://krilnon.github.io/swift-book npm run serve`
2. Open http://localhost:8080/documentation/the-swift-programming-language/basicoperators
3. Click the link to "Overflow Operators" and verify that the page is scrolled to the right heading
4. Test out lots of other pages and links to subsections throughout this content since there are many examples that don't work very well today.
5. Try other content and browsers and try to test for any possible regressions

## Checklist

- [X] Added tests

### DIFF
--- a/src/constants/ImageLoadingStrategy.js
+++ b/src/constants/ImageLoadingStrategy.js
@@ -8,6 +8,11 @@
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
+// These are possible values for the `loading` attribute of `<img>` elements.
+// When the "eager" loading strategy is used (the default without specifying),
+// images are fetched and rendered on the page as soon as possible. With the
+// "lazy" loading strategy, the images are only fetched and rendered on the page
+// when necessary as the element comes into view in the browser.
 export default {
   eager: 'eager',
   lazy: 'lazy',
--- a/src/constants/ImageLoadingStrategy.js
+++ b/src/constants/ImageLoadingStrategy.js
@@ -1,7 +1,7 @@
 /**
  * This source file is part of the Swift.org open source project
  *
- * Copyright (c) 2021 Apple Inc. and the Swift project authors
+ * Copyright (c) 2022 Apple Inc. and the Swift project authors
  * Licensed under Apache License v2.0 with Runtime Library Exception
  *
  * See https://swift.org/LICENSE.txt for license information
--- a/src/constants/ImageLoadingStrategy.js
+++ b/src/constants/ImageLoadingStrategy.js
@@ -1,0 +1,14 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+export default {
+  eager: 'eager',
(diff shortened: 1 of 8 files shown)


mportiz08:
It's a lot of code for something simple which I don't love and there is still the now hidden issue with `scrollBehavior` duplicating similar logic, but I think this is the only solution that works in every scenario due to the dynamic size of images on pages.

src/mixins/onPageLoadScrollToFragment.js
```diff
@@ -8,13 +8,57 @@
...
+    async scrollToElementIfAnchorPresent() {
+      const { hash } = this.$route;
+      if (!hash) {
+        return;
+      }
+
+      // Use "eager" loading for all images since they need to be loaded so
+      // that any dynamic height adjustments can be accounted for
+      const { imageLoadingStrategy } = AppStore.state;
+      AppStore.setImageLoadingStrategy(ImageLoadingStrategy.eager);
```
dobromir-hristov, line 50: would changing the loading mid render actually make a difference?

dobromir-hristov (approved):
Looks pretty cool. Code checks out, works great.

image

src/mixins/onPageLoadScrollToFragment.js
mportiz08: I'll do some more testing to make sure, but I think it was necessary. Without it, some images might not load at all until they're visible due to the default lazy loading strategy, and this code would get stuck waiting forever.

mportiz08: (Yeah, as I expected, it doesn't seem to work for me when commenting out the code related to toggling the loading strategy.)

dobromir-hristov: thanks for looking into it

marinaaisa (approved):
Tested and works great. I didn't know that lazy images could be related to this linking issue. Great finding :)

mportiz08:
@swift-ci test

src/constants/ImageLoadingStrategy.js
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Copyright (c) 2022 Apple Inc. and the Swift project authors
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+export default {
+  eager: 'eager',
```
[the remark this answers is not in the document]
mportiz08: Added some documentation: 210dd485e99690dd8c2307b507fcee418d2f9409

PR #494, opened 2022-12-08
throttle.js refactor

mertcanaltin:
In this code, we use the clearTimeout() function to reset the timer and keep the lastRanTime variable to record the time it was last called. In this way, the code becomes more understandable and readable.

### DIFF
--- a/src/components/DocumentationTopic/OnThisPageStickyContainer.vue
+++ b/src/components/DocumentationTopic/OnThisPageStickyContainer.vue
@@ -24,14 +24,18 @@ export default {
 @import 'docc-render/styles/_core.scss';
 
 .OnThisPageStickyContainer {
+  $top: $nav-height + rem(10px);
   margin-top: $contenttable-spacing-single-side;
   position: sticky;
-  top: $nav-height + rem(10px);
+  top: $top;
   align-self: flex-start;
   flex: 0 0 auto;
   width: $on-this-page-aside-width;
   padding-right: $nav-padding;
   box-sizing: border-box;
+  padding-bottom: $stacked-margin-small;
+  max-height: calc(100vh - #{$top});
+  overflow: auto;
 
   @media print {
     display: none;
--- a/src/utils/throttle.js
+++ b/src/utils/throttle.js
@@ -16,20 +16,23 @@
  */
 export default function throttle(func, limit) {
   let timer;
-  let ranLastTimeOn;
-  return function innerThrottle(...args) {
-    const context = this;
-    if (!ranLastTimeOn) {
-      func.apply(context, args);
-      ranLastTimeOn = Date.now();
+  let lastRanTime;
+
+  return function(...args) {
+    const currentTime = Date.now();
+
+    if (!lastRanTime) {
+      func.apply(this, args);
+      lastRanTime = currentTime;
       return;
     }
+
     clearTimeout(timer);
     timer = setTimeout(() => {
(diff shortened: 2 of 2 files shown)


mportiz08:
@swift-ci test

mportiz08 (changes requested):
Hi @mertcanaltin — thanks for the PR!

I'm noticing that the unit tests for this function are no longer passing. Can you see if there's a potential issue with the code or if the unit tests need to be rewritten based on these changes? Once they're passing, we can better review the code. Thanks!

You can run the full unit test suite with `npm run test:unit` or just the throttle specs with `npx vue-cli-service test:unit tests/unit/utils/throttle.spec.js`

Example:
```shell
λ npx vue-cli-service test:unit tests/unit/utils/throttle.spec.js 
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
 FAIL  tests/unit/utils/throttle.spec.js
  throttle
    ✕ calls a function only once within the given interval of time (5ms)

  ● throttle › calls a function only once within the given interval of time

    expect(jest.fn()).toHaveBeenNthCalledWith(n, ...expected)

    n: 2
    Expected: 49
    Received
           1: 0
    ->     2: 50

    Number of calls: 2

      49 |     expect(func).toHaveBeenCalledTimes(2);
      50 |     expect(func).toHaveBeenNthCalledWith(1, 0);
    > 51 |     expect(func).toHaveBeenNthCalledWith(2, 49);
         |                  ^
      52 |   });
      53 | });
      54 |

      at Object.<anonymous> (tests/unit/utils/throttle.spec.js:51:18)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        0.733s, estimated 1s
Ran all test suites matching /tests\/unit\/utils\/throttle.spec.js/i.
```

dobromir-hristov (commented):
Hey. From what I see, you just have a single call to Date.now(), which I think is fine.

There might be slight issues with the test, because now inside the setTimeout callback, you no longer call Date.now(), but use the one that was called outside the callback, which technically happened one tick earlier. 

We are talking js cycles here, so in a real world we should not see a diff. You just need to fix the test, as it might be asserting in a wonky way.

mertcanaltin:
> Hey. From what I see, you just have a single call to Date.now(), which I think is fine.
> 
> There might be slight issues with the test, because now inside the setTimeout callback, you no longer call Date.now(), but use the one that was called outside the callback, which technically happened one tick earlier.
> 
> We are talking js cycles here, so in a real world we should not see a diff. You just need to fix the test, as it might be asserting in a wonky way.

uh yes thank you very much I applied a solution as you said for date.now thank you very much

mertcanaltin:
> Hi @mertcanaltin , thanks a lot for your PR!

hello, thank you very much for your message, yes, I definitely agree with you, unfortunately, the development did not go as I planned, is it possible to assign me from todo issues? I want to contribute @marinaaisa

mertcanaltin:
thank you for answers team 🚀 ❤️

marinaaisa:
> hello, thank you very much for your message, yes, I definitely agree with you, unfortunately, the development did not go as I planned, is it possible to assign me from todo issues? I want to contribute @marinaaisa

For contribution guidelines I recommend you reading: https://github.com/apple/swift-docc-render/blob/main/CONTRIBUTING.md

We look forward to any future contribution from you! :)

Issue #489, opened 2022-12-04
[Quick Navigation] Pages that are curated multiple times appear duplicated in search results


ethan-kusters:
### Description

Searching for "defining" in [ArgumentParser](https://staging.swiftpackageindex.com/apple/swift-argument-parser/main/documentation/argumentparser) produces duplicated results:

<img width="1436" alt="Screenshot 2022-12-04 at 10 13 38 AM" src="https://user-images.githubusercontent.com/6750147/205508132-468156d2-7824-4c3c-a976-0a7063b3a1bc.png">

This is likely because that page appears twice when filtering in the navigator:

<img width="1436" alt="Screenshot 2022-12-04 at 10 15 20 AM" src="https://user-images.githubusercontent.com/6750147/205508157-b97cb5b7-f9c8-4ad8-8b8a-84ef4daa3a55.png">

Because it's curated by the documentation author on both the top-level page and as a child of the `ParsableCommand` symbol. For the purposes of Quick Navigation, we should filter out duplicate results like this.

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc-render/issues).

### Expected Behavior

I would expect that each page only appears once in the Quick Navigation search results.

### Actual behavior

The "Defining Commands and Subcommands" page appears twice.

### Steps To Reproduce

1. Navigate to https://staging.swiftpackageindex.com/apple/swift-argument-parser/main/documentation/argumentparser
2. Search "defining"

ethan-kusters:
CC: @sofiaromorales @marinaaisa

dobromir-hristov:
Even though its the same page, it is unique because it has different parents, at diff levels and so on, so its UID is diff. 

I guess at the last step, where we slice off the first 20 items, we could check for uniqueness by the URL?

sofiaromorales:
> I guess at the last step, where we slice off the first 20 items, we could check for uniqueness by the URL?

+1 to this, we can filter them by URL uniqueness and keep the top-level one

PR #496, opened 2022-12-13
Bugfix: Remove repeated symbols from QN search results

sofiaromorales:
Bug/issue #489, if applicable: 

## Summary

When an author curates the documentation symbols these might appear repeated times in the index, even if they have different UIDs conceptually they are the same, so when filtering symbols on quick navigation we have to check path uniqueness and remove the duplicates.

In this PR I tweaked the `filteredSymbols` computed property to filter symbols by path uniqueness, removing the duplicates from the results

## Dependencies

N/A

## Testing

_Describe how a reviewer can test the functionality of your PR. Provide test content to test with if
applicable._

Steps:
1. Turn on the feature flag for Quick Navigation
2. Assert that symbols with the same path appear once on Quick Navigation results

## Checklist

- [X] Added tests
- [ ] Updated documentation if necessary (N/A)

### DIFF
--- a/src/components/Navigator/QuickNavigationModal.vue
+++ b/src/components/Navigator/QuickNavigationModal.vue
@@ -172,8 +172,12 @@ export default {
         processedInputRegex: new RegExp(constructFuzzyRegex(processedUserInput), 'i'),
         childrenMap,
       });
-      // Return the first 20 symbols out of sorted ones
-      return orderSymbolsByPriority(matches).slice(0, 20);
+      // Filter symbols with repeated paths and return the first 20
+      return orderSymbolsByPriority(
+        matches.filter((s1, index) => (
+          index === matches.findIndex(s2 => s2.path === s1.path)
+        )),
+      ).slice(0, 20);
     },
     isVisible: {
       get: ({ showQuickNavigationModal }) => showQuickNavigationModal,
--- a/src/components/Navigator/QuickNavigationModal.vue
+++ b/src/components/Navigator/QuickNavigationModal.vue
@@ -172,12 +172,9 @@ export default {
         processedInputRegex: new RegExp(constructFuzzyRegex(processedUserInput), 'i'),
         childrenMap,
       });
-      // Filter symbols with repeated paths and return the first 20
-      return orderSymbolsByPriority(
-        matches.filter((s1, index) => (
-          index === matches.findIndex(s2 => s2.path === s1.path)
-        )),
-      ).slice(0, 20);
+      // Filter repeated matches and return the first 20
+      const uniqueMatches = [...new Map(matches.map(match => [match.path, match])).values()];
+      return orderSymbolsByPriority(uniqueMatches).slice(0, 20);
     },
     isVisible: {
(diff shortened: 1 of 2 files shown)


sofiaromorales:
CC: @marinaaisa

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -172,8 +172,12 @@ export default {
         processedInputRegex: new RegExp(constructFuzzyRegex(processedUserInput), 'i'),
         childrenMap,
       });
-      // Return the first 20 symbols out of sorted ones
-      return orderSymbolsByPriority(matches).slice(0, 20);
+      // Filter symbols with repeated paths and return the first 20
+      return orderSymbolsByPriority(
+        matches.filter((s1, index) => (
```
marinaaisa, line 177: Instead of filtering inside the call of `orderSymbolsByPriority` I suggest creating an `uniqueMatches` const outside and then pass it to `orderSymbolsByPriority`.

The comment of `// Filter symbols with repeated paths and return the first 20` should be on top of the new const instead of `orderSymbolsByPriority`

tests/unit/components/Navigator/QuickNavigationModal.spec.js
```diff
@@ -310,4 +310,26 @@ describe('QuickNavigationModal', () => {
     });
     expect(wrapper.vm.processedUserInput).toBe('barfoo');
   });
+  it('it removes filtered symbols with duplicate paths', () => {
+    const symbolsWithRepeatedPaths = [
```
marinaaisa, line 314: Minor comment: you could add another obj with a different path to the `symbolsWithRepeatedPaths` array so you can make sure that you are not returning duplicate ones instead of filtering them out.

marinaaisa (commented):
Thanks a lot @sofiaromorales , it looks very good 👏 
I added minor comments.

Other way of filtering the duplicated matches by path using ES6+ could be:

```js
const uniqueMatches = [...new Map(matches.map(match => [match.path, match])).values()];
```

just for you to know in case you prefer to use it like that.

I'll give it another review after your reply.

sofiaromorales:
Thanks for the suggestions @marinaaisa 😄

src/components/Navigator/QuickNavigationModal.vue
sofiaromorales: Resolved https://github.com/apple/swift-docc-render/pull/496/commits/aa5168129b19e9b4f1643cf99059659b4fb42247

tests/unit/components/Navigator/QuickNavigationModal.spec.js
sofiaromorales: Resolved https://github.com/apple/swift-docc-render/pull/496/commits/aa5168129b19e9b4f1643cf99059659b4fb42247

marinaaisa:
@swift-ci test

marinaaisa (approved):
Great job, as always.
Thanks a lot for working on this issue @sofiaromorales  👏

PR #498, opened 2022-12-15
Minimized view should display less content 

hqhhuang:
Bug/issue #, if applicable: 103149476

## Summary
The minimized view should not display the following items:
1. Title and eyebrow
2. Language Switcher
3. Availability
5. Topics section
6. Default Implementations
7. Relationships
8. See Also

## Testing
Steps:
1. Test with this DocC Archive: 
[SlothCreator_Minimized_View.doccarchive.zip](https://github.com/apple/swift-docc-render/files/10240143/SlothCreator_Minimized_View.doccarchive.zip)

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -31,7 +31,7 @@
           :objcPath="objcPath"
           :swiftPath="swiftPath"
         />
-        <Title :eyebrow="roleHeading">
+        <Title :eyebrow="roleHeading" v-if="!enableMinimized">
           <component :is="titleBreakComponent">{{ title }}</component>
           <small
             v-if="isSymbolDeprecated || isSymbolBeta"
@@ -45,7 +45,7 @@
           <DownloadButton class="sample-download" :action="sampleCodeDownload.action" />
         </div>
         <Availability
-          v-if="hasAvailability"
+          v-if="shouldShowAvailability"
           :platforms="platforms" :technologies="technologies"
         />
       </DocumentationHero>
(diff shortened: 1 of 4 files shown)


src/components/DocumentationTopic.vue
```diff
@@ -409,10 +417,14 @@ export default {
     shouldRenderTopicSection: ({
       topicSectionsStyle,
       topicSections,
-    }) => topicSections && topicSectionsStyle !== TopicSectionsStyle.hidden,
+      enableMinimized,
+    }) => topicSections && topicSectionsStyle !== TopicSectionsStyle.hidden && !enableMinimized,
     isOnThisPageNavVisible: ({ topicState }) => (
       topicState.contentWidth > ON_THIS_PAGE_CONTAINER_BREAKPOINT
     ),
+    enableMinimized({ $route }) {
```
mportiz08, line 425: I suggest that we make a `minimal` prop for this instead of creating a computed property that checks the route.

In the actual routing config, we can map a URL path to a component/view that then sets this prop.

src/views/DocumentationTopic.vue
```diff
@@ -150,6 +151,7 @@ export default {
       showQuickNavigationModal: false,
       store: DocumentationTopicStore,
       BreakpointName,
+      enableMinimized: false,
```
mportiz08, line 154: Since this is never actually expected to change during the lifecycle of an instance of this component, I would recommend just making it a prop with the `false` default—just like you already have it in the other component.

hqhhuang:
@swift-ci test

mportiz08 (approved):

Issue #503, opened 2022-12-23
Declaration source indents from incorrect closing parenthesis when return value is a tuple


qasim:
### Description

When viewing a method declaration which has a return value that contains its own parentheses (i.e. a tuple), the last line's indentation is off. For an example of this in Apple Developer Documentation, see [`URLSession` / `download(for:delegate:)`](https://developer.apple.com/documentation/foundation/urlsession/3767354-download):

```swift
    func download(
        for request: URLRequest,
        delegate: URLSessionTaskDelegate? = nil) async throws -> (URL, URLResponse
                                            // ^ newline should be inserted here instead ✅
    )
 // ^ not here ❌
```

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc-render/issues).

### Expected Behavior

```swift
func download(
    for request: URLRequest,
    delegate: URLSessionTaskDelegate? = nil
) async throws -> (URL, URLResponse)
```

### Actual behavior

```swift
func download(
    for request: URLRequest,
    delegate: URLSessionTaskDelegate? = nil) async throws -> (URL, URLResponse
)
```

### Steps To Reproduce

1. Define a public structure/object that has a method which returns a tuple, i.e:
```swift
public struct DeclarationIndentationBug<A, B> {
    public static func foo(
        _ a: A,
        _ b: B
    ) -> (A, B) {
        return (a, b)
    }
}
```

2. Build a `.doccarchive`
3. Render the `.doccarchive` using `swift-docc-render`

### Swift-DocC-Render Version Information

main

qasim:
I can take a go at building `swift-docc-render` and proposing a fix. 🙂

PR #504, opened 2022-12-23
Fix `DeclarationSource` inserts newline at incorrect close parenthesis

qasim:
Bug/issue #, if applicable: #503

## Summary

`DeclarationSource`, when formatting for multi-param symbols, incorrectly identifies the last closing parenthesis of the entire declaration, rather than the closing parenthesis for the first opening parenthesis in the declaration.

To fix, we only set closing parenthesis index when it is previously `null`.

## Testing

1. Define a public structure/object that has a method which returns a tuple, i.e:
```swift
public struct DeclarationIndentationBug<A, B> {
    public static func foo(
        _ a: A,
        _ b: B
    ) -> (A, B) {
        return (a, b)
    }
}
```

2. Build a `.doccarchive`
3. Render the `.doccarchive` using `swift-docc-render`

### Outcome

| Before | After |
| --- | --- |
|Before|After|

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -114,7 +114,11 @@ export default {
             numUnclosedParens -= 1;
             // if this ")" balances out the number of "(" characters that have
             // been seen, this is the one that pairs up with the first one
-            if (openParenTokenIndex !== null && numUnclosedParens === 0) {
+            if (
+              openParenTokenIndex !== null
+              && closeParenTokenIndex == null
+              && numUnclosedParens === 0
+            ) {
               closeParenCharIndex = k;
               closeParenTokenIndex = i;
             }
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -121,6 +121,7 @@ export default {
             ) {
               closeParenCharIndex = k;
               closeParenTokenIndex = i;
+              break;
             }
           }
         }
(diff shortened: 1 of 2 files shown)


dobromir-hristov:
@swift-ci test

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -114,7 +114,11 @@ export default {
...
             // if this ")" balances out the number of "(" characters that have
             // been seen, this is the one that pairs up with the first one
-            if (openParenTokenIndex !== null && numUnclosedParens === 0) {
+            if (
+              openParenTokenIndex !== null
+              && closeParenTokenIndex == null
+              && numUnclosedParens === 0
+            ) {
               closeParenCharIndex = k;
               closeParenTokenIndex = i;
```
qasim, line 123: Can we break this for loop early once we've found the closing parenthesis?
```suggestion
              closeParenTokenIndex = i;
              break;
```

mportiz08: Yeah, I think that makes sense as an additional improvement since this inner loop doesn't need to do anything else once that closing paren has been found—good call!

mportiz08 (approved):
Awesome work @qasim! Thanks for identifying this bug and extra thanks for finding the small fix for it :)

qasim:
Thank you @mportiz08 🙇🏽‍♂️

Just pushed the change for an early loop termination, ready for one more CI test run and we should be good to go. :)

mportiz08:
@swift-ci test

mportiz08:
Thanks for the awesome contribution @qasim!

PR #506, opened 2023-01-05
Page should have a transparent background when `Minimized` is enabled

hqhhuang:
Bug/issue #, if applicable: 103150955

## Summary
Documentation pages should have a transparent background when the prop `enableMinimized` is set to true

## Testing
Steps:
1. Manually verify that the background is set to transparent in both light and dark modes.
2. Verify that no regression is introduced when Minimized mode is not enabled. Make sure to scroll further up when the page is already at the top, and scroll further down when the page is already at the bottom.

### DIFF
--- a/src/styles/base/_typography.scss
+++ b/src/styles/base/_typography.scss
@@ -21,7 +21,9 @@ html {
 
 body {
   @include font-styles(body);
-  background-color: var(--color-text-background);
+  @include inTargetWeb {
+    background-color: var(--color-text-background);
+  }
   color: var(--colors-text, var(--color-text));
   font-style: normal;
   word-wrap: break-word;
--- a/src/views/DocumentationTopic.vue
+++ b/src/views/DocumentationTopic.vue
@@ -469,10 +469,6 @@ export default {
   }
 }
 
-.background-disabled {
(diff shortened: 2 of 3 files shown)


hqhhuang:
@swift-ci test

src/styles/base/_typography.scss
```diff
@@ -21,7 +21,9 @@ html {
 
 body {
   @include font-styles(body);
-  background-color: var(--color-text-background);
+  @include inTargetWeb {
```
mportiz08, line 24: You might consider removing this change and simply setting the `--color-text-background` to `transparent` in the place where that variable is being overwritten for IDE targets.

src/views/DocumentationTopic.vue
```diff
@@ -477,6 +481,10 @@ export default {
   }
 }
 
+.background-disabled {
```
mportiz08, line 484: This change can likely be avoided altogether if we instead simply override the `--color(s)-text-background` variable to be transparent for IDE targets.

mportiz08:
@hqhhuang can we close this now or was there a specific reason for keeping this up as a draft PR?

PR #507, opened 2023-01-05
`Minimized` style updates

hqhhuang:
Bug/issue #, if applicable: 103150108

## Summary
This PR covers some style updates for minimized.

## Testing
Steps:
1. Manually verify that the following items are achieved when the prop `enableMinimized` is set to true:
- Use "Summary" as the heading for abstracts
- Remove extra padding at the bottom of hero section
2. Verify that we now allow browsers to set the default font size when the target is IDE (Here's the [Draft PR](https://github.com/apple/swift-docc-render/pull/500) from earlier)

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -17,6 +17,7 @@
       <DocumentationHero
         :role="role"
         :enhanceBackground="enhanceBackground"
+        :enableMinimized="enableMinimized"
         :shortHero="shortHero"
         :shouldShowLanguageSwitcher="shouldShowLanguageSwitcher"
         :iconOverride="references[pageIcon]"
@@ -40,7 +41,12 @@
             :data-tag-name="tagName"
           />
         </Title>
-        <Abstract v-if="abstract" :content="abstract" />
+        <LinkableHeading v-else class="minimized-summary">Summary</LinkableHeading>
+        <Abstract
+          v-if="abstract"
+          :class="{ 'minimized-abstract': enableMinimized }"
+          :content="abstract"
+        />
         <div v-if="sampleCodeDownload">
           <DownloadButton class="sample-download" :action="sampleCodeDownload.action" />
         </div>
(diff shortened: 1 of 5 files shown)


src/components/DocumentationTopic.vue
```diff
@@ -40,7 +41,12 @@
             :data-tag-name="tagName"
           />
         </Title>
-        <Abstract v-if="abstract" :content="abstract" />
+        <LinkableHeading v-else class="minimized-summary">Summary</LinkableHeading>
```
mportiz08, line 44: To make this a little logic slightly more straightforward to read, you may want to consider moving this before `Title` and swapping the conditions.

Instead of:
```
Title if not minimized
Summary else
```

Something like:
```
Summary if minimized
Title else
```

I would also consider just using the title component for this and changing the text but it sounds like these may be using a different heading level which makes that more complex and probably not worth looking into.

tests/unit/components/DocumentationTopic.spec.js
```diff
@@ -333,6 +338,13 @@ describe('DocumentationTopic', () => {
     expect(wrapper.find(DocumentationHero).find(Title).exists()).toBe(false);
   });
 
+  it('renders a `LinkableHeading`, in minimized mode, if `enableMinimized` prop is `true`', () => {
+    const heading = wrapper.find(LinkableHeading);
+    expect(heading.exists()).toBe(false);
+    wrapper.setProps({ enableMinimized: true });
+    expect(wrapper.find(LinkableHeading).exists()).toBe(true);
```
mportiz08, line 345: You might also want to assert that a `Title` was not rendered since this is supposed to be taking its place in that scenario

hqhhuang: That's tested in the[ test for `Title`](https://github.com/apple/swift-docc-render/blob/67a60eb6a2f6908eba1374d88990e6070f1924bc/tests/unit/components/DocumentationTopic.spec.js#L336) from an earlier PR. I think keeping it there that makes more sense than moving it to the test for the `LinksableHeading`, what do you think?

src/components/DocumentationTopic.vue
hqhhuang: Good point! I put `Summary` second because it's a less frequent scenario. I can see how that's confusing and not straightforward. I fixed it.

> I would also consider just using the title component for this and changing the text but it sounds like these may be using a different heading level which makes that more complex and probably not worth looking into.

Yup, the `Summary` heading is meant to be the same level as all the headers for primary content. I will keep this in mind for the future.

hqhhuang:
@swift-ci test

mportiz08 (approved):

PR #515, opened 2023-01-12
Add DeviceFrames support

dobromir-hristov:
Bug/issue #, if applicable: 102182966

## Summary

Adds support for DeviceFrames for images and videos.

localhost_8082_documentation_simple_simpler

## Dependencies

- https://github.com/apple/swift-docc/pull/462

## Testing

[simpler.json.zip](https://github.com/apple/swift-docc-render/files/10400403/simpler.json.zip) - this file is built with the docc branch above

Steps:
1. To test this you will need an archive that has `@Image` or `@Video` tags with `deviceFrames`.
2. You will also need to add deviceFrame definitions to your `theme-settings.json` file, as by default docc-render does not ship with frames. - see https://forums.swift.org/t/support-for-device-frames/62151 for more info

## Checklist

- [x] Added tests

### DIFF
--- a/bin/check-source
+++ b/bin/check-source
@@ -52,10 +52,10 @@ EOF
         cat > "$tmp" <<"EOF"
 /*!
  * This source file is part of the Swift.org open source project
- *
+ * 
  * Copyright (c) YEARS Apple Inc. and the Swift project authors
  * Licensed under Apache License v2.0 with Runtime Library Exception
- *
+ * 
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 EOF
--- a/bin/check-source
+++ b/bin/check-source
@@ -18,7 +18,7 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 function replace_acceptable_years() {
     # this needs to replace all acceptable forms with 'YEARS'
-    sed -e 's/20[12][78901]-20[12][89012]/YEARS/' -e 's/20[12][89012]/YEARS/'
+    sed -e 's/20[12][78901]-20[12][890123]/YEARS/' -e 's/20[12][890123]/YEARS/'
 }
 
 printf "=> Checking license headers… "
--- a/bin/check-source
+++ b/bin/check-source
@@ -52,10 +52,10 @@ EOF
         cat > "$tmp" <<"EOF"
 /*!
  * This source file is part of the Swift.org open source project
- * 
+ *
  * Copyright (c) YEARS Apple Inc. and the Swift project authors
  * Licensed under Apache License v2.0 with Runtime Library Exception
- * 
+ *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 EOF
--- a/src/components/ContentNode/DeviceFrame.vue
+++ b/src/components/ContentNode/DeviceFrame.vue
@@ -25,7 +25,8 @@
 import DeviceFrames from 'theme/constants/DeviceFrames.js';
 import { getSetting } from 'docc-render/utils/theme-settings';
 
-const round = (v, dec = 4) => +(`${Math.round(`${v}e+${dec}`)}e-${dec}`);
+const isTruthy = val => val && val !== Infinity;
+const round = (v, dec = 4) => (isTruthy(v) ? +(`${Math.round(`${v}e+${dec}`)}e-${dec}`) : null);
 
 export default {
   name: 'DeviceFrame',
@@ -67,9 +68,9 @@ export default {
     }),
   },
   methods: {
-    toPixel: val => (val ? `${val}px` : null),
-    toUrl: val => (val ? `url(${val})` : null),
-    toPct: val => (val ? `${round(val * 100)}%` : null),
+    toPixel: val => (isTruthy(val) ? `${val}px` : null),
+    toUrl: val => (isTruthy(val) ? `url(${val})` : null),
+    toPct: val => (isTruthy(val) ? `${round(val * 100)}%` : null),
   },
 };
 </script>
@@ -84,6 +85,7 @@ export default {
   max-width: 100%;
   margin-left: auto;
   margin-right: auto;
+  overflow: hidden;
   @include space-out-between-siblings($article-stacked-margin-large);
 }
 
--- a/src/components/ContentNode/DeviceFrame.vue
+++ b/src/components/ContentNode/DeviceFrame.vue
@@ -1,0 +1,135 @@
+<!--
+  This source file is part of the Swift.org open source project
+
+  Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+<template>
+  <div
+    class="device-frame"
+    :data-device="device"
+    :class="classes"
+    :style="styles"
+  >
+    <div class="device-screen" :class="{ 'with-device': currentDeviceAttrs }">
+      <slot />
+    </div>
+    <div class="device" />
+  </div>
+</template>
+
+<script>
+import DeviceFrames from 'theme/constants/DeviceFrames.js';
+import { getSetting } from 'docc-render/utils/theme-settings';
+
+const round = (v, dec = 4) => +(`${Math.round(`${v}e+${dec}`)}e-${dec}`);
+
+export default {
+  name: 'DeviceFrame',
+  props: {
+    device: {
+      type: String,
+      required: true,
+    },
+  },
+  provide: {
+    insideDeviceFrame: true,
+  },
+  computed: {
+    currentDeviceAttrs: ({ device }) => getSetting(['theme', 'device-frames', device], DeviceFrames[device]),
+    styles: ({
+      toPixel, toUrl, toPct, currentDeviceAttrs: attrs = {},
+    }) => {
+      const {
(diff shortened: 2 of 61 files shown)


dobromir-hristov:
One thing I feel that needs discussion is the different way a DeviceFrame is rendered for an Image vs Videos.

For the `Image`, I wrap it around a `DeviceFrame` component, right inside the `ContentNode`. 

For the `Video`, I instead added a special conditional component inside the `VideoAsset` instead. This is because replayable videos cannot be wrapped in a frame, because then the replay button overlaps the frame, along with other resizing complications.

I am open to other suggestions, but this is what I could think of after much trial and error.

dobromir-hristov:
@swift-ci test

bin/check-source
```diff
@@ -52,10 +52,10 @@ EOF
         cat > "$tmp" <<"EOF"
 /*!
  * This source file is part of the Swift.org open source project
- *
```
mportiz08, line 55: These whitespace changes were likely mistakenly added by your editor or something.

src/components/ContentNode/DeviceFrame.vue
```diff
@@ -0,0 +1,135 @@
...
+    toUrl: val => (val ? `url(${val})` : null),
+    toPct: val => (val ? `${round(val * 100)}%` : null),
+  },
+};
+</script>
+
+<style scoped lang='scss'>
+@import 'docc-render/styles/_core.scss';
+
+.device-frame {
```
mportiz08, line 80: We may want to add `overflow: hidden;` to this element. When testing, I found it was easy to accidentally break the layout with the framed images when making a mistake with the dimensions.

src/components/ContentNode/DeviceFrame.vue
```diff
@@ -0,0 +1,135 @@
...
+    styles: ({
+      toPixel, toUrl, toPct, currentDeviceAttrs: attrs = {},
+    }) => {
+      const {
+        screenTop, screenLeft, screenWidth, frameWidth,
+        lightUrl, darkUrl, screenHeight, frameHeight,
+      } = attrs;
+
+      return {
+        '--screen-top': toPct(screenTop / frameHeight),
```
mportiz08, line 52: I'm not sure how you would want to handle it yet, but we should likely add some sort of guard for when someone mistakenly uses dimensions that result in dividing by zero here—otherwise we end up with funky NaN percentage strings.

src/components/ContentNode/DeviceFrame.vue
```diff
@@ -0,0 +1,135 @@
...
+};
+</script>
+
+<style scoped lang='scss'>
+@import 'docc-render/styles/_core.scss';
+
+.device-frame {
+  position: relative;
+  width: var(--frame-width);
+  aspect-ratio: var(--frame-aspect);
```
mportiz08, line 83: It looks like the dimensions for the screen and frame always have to match up exactly. Is there anything we could realistically do to better handle situations where the given image is actually shorter (as one example) and try to center it in the frame or something? Maybe that could just be something to look at in the future. Just thinking out loud.

src/components/ContentNode/DeviceFrame.vue
```diff
@@ -0,0 +1,135 @@
...
+
+        '--frame-width': toPixel(frameWidth),
+        '--frame-aspect': round(frameWidth / frameHeight) || null,
+
+        '--device-light-url': toUrl(lightUrl),
+        '--device-dark-url': toUrl(darkUrl),
+      };
+    },
+    classes: ({ currentDeviceAttrs }) => ({
+      'no-device': !currentDeviceAttrs,
```
mportiz08, line 66: This part is a little odd—I would expect that this component wouldn't ever even need to be utilized if we don't have any attributes for the requested device.

However, I'm guessing you did that to avoid complicating the already complex logic in `ContentNode` which would likely have to then check for the device attributes and stuff before rendering this component. Is that right? Just curious.

mportiz08 (commented):
It takes a lot of work and configuration to get this working, but it looks pretty cool once it's going :)

I've only had time to test this with images so far and partially review the code for that—I'll try to test/review the frames with video tomorrow.

src/components/ContentNode/DeviceFrame.vue
dobromir-hristov: If the image/video is the wrong aspect ratio it gets top-aligned last time I checked via:

```
 object-fit: contain;
 object-position: top;
```

@SamLanier I think preferred this way, over trying to center and crop on behalf of the user.

dobromir-hristov:
@swift-ci test

src/constants/DeviceFrames.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2021 Apple Inc. and the Swift project authors
```
marinaaisa, line 4:
```suggestion
 * Copyright (c) 2023 Apple Inc. and the Swift project authors
```

marinaaisa (changes requested):
Very nice job, @dobromir-hristov .
I left minor comments.

Also I saw that when you declare an image path but it doesn't exist in the doccarchive image folder, this error in console comes up:

Screenshot 2023-01-16 at 15 32 50

We should consider to throw a better explained error at `getIntrinsicDimensions`

src/components/VideoAsset.vue
```diff
@@ -64,12 +75,17 @@ export default {
...
+      type: String,
+      required: false,
+    },
   },
   data: () => ({
     appState: AppStore.state,
     optimalWidth: null,
   }),
   computed: {
+    DeviceFrameComponent: () => DeviceFrame,
```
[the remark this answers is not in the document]
dobromir-hristov: I am passing the component itself as a prop :)

bin/check-source
```diff
@@ -52,10 +52,10 @@ EOF
         cat > "$tmp" <<"EOF"
 /*!
  * This source file is part of the Swift.org open source project
- * 
+ *
  * Copyright (c) YEARS Apple Inc. and the Swift project authors
  * Licensed under Apache License v2.0 with Runtime Library Exception
- * 
+ *
```
dobromir-hristov, line 58:
```suggestion
 * 
```

marinaaisa:
> Also I saw that when you declare an image path but it doesn't exist in the doccarchive image folder, this error in console comes up.

I'll fill a different radar for it, no a blocker for this PR.

src/components/ContentNode/DeviceFrame.vue
mportiz08: Ok. Sounds reasonable, and I suppose we could always change it to center if we decide that's better for that scenario.

src/components/ContentNode/DeviceFrame.vue
mportiz08: Agreed :/

src/components/VideoAsset.vue
```diff
@@ -1,33 +1,41 @@
...
-    :controls="showsControls"
-    :autoplay="autoplays"
-    :poster="normalisedPosterPath"
-    :muted="muted"
-    :width="optimalWidth"
-    playsinline
-    @playing="$emit('playing')"
-    @pause="$emit('pause')"
-    @ended="$emit('ended')"
+  <ConditionalWrapper
```
mportiz08, line 12: Could you briefly explain the need for `<ConditionalWrapper>`? I think you added it because you want to conditionally wrap the video in another element only if it has a device frame, right?

And you didn't just use normal `v-if` templating because it would make the inner `<video ...>` part have to be copied twice?

And `<component :is="...">` would be limited to using a plain div or something in the case where there is no device frame?

Is that all right? Just making sure I understand the need for it.

src/constants/DeviceFrames.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2023 Apple Inc. and the Swift project authors
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
```
mportiz08, line 10: Could you maybe add a comment here to explain that this is empty because swift-docc-render does not provide any actual device frames on its own by default?

mportiz08 (commented):
Looks good, mostly just some minor comments.

I have some concerns about the overall feature and its usability, but I've already expressed those offline, and I understand that this is just an experimental feature for now.

Everything seems to be working well with images/videos in frames that I've tested so far. I haven't tested using this in a tutorial which might be interesting since we already have the runtime preview component—maybe someone can help test that too.

src/components/ContentNode/DeviceFrame.vue
dobromir-hristov: Its a simple property, ofc :)

src/components/VideoAsset.vue
dobromir-hristov: Yes exactly all of those. I wanted to make it as transparent as possible. 

Initially I did it by moving everything to render functions in the VideoAsset. By doing that, you can easily wrap things, but I then reverted back to the template syntax since it read easier, and we may use this handy component somewhere else in the future?

src/constants/DeviceFrames.js
dobromir-hristov: yep yep.

dobromir-hristov:
```suggestion

// Swift-DocC-Render does not provide device frames out of the box.
```

dobromir-hristov:
@swift-ci test

mportiz08 (approved):

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #524, opened 2023-01-31
Making minimized style look more similar to documentation page

hqhhuang:
Bug/issue #, if applicable: 104357619 and 88423841

## Summary
This PR makes minimized style look more similar to documentation pages.

## Testing
Steps:
1. Test & make sure there is no regression for the places we use the stacked-margins variables.
2. Manually verify that the following items are achieved when the prop enableMinimized is set to true:
- `Title` displayed, instead of `Summary` heading 
-  Role heading should not be displayed

## Checklist

- [x] Added tests


hqhhuang:
@swift-ci test

hqhhuang:
@swift-ci test

dobromir-hristov (commented):
Code looks fine, havent tested it yet. Only comment I could make is that we could have used this as an opportunity to clean up the naming scheme:

`--spacing-X` instead of stacked bla bla bla, which made sense at one point, but no really now I think.

src/components/ContentNode/BlockVideo.vue
```diff
@@ -42,7 +42,7 @@ export default {
   margin-left: auto;
   margin-right: auto;
 
-  @include space-out-between-siblings($stacked-margin-xlarge);
+  @include space-out-between-siblings(var(--stacked-margin-xlarge));
```
mportiz08, line 45: I haven't tested this yet, but I'm not 100% sure this will work as written, because it's very tricky to combine Sass functions and CSS properties.

Since Sass works as a preprocessor, it substitutes the actual value for the variables at build time whereas the CSS property is evaluated at runtime—I don't know of a way to combine them for this particular usage, but we would likely need to find a workaround for this change.

src/components/ContentNode/TabNavigator.vue
```diff
@@ -74,7 +74,7 @@ export default {
 @import 'docc-render/styles/_core.scss';
 
 .TabNavigator {
-  @include space-out-between-siblings($stacked-margin-xlarge);
+  @include space-out-between-siblings(var(--stacked-margin-xlarge));
```
mportiz08, line 77: (Same comment about mixing Sass mixins/functions with CSS vars here)

src/components/ContentNode/Table.vue
```diff
@@ -34,7 +34,7 @@ export default {
 .table-wrapper {
   overflow: auto;
   -webkit-overflow-scrolling: touch;
-  @include space-out-between-siblings($stacked-margin-xlarge);
+  @include space-out-between-siblings(var(--stacked-margin-xlarge));
```
mportiz08, line 37: (Same comment about mixing Sass mixins/functions with CSS vars here)

src/components/DocumentationTopic.vue
```diff
@@ -32,8 +32,7 @@
           :objcPath="objcPath"
           :swiftPath="swiftPath"
         />
-        <LinkableHeading v-if="enableMinimized" class="minimized-summary">Summary</LinkableHeading>
-        <Title v-else :eyebrow="roleHeading">
+        <Title :eyebrow="roleHeading" :enableMinimized="enableMinimized">
```
mportiz08, line 35: I think the eyebrow is already an optional prop to the title component so you can just omit that in the minimized case and apply a special class here for the minimized title similar to how you're already doing it for the description/abstract in other places within this file—that should avoid the need to pass this minimized flag to deeper components and have it simply be a feature of this one.

src/components/DocumentationTopic/PrimaryContent.vue
```diff
@@ -65,6 +66,11 @@ export default {
         Object.prototype.hasOwnProperty.call(SectionKind, kind)
       )),
     },
+    enableMinimized: {
```
mportiz08, line 69: I think we should try to avoid passing this flag to lots of subcomponents if possible and try to keep it to the one main one in the documentation topic. It would be nice if most of the specialized classes/styles associated with that flag could be easily colocated in the same component and it would be easier to maintain.

src/views/DocumentationTopic.vue
```diff
@@ -89,6 +89,7 @@
           :languagePaths="languagePaths"
           :enableOnThisPageNav="enableOnThisPageNav"
           :enableMinimized="enableMinimized"
+          :disableHeroBackground="enableMinimized"
```
mportiz08, line 92: To avoid having to make any changes in this file, I think you could instead update the `src/components/DocumentationTopic.vue` component to consider the already passed `enableMinimized` flag in the `enhanceBackground` computed property.

Also, it avoids an interesting scenario where the background could be enabled even when the minimal flag is passed in another context.

mportiz08 (commented):
I want to do some more manual testing of these changes to make sure there aren't any regressions, but I did a pass on reviewing all the source changes.

src/styles/core/_margin.scss
```diff
@@ -0,0 +1,33 @@
...
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2023 Apple Inc. and the Swift project authors
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+@mixin margin-vars() {
```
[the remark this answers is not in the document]
hqhhuang: Do you think add spacing variables to `/base/typography.scss` file would be confusing, since right now it's only used for colors?

src/styles/core/_margin.scss
```diff
@@ -0,0 +1,33 @@
...
+  // Default margin for stacked elements.
+  --stacked-margin-small: 0.4em;
+
+  // Default margin-top for subsequent elements.
+  --stacked-margin-large: 0.8em;
```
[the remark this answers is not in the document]
hqhhuang: This is a good idea! But I think before exposing them, there are work we need to do to clean up the variables we use for margin in the entire app. Some places, we use variables, other places we hard code the spacing. We should make it very clear to people what/where would change if they decide to customize these variables.

src/views/DocumentationTopic.vue
hqhhuang: The reason why I add this is because we want the background to always be disabled when minimal flag is on. And Dobri added the `diableHeroBackground` prop (that's not currently being used) to allow convenient disabling. I thought just using that was cleaner than making the `enhanceBackground` property even more complicated.

src/styles/core/_margin.scss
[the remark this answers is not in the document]
mportiz08: It's used for anything related to typography like the font/layout and spacing. It's also fine to create a new file specifically for spacing styles if that's preferred. Either way, we should make sure to use something in base as opposed to core if the result is outputted styles (vs core functionality).

src/views/DocumentationTopic.vue
mportiz08: That makes sense to me but since the `src/components/DocumentationTopic.vue` component can take both boolean flags and it doesn't use both in the logic, it's possible to have a situation where your first sentence is true when you use this view (I can show you a quick in-person demo of where I noticed this on something else I'm working on)

src/components/ContentNode/TabNavigator.vue
dobromir-hristov: I think it should be fine, as it should get passed as a string kind of :D

src/components/ContentNode/BlockVideo.vue
hqhhuang: It seems to be working according to my testing. Can you test this as well?

mportiz08: You're right. That's surprising to me, but I'm glad it's working as-is!

Maybe the issue I was worrying about is only a limitation with Sass functions as opposed to mixins where the variables are directly substituted in rules—not sure.

Either way, very glad this works!

src/components/DocumentationTopic.vue
```diff
@@ -32,8 +32,10 @@
           :objcPath="objcPath"
           :swiftPath="swiftPath"
         />
-        <LinkableHeading v-if="enableMinimized" class="minimized-summary">Summary</LinkableHeading>
-        <Title v-else :eyebrow="roleHeading">
+        <Title
+          :eyebrow="enableMinimized ? '' : roleHeading"
```
mportiz08, line 36: This works since `!!''` evaluates to `false`, but I think using `null` or `undefined` instead of the empty string would make it a little more clear that we're purposefully omitting the eyebrow value in this case.

src/components/DocumentationTopic.vue
```diff
@@ -520,15 +539,37 @@ export default {
...
   .no-primary-content {
+    // remove border-top for first section of the page
     --content-table-title-border-width: 0px;
   }
+
+  .minimized-content {
+    & > * {
+      margin-bottom: 1.5em;
+      margin-top: 1.5em;
+      --spacing-stacked-margin-large: 0.667em;
```
mportiz08, line 560: I think you may want to move these property assignments so that they aren't applied to lots of different elements using the `*` selector as you have it here. We only need these to apply to some root parent element where we want everything within it to inherit these properties.

I think this would ideally either be directly within the `.minimized-content` class or possibly adding a minimal-only class to the entire root element of this component and defining them there.

src/components/DocumentationTopic/PrimaryContent.vue
```diff
@@ -152,7 +152,9 @@ export default {
     content: '';
     display: block;
   }
+}
 
+.primary-content:not(.minimized-content) {
```
mportiz08, line 157: There should be a way to eliminate the need for the addition of this `:not()` selector based on how you structure the overriding rule in `DocumentationTopic.vue`. Maybe you tried this at first and it didn't work? If not, maybe you weren't overriding it in the same way it's defined here?

mportiz08 (approved):

hqhhuang:
@swift-ci test

PR #529, opened 2023-02-08
add docs about using nvm to readme and contributing.md

QuietMisdreavus:
Bug/issue #, if applicable: None

## Summary

This PR adds a blurb to the README and CONTRIBUTING.md files about using `nvm` to handle installing Node and npm. Since Swift-DocC-Render has an `.nvmrc` file, this provides a quick way to ensure that you have the right versions of Node/npm before working on the project itself.

## Dependencies

None

## Checklist

- [ n/a ] Added tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,9 @@ more details.
 ### Run Steps
 
 > Note: requires [Node.js](https://nodejs.org/en/download/) v14
-> and [npm](https://www.npmjs.com/package/npm) v6.14
+> and [npm](https://www.npmjs.com/package/npm) v6.14. An easy way to set these up is to install
+> [nvm](https://github.com/nvm-sh/nvm) and run `nvm install` from within the Swift-DocC-Render
+> repository. To use these versions as the default, add `--default` to the installation command.
 
 1. Checkout this repository using:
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -51,7 +51,9 @@ more details.
 ### Run Steps
 
 > Note: requires [Node.js](https://nodejs.org/en/download/) v14
-> and [npm](https://www.npmjs.com/package/npm) v6.14
+> and [npm](https://www.npmjs.com/package/npm) v6.14. The quickest way to set these up is to install
```
mportiz08, line 54: Maybe there's a slightly different way of wording this that doesn't imply that nvm is the quickest and/or best way of managing node versions while still pointing out that it may be convenient for contributors?

I don't use it myself but I think this small change to the docs a helpful addition to facilitate a more efficient first experience with the project :)

mportiz08 (approved):

CONTRIBUTING.md
QuietMisdreavus: That's a good point. How about something like "an easy way to set these up"? That doesn't give a total value judgment, but it still provides a recommendation for people getting up to speed.

mportiz08: Sounds good to me.

QuietMisdreavus: Pushed the edit!

QuietMisdreavus:
@swift-ci Please test

PR #533, opened 2023-02-14
Fix issue with images not rendered inside a tabnav, in Safari

dobromir-hristov:
Bug/issue #, if applicable: 105363672

## Summary

This fixes a bug where images are not rendered inside tabnavs, for Safari only.

This only happens if the image is a `display: block`, and has some element before it. Really strange, but removing the `loading: lazy` fixes it.

## Dependencies

NA

## Testing

Compile and try with this:

```md
@TabNavigator{
  @Tab("Image With Title"){
    **Strong Title**

    @Image(source: example.png)
  }
  @Tab("Image with Caption"){
    @Image(source: example.png){ Some text as a placeholder }
  }
}
```

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode/TabNavigator.vue
+++ b/src/components/ContentNode/TabNavigator.vue
@@ -33,6 +33,7 @@
 <script>
 import Tabnav from 'docc-render/components/Tabnav.vue';
 import TabnavItem from 'docc-render/components/TabnavItem.vue';
+import ImageLoadingStrategy from '@/constants/ImageLoadingStrategy';
 
 /**
  * Tab navigation component, that renders `ContentNode`,
@@ -45,6 +46,9 @@ export default {
   components: {
     TabnavItem,
     Tabnav,
+  },
+  provide: {
+    imageLoadingStrategy: ImageLoadingStrategy.eager,
   },
   props: {
     vertical: {
--- a/src/components/ContentNode/TabNavigator.vue
+++ b/src/components/ContentNode/TabNavigator.vue
@@ -33,7 +33,7 @@
 <script>
 import Tabnav from 'docc-render/components/Tabnav.vue';
 import TabnavItem from 'docc-render/components/TabnavItem.vue';
-import ImageLoadingStrategy from '@/constants/ImageLoadingStrategy';
+import ImageLoadingStrategy from 'docc-render/constants/ImageLoadingStrategy';
 
 /**
  * Tab navigation component, that renders `ContentNode`,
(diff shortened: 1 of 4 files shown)


mportiz08:
I think we could potentially take these changes as a workaround, but I'm wondering if there's a simpler problem happening here.

It looks like the `overflow: hidden` rule from the `.tabs-content-container` element is the main thing preventing the image from being loaded. If I uncomment that one style, the image loads as expected—since it is technically overflowing, it seems like Safari is the only browser considering the image as not being visible and thus not loading it since we're using the lazy loading strategy.

Can you remind me if that rule is strictly necessary and what scenario it prevents? I'm sure there is one, just curious if it might be better to consider a solution at the CSS level for this or not.

dobromir-hristov:
When switching between tabs, there is a slight transition. 

When sections have different heights, for example you go from a long section to a short section, the `overflow` rule ensures the long section does not spill out into the next elements, below the tabnav. 

This would not be an issue if there is no animation at all, but since for a brief moment both sections are visible at the same time, one is absolutely positioned while the other takes the space it needs.

src/components/ContentNode/TabNavigator.vue
```diff
@@ -33,6 +33,7 @@
 <script>
 import Tabnav from 'docc-render/components/Tabnav.vue';
 import TabnavItem from 'docc-render/components/TabnavItem.vue';
+import ImageLoadingStrategy from '@/constants/ImageLoadingStrategy';
```
mportiz08, line 36:
```suggestion
import ImageLoadingStrategy from 'docc-render/constants/ImageLoadingStrategy';
```

tests/unit/components/ContentNode/TabNavigator.spec.js
```diff
@@ -6,13 +6,14 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
-*/
+ */
```
mportiz08, line 9:
```suggestion
*/
```

mportiz08 (commented):
Seems to be working well.

tests/unit/components/ImageAsset.spec.js
```diff
@@ -6,12 +6,13 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
-*/
+ */
```
dobromir-hristov, line 9:
```suggestion
*/
```

dobromir-hristov:
@swift-ci test

mportiz08 (approved):

mportiz08:
@swift-ci test

PR #536, opened 2023-02-24
Add view more link to Minimized mode

hqhhuang:
Bug/issue #, if applicable:  102876843

## Summary
Added a "View more" link at the bottom of Minimized Mode

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -85,6 +85,12 @@
               :source="remoteSource"
               :sections="primaryContentSections"
             />
+            <SeeMore
+              v-if="enableMinimized"
+              :url="seeMoreLink"
+            >
+              {{ seeMoreLinkTitle }}
+            </SeeMore>
           </div>
           <Topics
             v-if="shouldRenderTopicSection"
@@ -144,6 +150,7 @@ import Relationships from './DocumentationTopic/Relationships.vue';
 import RequirementMetadata from './DocumentationTopic/Description/RequirementMetadata.vue';
 import Availability from './DocumentationTopic/Summary/Availability.vue';
 import SeeAlso from './DocumentationTopic/SeeAlso.vue';
+import SeeMore from './DocumentationTopic/SeeMore.vue';
 import Title from './DocumentationTopic/Title.vue';
 import Topics from './DocumentationTopic/Topics.vue';
 import OnThisPageStickyContainer from './DocumentationTopic/OnThisPageStickyContainer.vue';
@@ -186,6 +193,7 @@ export default {
     RequirementMetadata,
     Availability,
     SeeAlso,
+    SeeMore,
     Title,
     Topics,
     WordBreak,
@@ -258,6 +266,11 @@ export default {
     title: {
       type: String,
       required: true,
+    },
+    seeMoreLinkTitle: {
+      type: String,
+      required: false,
+      default: () => 'See More',
     },
     topicSections: {
       type: Array,
@@ -425,6 +438,14 @@ export default {
       || (deprecationSummary && deprecationSummary.length)
       || (downloadNotAvailableSummary && downloadNotAvailableSummary.length)
       || (primaryContentSections && primaryContentSections.length)
+    ),
+    seeMoreLink: ({
+      references,
+      identifier,
+      objcPath,
+      swiftPath,
+    }) => (
+      references[identifier] ? references[identifier].url : (objcPath || swiftPath)
     ),
     tagName: ({ isSymbolDeprecated }) => (isSymbolDeprecated ? 'Deprecated' : 'Beta'),
     /**
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -349,6 +349,22 @@ export default {
     };
   },
   computed: {
+    normalizedSwiftPath: ({
+      normalizePath,
+      swiftPath,
+    }) => (
+      swiftPath ? buildUrl(normalizePath(swiftPath), {
+        language: undefined,
+      }) : null
+    ),
+    normalizedObjcPath: ({
+      normalizePath,
+      objcPath,
+    }) => (
+      objcPath ? buildUrl(normalizePath(objcPath), {
+        language: Language.objectiveC.key.url,
+      }) : null
+    ),
     defaultImplementationsCount() {
       return (this.defaultImplementationsSections || []).reduce(
         (count, section) => count + section.identifiers.length,
@@ -432,12 +448,11 @@ export default {
     ),
     viewMoreLink: ({
       interfaceLanguage,
-      objcPath,
-      swiftPath,
-      normalizePath,
+      normalizedObjcPath,
+      normalizedSwiftPath,
     }) => (
-      buildUrl(normalizePath(interfaceLanguage === Language.objectiveC.key.api
-        ? objcPath : swiftPath), { language: interfaceLanguage })
+      interfaceLanguage === Language.objectiveC.key.api
+        ? normalizedObjcPath : normalizedSwiftPath
     ),
     tagName: ({ isSymbolDeprecated }) => (isSymbolDeprecated ? 'Deprecated' : 'Beta'),
     /**
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -433,8 +433,10 @@ export default {
       interfaceLanguage,
       objcPath,
       swiftPath,
+      normalizePath,
     }) => (
-      interfaceLanguage === Language.objectiveC.key.api ? objcPath : swiftPath
+      interfaceLanguage === Language.objectiveC.key.api
+        ? normalizePath(objcPath) : normalizePath(swiftPath)
     ),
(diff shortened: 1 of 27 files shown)


src/components/DocumentationTopic.vue
```diff
@@ -426,6 +439,14 @@ export default {
       || (downloadNotAvailableSummary && downloadNotAvailableSummary.length)
       || (primaryContentSections && primaryContentSections.length)
     ),
+    viewMoreLink: ({
+      references,
+      identifier,
+      objcPath,
+      swiftPath,
+    }) => (
+      references[identifier] ? references[identifier].url : (objcPath || swiftPath)
```
mportiz08, line 448: If this is meant to be a link to the page itself, I'm not sure it makes sense to look in `references`, because there would only be one there if it referenced itself.

I think your default case makes sense as the only way to reliably get the link to the current page (`(objcPath || swiftPath)`) but it seems like it would always favor the ObjC variant if one exists, regardless of what the current preference is—we might want to tweak that.

src/components/DocumentationTopic.vue
```diff
@@ -259,6 +267,11 @@ export default {
       type: String,
       required: true,
     },
+    viewMoreLinkTitle: {
```
mportiz08, line 270: Since it's a pretty simple component, I'm wondering if a better alternative to making the title configurable might be to just use a themed import for the `ViewMore` component—that way, the whole component could be easily overwritten (and it's a pretty simple component with mostly just this text). That would also avoid the mess of having to pass this prop down from the root component.

What do you think?

src/components/DocumentationTopic.vue
dobromir-hristov: Right, we should try and check the `this.interfaceLanguage` and fallback to whatever is available. 

Is there a chance that one is not available, when its language is picked? Probably not... 🤔 

```suggestion
      interfaceLanguage === Language.objectiveC.key.api ? objcPath : swiftPath
```

src/components/DocumentationTopic.vue
```diff
@@ -426,6 +429,13 @@ export default {
       || (downloadNotAvailableSummary && downloadNotAvailableSummary.length)
       || (primaryContentSections && primaryContentSections.length)
     ),
+    viewMoreLink: ({
+      interfaceLanguage,
+      objcPath,
+      swiftPath,
+    }) => (
+      interfaceLanguage === Language.objectiveC.key.api ? objcPath : swiftPath
```
mportiz08, line 437: I think this is the right track, but it will likely need some slight tweaks to work exactly as you expect. For example, this may cause issues if the path doesn't start with a slash—I think the existing language toggle [component](https://github.com/apple/swift-docc-render/blob/main/src/components/DocumentationTopic/DocumentationNav/LanguageToggle.vue) already handles this, but this might not work exactly as it is now.

mportiz08 (changes requested):

hqhhuang:
@swift-ci test

src/components/DocumentationTopic/ViewMore.vue
```diff
@@ -0,0 +1,42 @@
...
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
+@import 'docc-render/styles/_core.scss';
+
+.base-link {
+  float: right;
```
dobromir-hristov, line 38: Wow, 🚀  time travel. Any way we can use something ... less ancient? 🤣

dobromir-hristov (commented):
Looking good, left a comment.

src/components/DocumentationTopic/ViewMore.vue
mportiz08: Me trying to remember to use flex box for most things these days:
me

hqhhuang: Pulled it back to modern time :P

src/components/DocumentationTopic.vue
```diff
@@ -426,6 +429,15 @@ export default {
...
       || (primaryContentSections && primaryContentSections.length)
     ),
+    viewMoreLink: ({
+      interfaceLanguage,
+      objcPath,
+      swiftPath,
+      normalizePath,
+    }) => (
+      interfaceLanguage === Language.objectiveC.key.api
+        ? normalizePath(objcPath) : normalizePath(swiftPath)
```
mportiz08, line 439: You could simplify this to `normalizePath(ternaryExpression)` to only call that function once since it is used in both conditions.

src/components/DocumentationTopic.vue
```diff
@@ -426,6 +429,15 @@ export default {
       || (downloadNotAvailableSummary && downloadNotAvailableSummary.length)
       || (primaryContentSections && primaryContentSections.length)
     ),
+    viewMoreLink: ({
+      interfaceLanguage,
+      objcPath,
+      swiftPath,
+      normalizePath,
+    }) => (
+      interfaceLanguage === Language.objectiveC.key.api
```
mportiz08, line 438: I think we also want to include the objc query parameter in the URL for a link to an Objective-C destination.

It will technically work without it but only because of logic we have for preserving the language choice which would add extra computation that isn't needed when we already know we want to go to the objc variant.

src/components/DocumentationTopic/ViewMore.vue
```diff
@@ -0,0 +1,43 @@
...
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <div class="view-more-link">
+    <router-link
+      :to="url"
+      class="base-link"
+    >
+      View more
```
dobromir-hristov, line 17: I would make this a slot or prop, so its easy to override if needed.

dobromir-hristov (commented):
Left a small comment, and i also agree with Marcus review

src/components/DocumentationTopic/ViewMore.vue
hqhhuang: I used a slot before, but Marcus suggested to use a themed import for the `ViewMore` component. I think this method does make more sense, since we avoid passing the text as a prop to `DocumentationTopic`. What do you think?

src/components/DocumentationTopic.vue
```diff
@@ -426,6 +430,15 @@ export default {
...
       || (primaryContentSections && primaryContentSections.length)
     ),
+    viewMoreLink: ({
+      interfaceLanguage,
+      objcPath,
+      swiftPath,
+      normalizePath,
+    }) => (
+      buildUrl(normalizePath(interfaceLanguage === Language.objectiveC.key.api
+        ? objcPath : swiftPath), { language: interfaceLanguage })
```
hqhhuang, line 440: Are we ok with including the language query parameter for swift? The code is a little cleaner this way since we don't have to conditionally set it to null just for swift.

mportiz08: I think we should only include the param for ObjC since that's the only special with the default being Swift for pages available in both. Essentially, this link should work very similarly to the language toggle (except that it's for the current page, not the other language variant)

src/components/DocumentationTopic/ViewMore.vue
mportiz08: Both would probably be the ideal—that way another app could theme this button and easily change _only_ the text.

tests/unit/components/DocumentationTopic.spec.js
```diff
@@ -434,6 +435,16 @@ describe('DocumentationTopic', () => {
     expect(wrapper.contains(PrimaryContent)).toBe(false);
   });
 
+  it('renders `ViewMore` if `enableMinimized`', () => {
```
dobromir-hristov, line 438: I think we should assert if the correct link is passed through?

tests/unit/components/DocumentationTopic/ViewMore.spec.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import { RouterLinkStub, shallowMount } from '@vue/test-utils';
+import ViewMore from 'docc-render/components/DocumentationTopic/ViewMore.vue';
+
+describe('ViewMore', () => {
+  it('renders a <router-link>', () => {
```
dobromir-hristov, line 15: Can you assert the slot?

src/components/DocumentationTopic.vue
```diff
@@ -345,6 +349,22 @@ export default {
...
   },
   computed: {
+    normalizedSwiftPath: ({
+      normalizePath,
+      swiftPath,
+    }) => (
+      swiftPath ? buildUrl(normalizePath(swiftPath), {
+        language: undefined,
+      }) : null
+    ),
```
dobromir-hristov, line 359: Wonder if this is enough 🤔 

```suggestion
    normalizedSwiftPath: ({
      normalizePath,
      swiftPath,
    }) => (
      swiftPath ? normalizePath(swiftPath) : null
    ),
```

dobromir-hristov (approved):
lgtm

mportiz08:
@swift-ci test

PR #538, opened 2023-02-24
Update items size when clicking around navigator

dobromir-hristov:
Bug/issue #, if applicable: 105885210

## Summary

Ensures items get their size properly updated, when clicking around the navigator. This prevents items that grew larger, to stay large if not needed anymore.

## Dependencies

NA

## Testing

Ensure items stay the correct size, like when being bolded.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -47,7 +47,10 @@
           @keydown.up.exact.capture.prevent="focusPrev"
           @keydown.down.exact.capture.prevent="focusNext"
         >
-          <DynamicScrollerItem v-bind="{ active, item, dataIndex: index }">
+          <DynamicScrollerItem
+            v-bind="{ active, item, dataIndex: index }"
+            :ref="`dynamicScroller_${item.uid}`"
+          >
             <NavigatorCardItem
               :item="item"
               :isRendered="active"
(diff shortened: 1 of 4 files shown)


src/components/Navigator/NavigatorCard.vue
```diff
@@ -480,6 +483,14 @@ export default {
       // if we remove APIChanges, remove all related tags as well
       this.selectedTags = this.selectedTags.filter(t => !ChangeNames[t]);
     },
+    async activeUID(newUid, oldUID) {
+      // update the dynamicScroller item's size, when we change the UID
```
mportiz08, line 487: Can you also mention the reason we need to do this, which is because the size may change due to styling we apply to the active item?

src/components/Navigator/NavigatorCard.vue
```diff
@@ -480,6 +483,14 @@ export default {
...
+      await this.$nextTick();
+      const item = this.$refs[`dynamicScroller_${oldUID}`];
+      if (item && item.updateSize) {
+        item.updateSize();
```
mportiz08, line 491: Is `updateSize` a method from the `DynamicScrollerItem` component?

dobromir-hristov:
@swift-ci test

mportiz08 (approved):

PR #542, opened 2023-03-01
Render preview of documentation in quick navigation

mportiz08:
Bug/issue #, if applicable: 104383742

## Summary

This updates the layout of the "Quick Navigation" feature introduced by @sofiaromorales to also show a compact preview of the documentation page for the selected result.

Example:
Screenshot 2023-03-01 at 1 08 08 PM

This update currently requires a feature flag to be enabled in `theme-settings.json`, but I would like to remove that if the feedback for this new UI is generally positive.

Setting: `features.docs.quickNavigationPreview.enable` (set to `true` to opt-in)

## Testing

Generally, you should be able to just add the feature flag from above to an existing DocC project and try it out using a build of the renderer built from this branch.

Example:
1. Checkout this branch and build it with `npm run build`
2. `git clone git@github.com:apple/swift-argument-parser.git`
3. To setup swift-argument-parser repo with swift-docc-plugin and needed feature flag: `cd swift-argument-parser; curl https://gist.githubusercontent.com/mportiz08/03193ba981c168697c89c2de1fccdd6f/raw/b6b5886be9d4cc18e8764bd2d90f655a6a22c1c9/add-plugin-and-settings.diff|git apply`
4. Run the preview with `env DOCC_HTML_DIR=/path/to/swift-docc-render/dist swift package --disable-sandbox preview-documentation --target ArgumentParser`

That's just one example of course—feel free to try out previewing your own example content however you'd like.

Things to test:
* preview works for doc pages
* preview works for tutorial pages
* loading UI should show up if it takes a full second or more to load data
* "Preview unavailable" message should show if an error happens when attempting to load data for a page
* preview should generally load in a performant manner, even when scrolling results quickly
* preview should not show up without feature flag enabling it (normal quicknav experience without it)

## Checklist

- [X] Added tests


mportiz08:
@swift-ci test

mportiz08:
@sofiaromorales I can't officially add you as a "Reviewer" to the PR, but I would very much appreciate your review if you have the time to check this out (and also no worries if not). Thanks for your hard work on the original quick navigation feature that this builds on!

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -190,10 +216,55 @@ export default {
...
+      (cachedSymbolResults[selectedSymbol.uid] || {}).json
+    ) : (
+      null
+    )),
+    previewState: ({
+      cachedSymbolResults,
+      previewIsLoadingSlowly,
+      selectedSymbol,
+      // eslint-disable-next-line no-nested-ternary
+    }) => (selectedSymbol && Object.hasOwnProperty.call(cachedSymbolResults, selectedSymbol.uid) ? (
```
dobromir-hristov, line 253: could this be simplified to something like this or that would cause false positives?
```suggestion
    }) => (selectedSymbol && cachedSymbolResults[selectedSymbol.uid] ? (
```

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -272,6 +343,78 @@ export default {
...
+
+      // exit early if the previwe data is already cached for this symbol as
+      // there is no additional work needed
+      if (!this.selectedSymbol || this.$cachedSymbolResults.has(this.selectedSymbol.uid)) {
+        clearTimeout(this.loadingTimeout);
+        this.previewIsLoadingSlowly = false;
+        return;
+      }
+
+      const fetchSymbolData = async (symbol) => {
```
dobromir-hristov, line 365: Do you think this can be extracted to its own method?

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -272,6 +343,78 @@ export default {
...
+      this.abortController = new AbortController();
+
+      await Promise.all([
+        fetchSymbolData(this.selectedSymbol).finally(() => {
+          // the timeout for the loading message only applies to the currently
+          // selected symbol
+          clearTimeout(this.loadingTimeout);
+          this.previewIsLoadingSlowly = false;
+        }),
+        fetchSymbolData(this.nextSymbol),
```
dobromir-hristov, line 415: Do we await this resulting promise anywhere? Just checking, because if we did, then waiting for the next item to finish might not be necessary always.

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -272,6 +343,78 @@ export default {
...
+            this.$cachedSymbolResults.set(symbol.uid, {
+              success: false,
+            });
+          }
+        } finally {
+          // `LRUMap` is a custom object that is very similar to a builtin `Map`
+          // and since `Map` objects can't directly hook into the reactivity
+          // in Vue 2, we need to convert this object into a plain `Object` every
+          // time it is mutated to workaround this lack of reactivity
+          this.cachedSymbolResults = Object.fromEntries(this.$cachedSymbolResults);
```
dobromir-hristov, line 393: The only thing that worries me here is that if the `cachedSymbolResults` becomes full with very very big JSON entries, Vue will try to make them deeply reactive, which for example caused issues in the Navigator.

With the Composition API we could have leveraged `shallowReactive`.

20 max items is not much, so maybe we should just keep an eye out. Easiest solution would be to `Object.freeze` it, if we notice perf issues.

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -326,6 +474,16 @@ $base-border-width: 1px;
       background-color: var(--color-navigator-item-hover);
     }
   }
+  &__refs {
+    flex: 1;
+  }
+  &__preview {
+    border-left: $base-border-width solid var(--color-grid);
+    flex: 0 0 61.8%;
```
dobromir-hristov, line 482: Thats oddly specific :D

dobromir-hristov (commented):
Cool stuff :) Left a few minor comments, but overall I think its great

src/components/Navigator/QuickNavigationModal.vue
mportiz08: Yes but it makes the linter angry 😝

src/components/Navigator/QuickNavigationModal.vue
mportiz08: It could be, although it's not used anywhere else yet, so I just kept it as a nested function for now.

src/components/Navigator/QuickNavigationModal.vue
mportiz08: Yeah, on line 408.

I agree that it would be better to optimize this so that it only waits for the first one and not necessarily the next one, although I was a little unclear on exactly how to get that code right—it's a little tricky, because we need to make sure to update the UI after waiting for the first one but also not mess up the next one once it's been selected as the current one. I can see if I can optimize this at all.

mportiz08: Oh, you mean waiting for this whole promise somewhere. I don't believe so, since it's an async watcher.

src/components/Navigator/QuickNavigationModal.vue
mportiz08: Good point, and it technically results in a copy of the full data due to this bit.

I'll see if there's a way around that. Otherwise, I'll look into `Object.freeze`—it sounds like using that would prevent Vue from attaching the getters/setters for the reactive system?

dobromir-hristov: Only shallow ones at the top level yes. You dont care about deep changes anyway.

src/components/Navigator/QuickNavigationModal.vue
hqhhuang: Why does it make the linter angry?

mportiz08: I don't think it applies to this situation or others in the codebase, but I think this is the one that usually pops up from time to time: https://eslint.org/docs/latest/rules/no-prototype-builtins

src/components/Navigator/QuickNavigationModal.vue
hqhhuang: I don't think we would ever have 2 back to back entries that are the same, would it make sense to update the `cachedSymbolResults` only when both the current entry and the next entry finished fetching?

src/utils/lru-map.js
```diff
@@ -0,0 +1,69 @@
...
+    if (!this.map.has(key)) {
+      return undefined;
+    }
+
+    // if the map already contains a value for this key, delete it and add it
+    // again so that the order of the map correlates to how recently the value
+    // for this key was accessed
+    const value = this.map.get(key);
+    this.map.delete(key);
+    this.map.set(key, value);
```
hqhhuang, line 38: This is so clever!

hqhhuang (commented):
This is so cool!! Especially the performance improvement logic :P 

It seems like some of the style changes in `DocumentationTopic` affects minimized mode in general. I will test some more tomorrow.

src/components/Navigator/QuickNavigationModal.vue
mportiz08: 8beabe4d1e1d57f0067be538fd54adc20473056f

src/components/DocumentationTopic.vue
```diff
@@ -485,6 +486,7 @@ export default {
     isOnThisPageNavVisible: ({ topicState }) => (
       topicState.contentWidth > ON_THIS_PAGE_CONTAINER_BREAKPOINT
     ),
+    disableMetadata: ({ enableMinimized }) => enableMinimized,
```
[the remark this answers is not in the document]
mportiz08: Without disabling it, as you scroll through the results, the title of the current page will be changing to the thing you've selected—one thing I noticed that came from re-using the doc component that I had to workaround 😅

tests/unit/components/DocumentationTopic.spec.js
```diff
@@ -460,13 +466,17 @@ describe('DocumentationTopic', () => {
...
       conformance: propsData.conformance,
       declarations: declarationsSection.declarations,
       source: propsData.remoteSource,
     });
-    wrapper.setProps({ enableMinimized: true });
+    // wrapper.setProps({ enableMinimized: true });
+    // commented this out and moved it to the above `setProps` call because
+    // there seems to be an obscure bug with vue-test-utils where things don't
+    // work right if `setProps` is called more than once with a prop that is
+    // also used in the component's `provide`...
```
dobromir-hristov, line 479: 🤔 `setProps` wont update the `provide`, as it's not reactive by design I guess?

dobromir-hristov (approved):
Things look good to me :)

src/components/Navigator/QuickNavigationPreview.vue
```diff
@@ -0,0 +1,136 @@
...
+  padding: 20px;
+
+  &-row {
+    animation: pulse 2.5s ease;
+    animation-delay: calc(1s + 0.3s * var(--index));
+    animation-fill-mode: forwards;
+    animation-iteration-count: infinite;
+    background-color: var(--color-fill-gray-tertiary);
+    border-radius: 4px;
+    height: 12px;
```
marinaaisa, line 123: In the future, as a refactor it would be nice to extract the styles of each loader row and create a component that we could reuse for the loader at  `LoadingNavigatorItem` .

marinaaisa (approved):
Very nice work, @mportiz08 Thanks a lot for working on this!
I like the caching part at the lru map class :)

src/components/Navigator/QuickNavigationModal.vue
```diff
@@ -54,56 +54,64 @@
...
-            :url="symbol.path"
-            @click.native="closeQuickNavigationModal"
-            @focus.native="focusIndex(index)"
-          >
-            <div
-              class="quick-navigation__symbol-match"
-              ref="match"
-              role="list"
-              :class="{ 'selected' : index == focusedIndex }"
+          <div class="quick-navigation__refs">
```
sofiaromorales, line 57: I think the `__refs` style should be applied only when results were found because it's messing with the `no-results` alignment. Rendering this div based on the value of `noResultsWereFound` can do the fix.
Screen Shot 2023-03-03 at 12 49 48 PM

tests/unit/components/DocumentationTopic.spec.js
mportiz08: Yeah, Hanqing pointed that out too, and that makes sense. However, I was seeing weird results where some things were treated as if the flag was true and others as if it was false—it was very confusing 😅

src/components/Navigator/QuickNavigationPreview.vue
mportiz08: Agreed. I looked into doing this but thought it would take too much effort/time since the design of the rows differs slightly here compared to the navigator.

We should try to make a more general component in the future that both places can use for sure though. I mostly copied your styling for the navigator loader here 😁

src/components/Navigator/QuickNavigationModal.vue
mportiz08: Good catch! I'll try to fix the issue with the alignment.

mportiz08: Fixed with 11ea6d44fdd2dd4badebd7ea2137443206c71121 :)

mportiz08:
@swift-ci test

mportiz08:
Merging now with the understanding that this feature is currently only opt-in using a feature flag, which may be later removed when we feel comfortable with this feature being a default experience.

Thanks everyone!

PR #543, opened 2023-03-07
Fix View more link when there's no Primary Content

hqhhuang:
Bug/issue #, if applicable: 106166386

## Summary
Fix the issue where view more link is not displayed when there's no Primary Content

## Dependencies
NA

## Testing
Steps:
1.  Turn on Navigation Preview feature
2. Test with a page with no primary content

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -454,11 +454,13 @@ export default {
       deprecationSummary,
       downloadNotAvailableSummary,
       primaryContentSectionsSanitized,
+      enableMinimized,
     }) => (
       isRequirement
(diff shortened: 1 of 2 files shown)


dobromir-hristov (approved):
Change looks good, I guess we dont have a specific way to know if a viewMoreLink is truthy or not, as its always shown for minimized mode.

@mportiz08 should give it a look as well.

hqhhuang:
There will be scenarios where we don't want to show the link that we will work on in the future. But for now, it is always shown for minimized mode.

hqhhuang:
@swift-ci test

mportiz08 (approved):

PR #544, opened 2023-03-07
Fix documentation page fails to load in minimized route for objc only pages

hqhhuang:
Bug/issue #, if applicable: 106383020 

## Summary
Fix documentation page fails to load in minimized route for objc only pages

## Testing
Load an objc only page in minimized mode

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -369,16 +369,16 @@ export default {
     };
   },
   computed: {
-    normalizedSwiftPath: ({ normalizePath, swiftPath }) => (
-      swiftPath ? normalizePath(swiftPath) : null
-    ),
+    normalizedSwiftPath: ({ normalizePath, swiftPath }) => (normalizePath(swiftPath)),
     normalizedObjcPath: ({
       normalizePath,
       objcPath,
+      swiftPath,
     }) => (
-      objcPath ? buildUrl(normalizePath(objcPath), {
+      // do not append language query parameter if no swiftPath exists
+      normalizePath((objcPath && swiftPath) ? buildUrl(objcPath, {
         language: Language.objectiveC.key.url,
-      }) : null
+      }) : objcPath)
     ),
     defaultImplementationsCount() {
       return (this.defaultImplementationsSections || []).reduce(
@@ -504,6 +504,7 @@ export default {
     normalizePath(path) {
       // Sometimes `paths` data from `variants` are prefixed with a leading
       // slash and sometimes they aren't
+      if (!path) return path;
       return path.startsWith('/') ? path : `/${path}`;
     },
     extractProps(json) {
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -369,7 +369,9 @@ export default {
     };
   },
   computed: {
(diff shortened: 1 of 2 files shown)


hqhhuang:
@swift-ci test

src/components/DocumentationTopic.vue
```diff
@@ -369,7 +369,9 @@ export default {
     };
   },
   computed: {
-    normalizedSwiftPath: ({ normalizePath, swiftPath }) => (normalizePath(swiftPath)),
+    normalizedSwiftPath: ({ normalizePath, swiftPath }) => (
```
mportiz08, line 372: Can we add/update a unit test for this? Maybe the `ViewMore` related specs or something since that seems to be the thing using this computed property?

mportiz08 (approved):

src/components/DocumentationTopic.vue
dobromir-hristov: 😬

src/components/DocumentationTopic.vue
```diff
@@ -502,6 +504,7 @@ export default {
     normalizePath(path) {
       // Sometimes `paths` data from `variants` are prefixed with a leading
       // slash and sometimes they aren't
+      if (!path) return path;
```
hqhhuang, line 507: Added this so we don't have to check if `path` is null everytime before we call this method.

tests/unit/components/DocumentationTopic.spec.js
```diff
@@ -507,6 +507,27 @@ describe('DocumentationTopic', () => {
...
+    expect(objcViewMore.props('url')).toEqual('/documentation/objc'); // normalized path
+
+    wrapper.setProps({
+      objcPath: null,
+      swiftPath: 'documentation/swift',
+      interfaceLanguage: 'swift',
+    });
+    const swiftViewMore = wrapper.find(ViewMore);
+    expect(swiftViewMore.exists()).toBe(true);
+    expect(swiftViewMore.props('url')).toEqual('/documentation/swift'); // normalized path
```
mportiz08, line 528: Looks good, thanks. You may want to add one more additional scenario for the page that has both a Swift/ObjC path since there is also logic for that scenario.

src/components/DocumentationTopic.vue
mportiz08: I can see why you did it this way. Personally, it feels a little odd to expect that `normalizePath` can be called without an actual path, although there's nothing technically wrong with it.

mportiz08 (approved):

tests/unit/components/DocumentationTopic.spec.js
hqhhuang: https://github.com/apple/swift-docc-render/blob/b8ba1d34ec705c2ce4dfa4916dc37ee4945d1847/tests/unit/components/DocumentationTopic.spec.js#L493

This test should be testing the scenario where we have both paths & by default we have `interfaceLanguage` set to `swift`. Do you think that's enough?

mportiz08: I think that's ok. It just might be good to have one additional one for the scenario where both variants are available and you're on the ObjC one, since that would trigger your logic in this PR for the query param.

hqhhuang:
@swift-ci test

PR #547, opened 2023-03-10
Refactor for i18n utils and LocaleSelector

marinaaisa:
Bug/issue #106158283, if applicable: 

## Summary

Refactor for i18n utils and LocaleSelector

## Dependencies

NA

## Testing

Steps:
1. Add the following to `theme-settings.json` file at public folder
```json
{
  "features": {
    "docs": {
      "i18n": {
        "enable": true
      }
    }
  }
}
```
3. Run `npm run serve`
2. Go to the bottom of the page and change the language in the locale selector
3. Assert that everything works properly

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/LocaleSelector.vue
+++ b/src/components/LocaleSelector.vue
@@ -13,7 +13,7 @@
     <select
       :value="$i18n.locale"
       :aria-label="$t('select-language')"
-      @change="updateRouter"
+      @change="$emit('change', $event)"
     >
       <option
         v-for="{ code, name } in locales"
@@ -30,8 +30,6 @@
 <script>
 import ChevronThickIcon from 'theme/components/Icons/ChevronThickIcon.vue';
 import locales from 'theme/lang/locales.json';
-import { defaultLocale } from 'theme/lang/index.js';
-import { updateLangTag } from 'docc-render/utils/metadata';
 
 export default {
   name: 'LocaleSelector',
@@ -42,17 +40,6 @@ export default {
     return {
       locales,
     };
-  },
-  methods: {
-    updateRouter({ target: { value: currentLocale } }) {
-      this.$i18n.locale = currentLocale;
-      this.$router.push({
-        params: {
-          locale: currentLocale === defaultLocale ? null : currentLocale,
-        },
-      });
-      updateLangTag(currentLocale);
-    },
   },
 };
 
--- a/src/components/LocaleSelector.vue
+++ b/src/components/LocaleSelector.vue
@@ -13,7 +13,7 @@
     <select
       :value="$i18n.locale"
       :aria-label="$t('select-language')"
-      @change="$emit('change', $event)"
+      @change="updateRouter"
     >
       <option
         v-for="{ code, name } in locales"
@@ -30,6 +30,8 @@
 <script>
 import ChevronThickIcon from 'theme/components/Icons/ChevronThickIcon.vue';
 import locales from 'theme/lang/locales.json';
+import { defaultLocale } from 'theme/lang/index.js';
+import { updateLangTag } from 'docc-render/utils/metadata';
 
 export default {
   name: 'LocaleSelector',
@@ -40,6 +42,18 @@ export default {
     return {
       locales,
     };
+  },
+  methods: {
+    updateRouter({ target: { value: currentLocale } }) {
+      const { slug } = locales.find(locale => locale.code === currentLocale);
+      this.$i18n.locale = currentLocale;
+      this.$router.push({
+        params: {
+          locale: currentLocale === defaultLocale ? null : slug,
+        },
+      });
+      updateLangTag(currentLocale);
+    },
   },
 };
 
--- a/src/components/LocaleSelector.vue
+++ b/src/components/LocaleSelector.vue
@@ -16,9 +16,10 @@
       @change="updateRouter"
     >
       <option
-        v-for="{ code, name } in locales"
-        :key="code"
-        :value="code"
+        v-for="{ slug, name, code } in locales"
+        :key="slug"
+        :value="slug"
+        :lang="code"
       >
         {{ name }}
       </option>
@@ -31,8 +32,7 @@
 import ChevronThickIcon from 'theme/components/Icons/ChevronThickIcon.vue';
 import locales from 'theme/lang/locales.json';
 import { defaultLocale } from 'theme/lang/index.js';
-import { updateLangTag } from 'docc-render/utils/metadata';
-import { getSlug } from 'docc-render/utils/i18n-utils';
+import { updateLocale } from 'docc-render/utils/i18n-utils';
 
 export default {
   name: 'LocaleSelector',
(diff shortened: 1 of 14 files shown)


marinaaisa:
@swift-ci test

src/utils/data.js
```diff
@@ -138,7 +139,7 @@ export function clone(jsonObject) {
 }
 
 export async function fetchIndexPathsData({ currentLocale }) {
-  const locale = currentLocale === defaultLocale ? '' : currentLocale;
-  const path = new URL(`${pathJoin([baseUrl, 'index/', locale, 'index.json'])}`, window.location.href);
+  const slug = currentLocale === defaultLocale ? '' : getSlug(currentLocale);
```
mportiz08, line 142: I don't think you would need any changes to this file if you were to make a minor change in `NavigatorDataProvider`—instead of passing it the value of `$i18n.locale` as a prop, it should simply pass `$route.params.locale` to this function, which should always be the URL slug so no conversion would be needed here.

src/lang/locales.json
```diff
@@ -1,14 +1,17 @@
 [
   {
     "code": "en-US",
-    "name": "English"
+    "name": "English",
+    "slug": "en-US"
```
mportiz08, line 5: I think this is probably a good way of allowing the slug to be configurable by other systems, even if it matches the language tag exactly by default. 👍

src/utils/i18n-utils.js
```diff
@@ -25,16 +25,34 @@ export function localeIsValid(localeCode) {
...
+  const currentLocale = to.params.locale;
+  updateLocale(currentLocale, env);
+}
+
+/**
+ * Get locale slug from locale code
+ * @param {String} localeCode - locale code
+ * @return {String}
+ */
+export function getSlug(localeCode) {
```
mportiz08, line 55: Not really blocking feedback but an optimization of this to avoid doing a search with `find` every time this is called would be to create a mapping at the top level so this function is always a constant lookup.

Something like:
```js
const slugForLocale = locales.reduce((map, locale) => ({
  ...map,
  [locale.slug]: locale.code,
}), {});

// ...

export function getSlugForLocale(localeCode) {
  return slugForLocale(localeCode);
}
```

src/components/LocaleSelector.vue
```diff
@@ -45,10 +46,11 @@ export default {
   },
   methods: {
     updateRouter({ target: { value: currentLocale } }) {
+      const slug = getSlug(currentLocale);
```
mportiz08, line 49: To minimize the amount of necessary code/logic, I think you might actually want to populate the selector items with slug values. Since you're going to need to convert slugs to locale codes in the router guards (`updateCurrentLocale`) _anyway_, it would save you from also doing the opposite conversion here.

Essentially, you would be doing the reverse of this, ie `getCode(localeSlug)`—however, you wouldn't then need to add the reverse mapping in the router guard (which is missing here right now since it assumes the slugs are always the same).

src/utils/i18n-utils.js
```diff
@@ -25,16 +25,34 @@ export function localeIsValid(localeCode) {
...
+}
+
+/**
+ * Updates i18n global var and html lang
+ * @param {{ params: { locale: String } }} to - where the route navigates to
+ * @param {Object{}} env - context
+ */
+export function updateCurrentLocale(to, env) {
+  const currentLocale = to.params.locale;
+  updateLocale(currentLocale, env);
```
mportiz08, line 47: This is missing code to map the URL slug (`currentLocale`) to the appropriate locale code that `$i18n` expects and only works here, because it assumes that the slug will always be an exact match.

src/utils/i18n-utils.js
```diff
@@ -25,16 +25,34 @@ export function localeIsValid(localeCode) {
...
 /**
  * Updates i18n global var and html lang
- * @param {{ params: { locale: String } }} to - where the route navigates to
+ * @param {String} locale - locale used
  * @param {Object{}} env - context
- * @param {{ code: String, name: String }[]} locales
  */
-export function updateCurrentLocale(to, env) {
-  const currentLocale = to.params.locale || defaultLocale;
+export function updateLocale(locale = defaultLocale, env) {
```
mportiz08, line 31: I'm not sure we would need to export this function since it is only used once in the same file—maybe it doesn't even need to be a separate function?

src/utils/i18n-utils.js
mportiz08: (Or the reverse if you end up doing the reverse mapping for `getLocaleCode(localeSlug)` instead of course)

src/components/LocaleSelector.vue
```diff
@@ -45,10 +46,11 @@ export default {
...
       this.$i18n.locale = currentLocale;
```
mportiz08, line 50: You may want to consider re-using the `updateCurrentLocale` utility to avoid repeating the `$i18n` and `<meta>` logic here.

mportiz08 (commented):
I think we're getting pretty close to how we want this to work. I covered this in the code, but I think we'll want to make sure to map URL slugs to locale codes if we're going to expose that as a configuration, even if they are the same in the default case.

The other comments are pretty minor and mostly optimizing things.

src/utils/i18n-utils.js
marinaaisa: great optimization! thanks a lot :)

marinaaisa:
Ready for another round!

dobromir-hristov (approved):
Looks good to me. @hqhhuang take a look as well.

src/utils/i18n-utils.js
```diff
@@ -8,33 +8,44 @@
...
-import appLocales from 'theme/lang/locales.json';
+import locales from 'theme/lang/locales.json';
 import { defaultLocale } from 'theme/lang/index.js';
 import { updateLangTag } from 'docc-render/utils/metadata';
 
-const localeCodes = new Set(appLocales.map(appLocale => appLocale.code));
+const slugsForLocale = locales.reduce((map, locale) => ({
+  ...map,
+  [locale.slug]: locale.code,
+}), {});
```
hqhhuang, line 18: I think it would be a little less confusing to name this `codeForLocale`? Especially since we are using it to get locale code from slug.

hqhhuang (approved):
Looks good to me as well! Just a nitpick - should we also test the scenario when the slug is different than the locale name, just to make sure it is set to the right thing?

src/utils/i18n-utils.js
marinaaisa: I replaced `codeForLocale` with `codeForSlug`, you are right that it's easier to understand

hqhhuang (approved):
LGTM! Thanks Marina!

src/utils/data.js
```diff
@@ -137,8 +136,7 @@ export function clone(jsonObject) {
   return JSON.parse(JSON.stringify(jsonObject));
 }
 
-export async function fetchIndexPathsData({ currentLocale }) {
-  const locale = currentLocale === defaultLocale ? '' : currentLocale;
-  const path = new URL(`${pathJoin([baseUrl, 'index/', locale, 'index.json'])}`, window.location.href);
+export async function fetchIndexPathsData({ slug }) {
```
mportiz08, line 139: I know this was already merged, but it might be good to rename this in a future PR to make it more clear that the slug here refers to the slug for a locale and not something else, like the technology name or whatever.

PR #551, opened 2023-03-15
Make highlight.js languages user-configurable

carson-katri:
Bug/issue #, if applicable: N/A

## Summary

In some cases, documentation may use codeblocks with languages other than those enabled by default in `swift-docc-render`. In our case, we want to include Elixir codeblocks in the [LiveViewNative](https://github.com/liveview-native/liveview-client-swiftui) docs.

This PR makes it possible to build with a custom set of supported languages which are added to the default set with an environment variable:

```sh
VUE_APP_HLJS_LANGUAGES=elixir npm run build
```

Multiple languages can be added with a comma-separated list:

```sh
VUE_APP_HLJS_LANGUAGES=elixir,haskell npm run build
```

Now codeblocks for Elixir are properly highlighted!

Screenshot 2023-03-15 at 1 27 08 PM

I used webpack's `ContextReplacementPlugin` to ensure only the necessary languages are included in the bundle. This replaces the `webpackInclude` magic comment in `src/utils/syntax-highlight.js`.

## Dependencies

N/A

## Testing

Steps:
1. Write documentation with a codeblock in an unsupported language, such as Elixir.

```markdown
    ```elixir
    def hello do
        :world
    end
    ```
```

2. Build `swift-docc-render` with the `VUE_APP_HLJS_LANGUAGES` variable.

```sh
VUE_APP_HLJS_LANGUAGES=elixir npm run build
```

3. Preview the documentation and confirm the codeblock is properly highlighted.

```sh
DOCC_HTML_DIR=dist xcrun docc preview SwiftDocCRender.docc
```

## Checklist

- [ ] Added tests - I am unsure how to test environment variables, I'd be happy to add tests if needed.

### DIFF
--- a/src/utils/syntax-highlight.js
+++ b/src/utils/syntax-highlight.js
@@ -37,6 +37,12 @@ const Languages = {
   shell: ['console', 'shellsession'],
   swift: [],
   xml: ['html', 'xhtml', 'rss', 'atom', 'xjb', 'xsd', 'xsl', 'plist', 'wsf', 'svg'],
+  // load more languages from the environment
+  ...(
+    process.env.VUE_APP_HLJS_LANGUAGES
+      ? Object.fromEntries(process.env.VUE_APP_HLJS_LANGUAGES.split(',').map(l => [l, []]))
+      : undefined
+  ),
 };
 
 export const CustomLanguagesSet = new Set([
@@ -83,8 +89,6 @@ async function importHighlightFileForLanguage(language) {
       } else {
         languageFile = await import(
           /* webpackChunkName: "highlight-js-[request]" */
-          // eslint-disable-next-line max-len
-          /* webpackInclude: /\/(bash|c|s?css|cpp|diff|http|java|llvm|perl|php|python|ruby|xml|javascript|json|markdown|objectivec|shell|swift)\.js$/ */
           `highlight.js/lib/languages/${file}`
         );
       }
--- a/SwiftDocCRender.docc/contributing/Internals.md
+++ b/SwiftDocCRender.docc/contributing/Internals.md
@@ -177,6 +177,7 @@ DocC-Render has a few build-time environment flags, that allow you to set config
 
 * **VUE_APP_DEV_SERVER_PROXY** - The HTTP endpoint or  local filepath to read render JSON from when using the development server
 * **VUE_APP_TITLE** - An optional default page title to apply to pages
+* **VUE_APP_HLJS_LANGUAGES** - An optional comma-separated list of highlight.js languages to include in the build
 
 #### Available Scripts
 
--- a/src/setup-utils/vue-config-utils.js
+++ b/src/setup-utils/vue-config-utils.js
@@ -12,6 +12,8 @@ const fs = require('fs');
 const path = require('path');
 // eslint-disable-next-line import/no-extraneous-dependencies
 const ThemeResolverPlugin = require('webpack-theme-resolver-plugin');
+// eslint-disable-next-line import/no-extraneous-dependencies
+const webpack = require('webpack');
 const themeUtils = require('./theme-build-utils');
 const { BannerPlugin, LICENSE_HEADER } = require('./license-header-built-files');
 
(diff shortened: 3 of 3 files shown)


vue.config.js
```diff
@@ -31,5 +32,16 @@ module.exports = vueUtils({
...
+
+    // limit highlight.js to only the necessary languages
+    const builtinLanguages = 'bash|c|s?css|cpp|diff|http|java|llvm|perl|php|python|ruby|xml|javascript|json|markdown|objectivec|shell|swift';
+    const envLanguages = (process.env.VUE_APP_HLJS_LANGUAGES ?? '').split(',').join('|');
+    config
+      .plugin('languages')
+      .use(webpack.ContextReplacementPlugin, [
+        /highlight\.js\/lib\/languages$/,
+        new RegExp(`\/(${[builtinLanguages, envLanguages].join('|')})$`),
+        /\.js$/,
```
dobromir-hristov, line 44: i think the `.js` part should be inside the regex, no? Otherwise that is the third parameter to `ContextReplacementPlugin`, which seems to be a boolean - https://webpack.js.org/plugins/context-replacement-plugin/#usage

Example of what you did:

https://github.com/date-fns/date-fns/blob/main/docs/webpack.md

dobromir-hristov (commented):
This is a pretty neat idea. Left a minor comment, but overall I think this could be useful.

src/utils/syntax-highlight.js
```diff
@@ -83,8 +89,6 @@ async function importHighlightFileForLanguage(language) {
       } else {
         languageFile = await import(
           /* webpackChunkName: "highlight-js-[request]" */
-          // eslint-disable-next-line max-len
```
mportiz08, line 86: I love the idea of making the set of supported languages configurable, but I'm running into issues testing this. I'm not sure if it's just a mistaken assumption on how it should work though.

My assumption is that the `VUE_APP_HLJS_LANGAGES` variable would be an opt-in feature to allow supporting additional languages not present in the default build and that the normal behavior should be unchanged if the flag is not provided—is that correct?

If so, there may be an issue with how this is implemented, because I'm finding that existing JavaScript code listings are no longer highlighted even when not specifying this env var (as one example, haven't tried others yet).

mportiz08: I think this line might be the problematic one, since I believe this magic comment is what allowed us to dynamically reference the necessary syntax file for a given page as needed.

mportiz08: To reproduce the issue I'm seeing:

Run `npm run docs:preview` and load http://localhost:8080/documentation/swift-docc-render/internals to see the js/html highlighted code listings

Then build your branch with `npm run build` and run it again with `DOCC_HTML_DIR=/path/to/swift-docc-render/dist npm run docs:preview` and load http://localhost:8080/documentation/swift-docc-render/internals and note that the listings are no longer highlighted, even without specifying the new `$VUE_APP_HLJS_LANGUAGES` var

carson-katri: Good call, I had broken it with the last changes. Removing `\.js` from the regex fixed it, as that was adding `-js` to the name of the all the language files in the bundle. Now it correctly highlights the original set of languages and any additional ones if specified.

mportiz08: Awesome, glad to find out it was such an easy fix!

mportiz08 (commented):
Looks good now from a functional perspective—thanks for the contribution!

I think there are some minor linter issues to address before we can merge this. You can see them by running `npm test` or one of the more specific linter scripts.

vue.config.js
```diff
@@ -9,6 +9,7 @@
 */
 
 const path = require('path');
+const webpack = require('webpack');
```
dobromir-hristov, line 12: eslint is throwing an error here, because webpack is a dev dep. Generally this can be safely ignored, with an eslint ignore comment.

vue.config.js
```diff
@@ -31,5 +32,15 @@ module.exports = vueUtils({
...
       }]]);
+
+    // limit highlight.js to only the necessary languages
+    const builtinLanguages = 'bash|c|s?css|cpp|diff|http|java|llvm|perl|php|python|ruby|xml|javascript|json|markdown|objectivec|shell|swift';
+    const envLanguages = (process.env.VUE_APP_HLJS_LANGUAGES ?? '').split(',').join('|');
+    config
+      .plugin('languages')
+      .use(webpack.ContextReplacementPlugin, [
+        /highlight\.js\/lib\/languages$/,
+        new RegExp(`\/(${[builtinLanguages, envLanguages].join('|')})$`),
```
dobromir-hristov, line 43: I think you dont need to escape here, since you are writing a string inside a RegExp, it will escape it for you.

```suggestion
        new RegExp(`/(${[builtinLanguages, envLanguages].join('|')})$`),
```

dobromir-hristov (changes requested):
Left two small comments, regarding the code.

I would however suggest you move this config code into the `vue-config-utils.js` file, in `baseChainWebpack`, so that everyone that extends upon`docc-render`, can benefit from this functionality.

As it is now, it will only be used for this app.

carson-katri:
I moved the logic into `vue-config-utils.js` and resolved the linter errors.

dobromir-hristov (approved):
This is looking great. @mportiz08 give a look too and I think we are good to go.

mportiz08:
@swift-ci test

mportiz08 (approved):

dobromir-hristov:
Thank you very much for your contribution :)

PR #556, opened 2023-03-27
Remove the WordBreak component from the OnThisPageNav items

dobromir-hristov:
Bug/issue #, if applicable: 107193315

## Summary

Removes the WordBreak component from the OnThisPage as it was breaking `camelCaseText` in weird ways sometimes.

## Note

H1 items, aka the page titles are often code-like strings, when considering code documentation, removing the WordBreak may cause regressions on those cases. 

## Dependencies

NA

## Testing

Open pages that have camelCaseWords in the sections headings and assert it does not break them up.

## Checklist

- [ ] Added tests

### DIFF
--- a/src/components/OnThisPageNav.vue
+++ b/src/components/OnThisPageNav.vue
@@ -20,8 +20,7 @@
           class="base-link"
           @click.native="handleFocusAndScroll(item.anchor)"
         >
-          <WordBreak v-if="item.i18n">{{ $t(item.title) }}</WordBreak>
-          <WordBreak v-else>{{ item.title }}</WordBreak>
+          <span>{{ item.i18n ? $t(item.title): item.title }}</span>
         </router-link>
       </li>
     </ul>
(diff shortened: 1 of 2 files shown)


src/components/OnThisPageNav.vue
```diff
@@ -20,8 +20,7 @@
           class="base-link"
           @click.native="handleFocusAndScroll(item.anchor)"
         >
-          <WordBreak v-if="item.i18n">{{ $t(item.title) }}</WordBreak>
-          <WordBreak v-else>{{ item.title }}</WordBreak>
+          <span>{{ item.i18n ? $t(item.title): item.title }}</span>
```
mportiz08, line 23: Would there even be a need for the span at all if we were to make this change?

dobromir-hristov: No, you are right.

dobromir-hristov:
> I don't expect that there would be a clear correlation between the page type and the types of content used in the headings, although I could be wrong.

@mportiz08 the page titles of pages are added as the first items in the OnThisPage, aka H1 of the page. Those are in fact in most cases code-like

image

mportiz08:
> > I don't expect that there would be a clear correlation between the page type and the types of content used in the headings, although I could be wrong.
> 
> @mportiz08 the page titles of pages are added as the first items in the OnThisPage, aka H1 of the page. Those are in fact in most cases code-like
> 
> image

Ah, didn't realize/remember that. In that case, I don't think the PR would be very helpful as-is 😢

PR #560, opened 2023-03-28
Images can eagerly load in IDE mode

hqhhuang:
Bug/issue #, if applicable: rdar://107320467

## Summary
Images should be allowed to eagerly load by default in IDE mode.

### DIFF
--- a/src/stores/AppStore.js
+++ b/src/stores/AppStore.js
@@ -22,7 +22,8 @@ const defaultColorScheme = supportsAutoColorScheme ? ColorScheme.auto : ColorSch
 
 export default {
   state: {
-    imageLoadingStrategy: ImageLoadingStrategy.lazy,
+    imageLoadingStrategy: process.env.VUE_APP_TARGET === 'ide'
+      ? ImageLoadingStrategy.eager : ImageLoadingStrategy.lazy,
     preferredColorScheme: Settings.preferredColorScheme || defaultColorScheme,
     supportsAutoColorScheme,
     systemColorScheme: ColorScheme.light,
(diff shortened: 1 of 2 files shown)


src/stores/AppStore.js
```diff
@@ -22,7 +22,8 @@ const defaultColorScheme = supportsAutoColorScheme ? ColorScheme.auto : ColorSch
 
 export default {
   state: {
-    imageLoadingStrategy: ImageLoadingStrategy.lazy,
+    imageLoadingStrategy: process.env.VUE_APP_TARGET === 'ide'
```
mportiz08, line 25: Can you add a test for this?

mportiz08 (approved):

mportiz08:
@swift-ci test

src/stores/AppStore.js
dobromir-hristov: If you want to clean it up a bit you can extract the load strategy to a variable:

```js
const strategy = process.env.VUE_APP_TARGET === 'ide'
      ? ImageLoadingStrategy.eager : ImageLoadingStrategy.lazy
```

dobromir-hristov (approved):
Just left a minor comment, but overall I think its good.

If we switch to Pinia at some point, lots of things might get simpler with these stores, like resetting for example.

src/stores/AppStore.js
hqhhuang: Discussed with Dobri - we need to make sure to re-import the file during testing, which isn't worth it to me. So I'm gonna keep it this way :)

hqhhuang:
@swift-ci test

PR #563, opened 2023-03-30
Prevent scrolling on `navigation` event if router `meta` field `preventScrolling` is `true`

hqhhuang:
Bug/issue #, if applicable: rdar://103146273

## Summary
If `preventScrolling` in the `meta` field is set to `true`, then we should not change the scroller location when navigating to a new page.

## Dependencies
rdar://103119754

## Testing
Steps:
1. set `preventScrolling` to `true` for a page (i.e. '/foo')
2. Scroll down a little bit on any page (i.e. '/bar').  ('/foo' should be longer than '/bar')
3. trigger a navigation event to '/foo'
4. Verify that the scroller did not move up/down nor left/right

## Checklist

- [x] Added tests


src/utils/router-utils.js
```diff
@@ -48,6 +48,12 @@ export async function scrollBehavior(to, from, savedPosition) {
     await this.app.$nextTick();
     return savedPosition;
   }
+  if (areEquivalentLocations(to, from)
```
dobromir-hristov, line 51: `areEquivalentLocations` does not detect hash changes, so say I move across hashes, like in the OnThisPageNav, wont this prevent the scrolling?

If so, then leaving the `to.meta.preventScrolling` on its own and leaving the `areEquivalentLocations` check down there is a better option.

dobromir-hristov: It does prevent it... Just tried

dobromir-hristov (changes requested):

hqhhuang:
@swift-ci test

PR #514, opened 2023-01-10
Ensure scrollbar is visually clipped by code listings with round borders in Safari

mportiz08:
Bug/issue #, if applicable: 104043983

## Summary

This fixes an issue In Safari where the browser scrollbar UI is not being visually clipped by rounded borders of code listings. 

By default, code listings don't have a round enough border for this problem to be observed, but it could present itself with an updated theme with a larger border radius for these elements.

### Example

Before:
Screenshot 2023-01-10 at 3 25 46 PM

After:
Screenshot 2023-01-10 at 3 27 04 PM

## Testing

Steps:
1. Configure the global macOS setting as follows ` / System Settings... / Appearance / Show scroll bars` to `"Always"`
2. Run `env VUE_APP_DEV_SERVER_PROXY=https://mportiz08.github.io/swift-docc npm run serve` or run this branch with any other example content
3. Increase the default border-radii settings for testing purposes. A quick/easy way is to open the inspector and click the `<body>` element—then you can directly add the following CSS variables to it `--border-radius: 15px;`, `--code-border-radius: 15px;` (this bug can't be seen without doing this)
4. Find code listings that scroll horizontally (or make them scroll by changing the size of the window/sidebar) and ensure that the scrollbar isn't extending past the rounded corners.
5. Test with other browsers and ensure that there aren't any obvious regressions from these changes.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [ ] ~~Added tests~~ (CSS only change)
- [X] Ran `npm test`, and it succeeded
- [X] Updated documentation if necessary

PR #569, opened 2023-04-03
Extract helper for establishing a new stacking context

mportiz08:
Bug/issue #, if applicable: 104872474

## Summary

This extracts and modifies the existing functionality needed to establish a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) without altering any visuals, which is used in various places to work around browser specific issues related to overflow and border-radii.

The [previous method](https://github.com/apple/swift-docc-render/pull/514) has issues with the "Export as PDF" feature of Safari, possibly because of the way it handles transparent background images for that kind of rendering output, and it cannot be avoided with media queries the same way that it can when "printing" to PDF.

## Testing

Follow the same testing steps as https://github.com/apple/swift-docc-render/pull/514 but also verify that there are no issues with the "Export as PDF" functionality of Safari.

## Checklist

### DIFF
--- a/src/components/ContentNode/CodeListing.vue
+++ b/src/components/ContentNode/CodeListing.vue
@@ -188,14 +188,9 @@ code {
   min-height: 100%;
   border-radius: var(--code-border-radius, $border-radius);
   overflow: hidden;
-  // this mask image is not actually used for any visual effect since there is
-  // no background being used on this element—however, we need this in order to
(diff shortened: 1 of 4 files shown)


dobromir-hristov (approved):
🚀

mportiz08:
@swift-ci test

PR #587, opened 2023-04-11
Use `WordBreak` based on the page `role`, inside the OnThisPageNav

dobromir-hristov:
Bug/issue #, if applicable: 107193315

## Summary

Refactors the OnThisPageNav to use the page `Role` in order to determine if using the WordBreak is necessary or not.

Article and Collection pages most often dont need the WordBreak

## Dependencies

NA

## Testing

Steps:
1. Use a docc archive that has both code and articles in order to test this out, like the DocC docs

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/OnThisPageNav.vue
+++ b/src/components/OnThisPageNav.vue
@@ -35,7 +35,7 @@ import { waitFrames } from 'docc-render/utils/loading';
 import ScrollToElement from 'docc-render/mixins/scrollToElement';
 import { buildUrl } from 'docc-render/utils/url-helper';
 import WordBreak from 'docc-render/components/WordBreak.vue';
-import { TopicRole } from '@/constants/roles';
+import { TopicRole } from 'docc-render/constants/roles';
 
 export default {
   name: 'OnThisPageNav',
--- a/src/components/OnThisPageNav.vue
+++ b/src/components/OnThisPageNav.vue
@@ -20,8 +20,9 @@
           class="base-link"
           @click.native="handleFocusAndScroll(item.anchor)"
         >
-          <WordBreak v-if="item.i18n">{{ $t(item.title) }}</WordBreak>
-          <WordBreak v-else>{{ item.title }}</WordBreak>
+          <component :is="wordWrapComponent">
+            {{ item.i18n ? $t(item.title) : item.title }}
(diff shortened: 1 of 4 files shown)


dobromir-hristov:
@swift-ci test

tests/unit/components/OnThisPageNav.spec.js
```diff
@@ -6,13 +6,14 @@
  *
  * See https://swift.org/LICENSE.txt for license information
  * See https://swift.org/CONTRIBUTORS.txt for Swift project authors
-*/
+ */
```
dobromir-hristov, line 9:
```suggestion
*/
```

dobromir-hristov:
@swift-ci test

SamLanier (approved):
Design approved.

src/components/OnThisPageNav.vue
```diff
@@ -54,6 +62,13 @@ export default {
       url: buildUrl(`#${item.anchor}`, $route.query),
     })),
     currentPageAnchor: ({ store }) => store.state.currentPageAnchor,
+    wordWrapComponent({ role }) {
+      // Articles and Collection pages normally dont need special wrapping for their titles.
+      if (role === TopicRole.collection || role === TopicRole.article) {
```
mportiz08, line 67: Would it be better to invert this maybe and only use the `WordBreak` for symbol-like roles? I expect that the more common case would be article-like pages, but I may be wrong.

You might have already looked into it and found that this approach works better though—just curious.

dobromir-hristov: I was not sure how many other cases we may have, hence why I did it like this.

If you think the `symbol` role is the only one we may need the break for, then I will refactor it, no prob.

mportiz08: I looked into this more extensively today, and I think that approach should generally be fine.

The only troubling case would be when a symbol name is used in an overview subheading, but we'll never have enough context to know that an actual symbol is being referenced in that situation, but it should be fairly uncommon and not an awful experience to not have the improved breaks in that situation.

I went ahead and opened #701 for that approach.

dobromir-hristov:
Closing in favour of #701

PR #594, opened 2023-04-13
Refactor: Move References to Stores

hqhhuang:
Bug/issue #, if applicable:  107635892

## Summary
Refactor to move references into respective page stores.


hqhhuang:
@swift-ci test

src/views/TutorialsOverview.vue
```diff
@@ -78,6 +90,7 @@ export default {
       this.$nextTick(() => {
         // Send a 'rendered' message to the host when new data has been patched onto the DOM.
         this.newContentMounted();
+        this.store.setReferences(this.overviewProps.references);
```
dobromir-hristov, line 93: If you really want to avoid adding this code here, you may be able to move the setting of the references into the related `/component` child of this page, similar to how you do it for the Article and Topic. 

The `references` at that point will be re-calculated as the whole component is re-rendered.

dobromir-hristov (commented):
At first glance things look good. We could have some mixin to further cleanup all the repetitive reference declaration logic, but apart from that, I think we are good.

src/components/ContentNode.vue
```diff
@@ -468,14 +468,16 @@ export default {
...
-      default() {
-        return {};
-      },
+    store: {
+      default: () => ({
+        state: {
+          references: {},
+        },
+      }),
     },
```
marinaaisa, line 481: why in some components we are injecting default values and in some others we don't?

```js
inject: {
    store: {
      default: () => ({
        state: {
          references: {},
        },
      }),
    },
 },
```

vs

```js
inject: ['store'],
```
do we need these default values at all?

marinaaisa (commented):
Great job! so many changes :)
I'll continue testing tomorrow, in my meantime I left 2 comments

src/components/TutorialsOverview.vue
```diff
@@ -59,16 +57,19 @@ export default {
   constants: { SectionKind },
   inject: {
     isTargetIDE: { default: false },
+    store: {
```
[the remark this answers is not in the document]
hqhhuang: Yes! We are using it for rendering things on the Tutorial Overviews page (such as chapter names)

src/components/ContentNode.vue
hqhhuang: I'm curious about that as well. I think in certain components we can be sure that we always provided the reference, but that's not true in other components. @mportiz08 Do you know why we sometimes don't provide defaults when injecting?

src/components/TutorialsOverview.vue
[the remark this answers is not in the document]
hqhhuang: Just took Dobri's advice and moved the store to this component, instead of the /views/TutorialsOverview component

hqhhuang:
@swift-ci test

src/components/TutorialsOverview.vue
```diff
@@ -97,12 +96,12 @@ export default {
   },
   provide() {
     return {
-      references: this.references,
       store: this.store,
     };
   },
   created() {
     this.store.reset();
+    this.store.setReferences(this.references);
```
marinaaisa, line 104: In this file we aren't injecting a default for state for his.store.reset() or this.store.setReferences(). Could you add it?

src/views/DocumentationTopic.vue
```diff
@@ -384,6 +384,7 @@ export default {
       this.$nextTick(() => {
         // Send a 'rendered' message to the host when new data has been patched onto the DOM.
         this.newContentMounted();
+        this.store.setReferences(this.topicProps.references);
```
marinaaisa, line 387: In this file we aren't injecting a default for state for this.store.setReferences(). Could you add it?

src/mixins/referencesProvider.js
```diff
@@ -0,0 +1,28 @@
...
+export default {
+  // inject the `store`
+  inject: {
+    store: {
+      default: () => ({
+        state: {
+          references: {},
+        },
+        setReferences() {},
+        reset() {},
```
marinaaisa, line 20: I noticed that we aren't using `setReferences` or `reset` in none of the files where this mixing is used for. We could consider deleting their default values

marinaaisa (changes requested):
Hey @hqhhuang . Very clever job on creating a mixing to add the default values for the state's references! 👏  It's much clearer now!
I added some minor comments. Thanks!

src/mixins/referencesProvider.js
hqhhuang: I added it here for testing purposes.

src/components/Tutorial.vue
```diff
@@ -136,14 +136,14 @@ export default {
   },
   created() {
     this.store.reset();
+    this.store.setReferences(this.references);
```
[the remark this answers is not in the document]
hqhhuang: Doesn't seem like we need a default state in this component, since it's an immediate child of `Topic.vue` were we provide the store. I think we are fine here.

src/components/TutorialsOverview.vue
hqhhuang: In this file we are directly importing the `TutorialsOverviewStore` from the source file, which I believe included the defaults.

src/views/DocumentationTopic.vue
hqhhuang: In this file we are directly importing the `DocumentationTopicStore` from the source file, which I believe included the defaults.

swift-docc-render-0.1.0.tgz
mportiz08, line 1: Can you remove this?

src/components/Tutorial.vue
mportiz08, line 139: I'm noticing a pattern where you call `setReferences` immediately after the `reset` call. This seems to indicate that you don't really want the "reset" to actually reset the references—in that case, maybe it would be better to just remove the logic for resetting the references in `reset` so that you don't need this additional call following every reset?

src/components/Tutorial/CodePreview.vue
```diff
@@ -65,11 +65,20 @@ function scaledSize({ width, height }, scale = 1) {
...
-  inject: [
-    'references',
-    'isTargetIDE',
-    'store',
-  ],
+  inject: {
+    isTargetIDE: {
+      default: false,
+    },
+    store: {
```
mportiz08, line 72: Is there a specific reason why you're not using the mixin here?

src/mixins/referencesProvider.js
```diff
@@ -0,0 +1,28 @@
...
+      }),
+    },
+  },
+  computed: {
+    // exposes the apiChanges for the current page
```
mportiz08, line 25: I'm not sure this comment is right

src/views/DocumentationTopic.vue
mportiz08: Should this logic be added to other views as well?

mportiz08 (commented):
Haven't had time to test this yet, but I did an initial pass on reviewing the code.

src/components/Tutorial.vue
hqhhuang: But the `reset` function, as the name suggests, should reset everything in the store. We are only setting references in the few components that are actually then providing the store. Is it a problem that we call `setReferences` immediately after?

src/components/Tutorial/CodePreview.vue
hqhhuang: We renamed the state `tutorialsState` in the original implementation and I wasn't sure why we did that, so I kept it the way it is.

src/mixins/referencesProvider.js
hqhhuang: ops

src/views/DocumentationTopic.vue
hqhhuang: For other views, I added the logic into the related ` /components` child like Dobri suggested here: https://github.com/apple/swift-docc-render/pull/594#discussion_r1165073228

That method doesn't really work with `DocumentationTopic` pages because it has so many child components. While the `Topic` component only has 2, and the `TutorialOverview` component only has one.

src/mixins/referencesProvider.js
hqhhuang: fixed

src/components/Tutorial.vue
mportiz08: It just seemed odd to me that we are basically doing the equivalent of the following with 2 different calls:

```
this.references = {};
this.references = other;
```

Maybe you could just pass the new references to the reset function instead? Or is there ever a case where we want to reset the store and _not_ also update the references?

src/components/ContentNode.vue
mportiz08: Mostly we don't ever need defaults since some top-level component is always providing the things we're depending on.

Defaults are super helpful for scenarios where you want to make the component more flexible to handle cases where those things aren't provided. This shows up a lot in unit tests specifically where we usually shallow mount things independently from the parent component that would usually provide these things 😅

dobromir-hristov (approved):
lgtm :)

mportiz08 (approved):

hqhhuang:
@swift-ci test

PR #601, opened 2023-04-20
Set tabnav to the tab with latest title update

hqhhuang:
Bug/issue #, if applicable: 108013117

## Summary
We should set the current tab of TabNav to the tab that just got added or just had its title updated.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode/TabNavigator.vue
+++ b/src/components/ContentNode/TabNavigator.vue
@@ -75,8 +75,7 @@ export default {
     titles(newVal, oldVal) {
       if (newVal.length < oldVal.length) {
         // set to the first tab if selected tab was removed
-        const removedTab = oldVal.find((tab, i) => newVal[i] !== tab);
-        if (removedTab === this.currentTitle) {
+        if (!newVal.includes(this.currentTitle)) {
           // eslint-disable-next-line prefer-destructuring
           this.currentTitle = newVal[0];
         }
--- a/src/components/ContentNode/TabNavigator.vue
+++ b/src/components/ContentNode/TabNavigator.vue
@@ -71,6 +71,28 @@ export default {
       currentTitle: this.titles[0],
     };
   },
+  watch: {
+    titles(newVal, oldVal) {
+      if (newVal.length === oldVal.length) {
+        // set to the tab with updated title
+        newVal.forEach((tab, index) => {
+          if (tab !== oldVal[index]) {
+            this.currentTitle = tab;
+          }
+        });
+      } else if (newVal.length > oldVal.length) {
+        // set to newly added tab, may not be the last tab
+        this.currentTitle = newVal.find(tab => !oldVal.includes(tab));
+      } else {
+        // set to the tab before the removed one
+        const removedTab = oldVal.find(tab => !newVal.includes(tab));
+        const removedTabIdx = oldVal.indexOf(removedTab);
+        const newTabIdx = (removedTabIdx === oldVal.length - 1)
+          ? (newVal.length - 1) : removedTabIdx;
+        this.currentTitle = newVal[newTabIdx];
+      }
+    },
+  },
 };
 </script>
 
--- a/src/components/ContentNode/TabNavigator.vue
+++ b/src/components/ContentNode/TabNavigator.vue
@@ -73,23 +73,17 @@ export default {
   },
   watch: {
     titles(newVal, oldVal) {
-      if (newVal.length === oldVal.length) {
-        // set to the tab with updated title
-        newVal.forEach((tab, index) => {
-          if (tab !== oldVal[index]) {
-            this.currentTitle = tab;
-          }
-        });
-      } else if (newVal.length > oldVal.length) {
-        // set to newly added tab, may not be the last tab
-        this.currentTitle = newVal.find(tab => !oldVal.includes(tab));
-      } else {
-        // set to the tab before the removed one
+      if (newVal.length < oldVal.length) {
+        // set to the first tab if selected tab was removed
         const removedTab = oldVal.find(tab => !newVal.includes(tab));
-        const removedTabIdx = oldVal.indexOf(removedTab);
-        const newTabIdx = (removedTabIdx === oldVal.length - 1)
-          ? (newVal.length - 1) : removedTabIdx;
-        this.currentTitle = newVal[newTabIdx];
+        if (removedTab === this.currentTitle) {
+          // eslint-disable-next-line prefer-destructuring
+          this.currentTitle = newVal[0];
+        }
+      } else {
+        // set to newly added/changed tab
+        const newTab = newVal.find(tab => !oldVal.includes(tab));
+        this.currentTitle = newTab || this.currentTitle;
       }
     },
   },
--- a/src/components/ContentNode/TabNavigator.vue
+++ b/src/components/ContentNode/TabNavigator.vue
@@ -76,8 +76,8 @@ export default {
       if (newVal.length < oldVal.length) {
         // set to the first tab if selected tab was removed
         if (!newVal.includes(this.currentTitle)) {
-          // eslint-disable-next-line prefer-destructuring
-          this.currentTitle = newVal[0];
+          const [firstTitle] = newVal;
+          this.currentTitle = firstTitle;
         }
       } else {
         // set to newly added/changed tab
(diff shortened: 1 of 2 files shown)


hqhhuang:
@swift-ci

hqhhuang:
@swift-ci test

tests/unit/components/ContentNode/TabNavigator.spec.js
```diff
@@ -69,4 +69,49 @@ describe('TabNavigator.spec', () => {
...
   });
+
+  it('selects correct tab when adding a tab', async () => {
+    const longerTitles = ['Long tab title',
+      'A Longer tab title', 'The Longest tab title', 'added title'];
+
+    const wrapper = createWrapper();
+    wrapper.setProps({ titles: longerTitles });
+
+    expect(wrapper.vm.currentTitle).toEqual('added title');
```
mportiz08, line 80: Can we check that the right tab gets rendered as the current one instead of doing assertions on the actual view model instance?

tests/unit/components/ContentNode/TabNavigator.spec.js
```diff
@@ -69,4 +69,49 @@ describe('TabNavigator.spec', () => {
     expect(wrapper.find('.tabs-content').text()).toBe('Second');
     expect(tabnav.props('value')).toEqual(titles[1]);
   });
+
+  it('selects correct tab when adding a tab', async () => {
```
mportiz08, line 73: Do these need to be asynchronous?

tests/unit/components/ContentNode/TabNavigator.spec.js
```diff
@@ -69,4 +69,49 @@ describe('TabNavigator.spec', () => {
...
+    const longerTitles = ['Long tab title',
```
mportiz08, line 74: Something like this might help save some typing if you need to do similar things in the future:

`const longerTitles = titles.concat('added title');`

So you don't have to copy the full array in each test.

tests/unit/components/ContentNode/TabNavigator.spec.js
```diff
@@ -69,4 +69,49 @@ describe('TabNavigator.spec', () => {
...
+  it('selects first tab when deleting current tab', async () => {
+    const longerTitles = ['Long tab title',
+      'A Longer tab title', 'The Longest tab title', 'added title'];
+
+    const wrapper = createWrapper();
+    // current tab is on 'added title'
+    wrapper.setProps({ titles: longerTitles });
+    wrapper.setProps({ titles });
+
+    expect(wrapper.vm.currentTitle).toEqual('Long tab title');
```
mportiz08, line 92: Shouldn't it change the current one to the one right before the one you deleted? Or am I misunderstanding?

tests/unit/components/ContentNode/TabNavigator.spec.js
```diff
@@ -69,4 +69,49 @@ describe('TabNavigator.spec', () => {
...
+  });
+
+  it('selects correct tab when changing a tab', async () => {
+    const changedTitle = ['Long tab title',
+      'A Longer tab title', 'changed tab title'];
+
+    const wrapper = createWrapper();
+    wrapper.setProps({ titles: changedTitle });
+
+    expect(wrapper.vm.currentTitle).toEqual('changed tab title');
```
mportiz08, line 115: You may want to test at least one other index (in the beginning or middle or both) to make sure your code is not just accidentally getting it right.

src/components/ContentNode/TabNavigator.vue
```diff
@@ -71,6 +71,22 @@ export default {
       currentTitle: this.titles[0],
     };
   },
+  watch: {
+    titles(newVal, oldVal) {
+      if (newVal.length < oldVal.length) {
+        // set to the first tab if selected tab was removed
+        const removedTab = oldVal.find(tab => !newVal.includes(tab));
```
mportiz08, line 78: The `.includes` part implies that there is a nested loop, which isn't necessary. To do the same check without the nested loop, I think you could do something like this:

```javascript
const removedTab = oldVal.find((tab, i) => newVal[i] !== tab));
```

src/components/ContentNode/TabNavigator.vue
```diff
@@ -71,6 +71,22 @@ export default {
...
+        if (removedTab === this.currentTitle) {
+          // eslint-disable-next-line prefer-destructuring
+          this.currentTitle = newVal[0];
```
mportiz08, line 81: Do we want to reset it to the first one or would it make more sense to select the one immediately before/after the one that was removed?

hqhhuang: We are keeping whatever tab is selected at the time of deletion. If the deleted tab is the selected tab, we just set it to the first item.

tests/unit/components/ContentNode/TabNavigator.spec.js
hqhhuang: I think we already have a test that tests whether the content changes based on which tab is picked. Do you think it's still necessary to test whether the v-model instance works in getting the correct tab rendered?

src/components/ContentNode/TabNavigator.vue
```diff
@@ -71,6 +71,22 @@ export default {
...
+        }
+      } else {
+        // set to newly added/changed tab
+        const newTab = newVal.find(tab => !oldVal.includes(tab));
+        this.currentTitle = newTab || this.currentTitle;
+      }
+    },
+  },
```
dobromir-hristov, line 89: Can this logic be simplified to something like:

1. If no `currentTitle` or the `currentTitle` is not part of the existing `titles`, just set the first one.
2. If `currentTitle` is in the `titles`, just do nothing.

src/components/ContentNode/TabNavigator.vue
```diff
@@ -71,6 +71,21 @@ export default {
       currentTitle: this.titles[0],
     };
   },
+  watch: {
+    titles(newVal, oldVal) {
+      if (newVal.length < oldVal.length) {
+        // set to the first tab if selected tab was removed
+        if (!newVal.includes(this.currentTitle)) {
+          // eslint-disable-next-line prefer-destructuring
+          this.currentTitle = newVal[0];
```
marinaaisa, line 80: Instead of disabling ESLint, you could:

```suggestion
          const [firstTitle] = newVal;
          this.currentTitle = firstTitle;
```

src/components/ContentNode/TabNavigator.vue
```diff
@@ -71,6 +71,21 @@ export default {
       currentTitle: this.titles[0],
     };
   },
+  watch: {
+    titles(newVal, oldVal) {
+      if (newVal.length < oldVal.length) {
```
marinaaisa, line 76: To not add nested ifs, you could consider exit if newVal includes current title and continue with the case of a newVal being greater or equal than the oldVal

```js
// if newVal includes current title exit
if (newVal.includes(this.currentTitle)) return;
// if newVal is greater than or equal than the oldVal
// find its new tab
if (newVal.length >= oldVal.length) {
  const newTab = newVal.find(tab => !oldVal.includes(tab));
  this.currentTitle = newTab || this.currentTitle;
  return;
}
// otherwise set first title as current title
const [firstTitle] = newVal;
this.currentTitle = firstTitle;
```

marinaaisa (commented):
Looks good! I added minor comments 👏  similar to the ones Dobri did

src/components/ContentNode/TabNavigator.vue
dobromir-hristov: Actually thats not how it should work. If a new items is added, it should be highlighted.

marinaaisa: Okay, so the checking if the newVal is greater or equal than the oldVal should come up first, right?

```js
// if newVal is greater than or equal than the oldVal
// find its new tab
if (newVal.length >= oldVal.length) {
  const newTab = newVal.find(tab => !oldVal.includes(tab));
  this.currentTitle = newTab || this.currentTitle;
  return;
}
// if newVal includes current title exit
if (newVal.includes(this.currentTitle)) return;
// otherwise set first title as current title
const [firstTitle] = newVal;
this.currentTitle = firstTitle;
```

marinaaisa (approved):

mportiz08 (approved):

hqhhuang:
@swift-ci test

PR #613, opened 2023-04-25
Change the tab-nav rendering to be v-show based

dobromir-hristov:
Bug/issue #, if applicable: 108379797

## Summary

Changes the TabNav content rendering to be v-show based, so images can be pre-loaded and wont flash when changing tabs.

## Dependencies

NA

## Testing

1. Load an archive with a tabnav that has mostly images
2. Assert images no longer flash when changing active tab
 
## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode/TabNavigator.vue
+++ b/src/components/ContentNode/TabNavigator.vue
@@ -22,7 +22,12 @@
       <div class="tabs-content-container">
         <transition-group name="fade">
           <template v-for="title in titles">
-            <div :key="title" v-show="title === currentTitle">
+            <div
+              v-show="title === currentTitle"
+              :key="title"
+              :class="{ active: title === currentTitle }"
+              class="tab-container"
+            >
               <slot :name="title" />
             </div>
           </template>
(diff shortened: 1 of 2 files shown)


mportiz08:
Not opposed to switching to a `v-show` based approach as long as we've done some performance testing of realistic scenarios to make sure that there's no major regression with loading everything up-front.

dobromir-hristov:
Played a bit with tabnavs with lots of images found a few things. Behaviour is consistent across browsers, which is great. 

We explicitly set that images under the tabnav component should be eager loaded, so it downloads all images, rendered or not. This can deff cause slowdown on slower/limited networks. 
Before this PR, it was not an issue, because the images are not rendered at all, compared to being hidden now.

If we however change the `imageLoadingStrategy` that back to `lazy`, browsers only load the images you can see. As you change the active tab, it starts downloading the ones on the new active tab.

dobromir-hristov:
@swift-ci test

hqhhuang (approved):
🚢

dobromir-hristov:
@swift-ci test

hqhhuang:
@swift-ci test

PR #618, opened 2023-04-26
Improve swift attribute formatting

mportiz08:
Bug/issue #, if applicable: 108412440

## Summary

Updates the formatting logic for Swift declarations to break each [attribute][1] onto its own line.

Example:
```swift
// before
@discardableResult @objc(baz) func foobarbaz() -> Int

// after
@discardableResult
@objc(baz)
func foobarbaz() -> Int
```

\cc @ethan-kusters 

## Testing

Steps:
1. Configure [swift-docc-plugin][2] for a new or existing swift package that has examples of Swift symbols with attributes
2. Build this branch and note the location of `dist`
3. Run `DOCC_HTML_DIR=/path/to/dist swift package --disable-sandbox preview-documentation --target [target]` to preview it and verify that the symbols now have each attribute on its own line.

Here is an example package (target `AttributeExamples`) for convenience:
[AttributeExamples.zip](https://github.com/apple/swift-docc-render/files/11338015/AttributeExamples.zip)

## Checklist

- [X] Added tests

[1]: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes
[2]: https://github.com/apple/swift-docc-plugin

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -127,6 +127,18 @@ export default {
           }
         }
 
+        // check if this is a text token following an attribute token
+        // so we can insert a newline here and split each attribute onto its
+        // own line
+        //
+        // we want to avoid doing this when the attribute is encountered
+        // in a param clause for attributes like `@escaping`
+        if (token.kind === TokenKind.text && i > 0
+          && tokens[i - 1].kind === TokenKind.attribute
+          && numUnclosedParens === 0) {
+          newToken.text = `${token.text.trimEnd()}\n`;
+        }
+
         // if we find some text ending with ", " and the next token is the start
         // of a new param, update this token text to replace the space with a
         // newline followed by 4 spaces
--- a/tests/unit/components/DocumentationTopic/PrimaryContent/DeclarationSource.spec.js
+++ b/tests/unit/components/DocumentationTopic/PrimaryContent/DeclarationSource.spec.js
@@ -741,4 +741,52 @@ describe('Swift function/initializer formatting', () => {
     expect(tokenComponents.at(1).props('text')).toBe('\n');
     expect(tokenComponents.at(3).props('text')).toBe('(baz)\n');
   });
+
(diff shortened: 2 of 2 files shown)


ethan-kusters:
> ```swift
> // before
> @discardableResult @objc(baz) func foobarbaz() -> Int
> 
> // after
> @discardableResult
> @objc(baz)
> func foobarbaz() -> Int
> ```

Did you consider putting all attributes on a single line instead of an additional line per attribute? Most attributes are pretty short so I think that _may_ be a better balance here. I don't have a very strong opinion either way though – we can always make adjustments later after we see how things look in-practice.

```swift
// before
@discardableResult @objc(baz) func foobarbaz() -> Int

// after
@discardableResult @objc(baz)
func foobarbaz() -> Int
```

ethan-kusters:
> I didn't give it much consideration. From looking at code examples, I didn't see many scenarios with more than one attribute, and it seemed that the common pattern was splitting the attributes onto their own lines or having everything inline.
> 
> It would be a more involved implementation, but I could look into that if we wanted to pursue the design direction of having all attributes on one line.

I think you're right that the vast-majority of current use-cases involve a single attribute. If this is the simpler implementation I think it makes sense to start with this and we can adjust in the future if there's a clear need.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -127,6 +127,18 @@ export default {
           }
         }
 
+        // check if this is a text token following an attribute token
+        // so we can insert a newline here and split each attribute onto its
+        // own line
+        //
+        // we want to avoid doing this when the attribute is encountered
+        // in a param clause for attributes like `@escaping`
```
hqhhuang, line 135: Looks great! Can we add a test for this scenario?

mportiz08: 0cc2a2476858c45cd0c376444cc12c8edbdeaf6c

hqhhuang (approved):
LGTM!

mportiz08:
@swift-ci test

PR #633, opened 2023-05-02
Render all attributes for Swift symbol declarations on a single line

mportiz08:
Bug/issue #, if applicable: 108798969

## Summary

This changes the behavior introduced with #618 to render all attributes on a single line as opposed to each one getting its own line.

Example:

```swift
// before
@discardableResult @objc(baz) func foobarbaz() -> Int

// after
@discardableResult @objc(baz)
func foobarbaz() -> Int
```

## Testing

Same testing instructions as #618 but verify that attributes are now all on one line as opposed to each getting their own line.

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -127,15 +127,14 @@ export default {
           }
         }
 
-        // check if this is a text token following an attribute token
-        // so we can insert a newline here and split each attribute onto its
-        // own line
-        //
-        // we want to avoid doing this when the attribute is encountered
-        // in a param clause for attributes like `@escaping`
-        if (token.kind === TokenKind.text && i > 0
-          && tokens[i - 1].kind === TokenKind.attribute
-          && numUnclosedParens === 0) {
+        // Find the text following the last attribute preceding the start of a
+        // declaration by determining if this is the text token in between an
+        // attribute and a keyword outside of any parameter clause. A newline
+        // will be added to break these attributes onto their own single line.
+        const prevToken = i > 0 ? tokens[i - 1] : undefined;
+        if (token.kind === TokenKind.text && numUnclosedParens === 0
+          && prevToken && prevToken.kind === TokenKind.attribute
+          && nextToken && nextToken.kind === TokenKind.keyword) {
           newToken.text = `${token.text.trimEnd()}\n`;
         }
 
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -84,7 +84,6 @@ export default {
       let indentedParams = false;
       const newTokens = [];
       let i = 0;
-      let j = 1;
       let openParenTokenIndex = null;
       let openParenCharIndex = null;
       let closeParenTokenIndex = null;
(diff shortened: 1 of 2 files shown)


src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -127,15 +127,14 @@ export default {
...
-        // in a param clause for attributes like `@escaping`
-        if (token.kind === TokenKind.text && i > 0
-          && tokens[i - 1].kind === TokenKind.attribute
-          && numUnclosedParens === 0) {
+        // Find the text following the last attribute preceding the start of a
+        // declaration by determining if this is the text token in between an
+        // attribute and a keyword outside of any parameter clause. A newline
+        // will be added to break these attributes onto their own single line.
+        const prevToken = i > 0 ? tokens[i - 1] : undefined;
+        if (token.kind === TokenKind.text && numUnclosedParens === 0
```
hqhhuang, line 135: `numUnclosedParens === 0` 
I think we need another way to check this. From issue 08798969, we have examples like `@ViewBuilder` where the attribute is inside of an unclosed parenthesis, but is expected to be displayed on a new line.

ethan-kusters: This looks right to me actually – I think what you're referencing is a "Type Attribute" (one attached to a parameter type in the middle of a declaration) what we're trying to do here is move all "Declaration Attributes" on to their own line (those attached to the front of a declaration).

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes

hqhhuang: Got it! Marcus opened a separate [PR](https://github.com/mportiz08/swift-docc-render/pull/1/files) for that

mportiz08: Yeah. The `numUnclosedPararms === 0` check is specifically for a scenario I've seen where the `@escaping` attribute is used within the definition of a type where we want to avoid disturbing that. It's a makeshift way to make sure that I'm only touching the attributes before the actual parameter clause (inside the `()`).

And yes, there is separate logic for the `@resultBuilder` situation that shouldn't impact this.

mportiz08: To be more clear, this logic is only for attributes before the actual parameters that prefix the function/macro/initializer/etc as opposed to any attributes that might be a part of the parameters themselves.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -127,15 +127,14 @@ export default {
...
-        // we want to avoid doing this when the attribute is encountered
-        // in a param clause for attributes like `@escaping`
-        if (token.kind === TokenKind.text && i > 0
-          && tokens[i - 1].kind === TokenKind.attribute
-          && numUnclosedParens === 0) {
+        // Find the text following the last attribute preceding the start of a
+        // declaration by determining if this is the text token in between an
+        // attribute and a keyword outside of any parameter clause. A newline
+        // will be added to break these attributes onto their own single line.
+        const prevToken = i > 0 ? tokens[i - 1] : undefined;
```
dobromir-hristov, line 134: Accessing `-1` in an array returns `undefined`, so you should be good to simplify I think.

```suggestion
        const prevToken = tokens[i - 1];
```

dobromir-hristov (approved):
Changes look good, left minor suggestion.

Assertions are MUCH better. Thank you 🙇

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
mportiz08: That's dangerous but convenient 😁 thanks

32110898d9175017f0c9e58ace7b6fc118ccef26

mportiz08:
@swift-ci test

PR #639, opened 2023-05-05
Ensure the navigator filter picked tags are slightly visible, even on very narrow screens.

### LINKED ISSUE #603: UI bug when select "Article" in filter field
### Description

The article tag of the filter is partially blocked by the filter icon.

### Checklist

- [x] If possible, I've reproduced the issue using the `main` branch of this package.
image

- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc-render/issues).

### Expected Behavior

image

### Actual behavior

image

### Steps To Reproduce

1. Go to https://www.swift.org/documentation/docc/
2. Click the bottom left filter field
3. Choose `Article`

> Tips:
> If you can't reproduce it, try to use the slider to reduce the width of the `BaseNavigatorCard`

_No response_

dobromir-hristov:
Bug/issue #, if applicable: 108916627

## Summary

Adjusts the minimum filter width in the Navigator, so you can always see parts of the picked tags, on empty input.

closes #603 

## Dependencies

NA

## Testing

1. Make screen very narrow and then make the Navigator as narrow as possible.
2. Then try to pick a tag from the list
3. Assert at least some part of the tag is visible

## Checklist

- [ ] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -1144,8 +1144,14 @@ $filter-height-small: 60px;
     --input-border-color: var(--color-grid);
     --input-text: var(--color-figure-gray-secondary);
 
-    /deep/ .filter__input {
-      @include font-styles(body);
+    /deep/ {
+      .filter__input {
+        @include font-styles(body);
+
+        &-label::after {
+          min-width: 70px;
+        }
+      }
     }
   }
 }
--- a/src/components/Navigator/QuickNavigationModal.vue
+++ b/src/components/Navigator/QuickNavigationModal.vue
@@ -145,7 +145,7 @@ import keyboardNavigation from 'docc-render/mixins/keyboardNavigation';
 import LRUMap from 'docc-render/utils/lru-map';
 import { convertChildrenArrayToObject, getParents } from 'docc-render/utils/navigatorData';
 import { fetchDataForPreview } from 'docc-render/utils/data';
-import { SCROLL_LOCK_DISABLE_ATTR } from '@/utils/scroll-lock';
+import { SCROLL_LOCK_DISABLE_ATTR } from 'docc-render/utils/scroll-lock';
 
 const { PreviewState } = QuickNavigationPreview.constants;
 


Kyle-Ye:
Does this close #603?

dobromir-hristov:
Should yes. What do you think? is it better?

Kyle-Ye:
> Should yes. What do you think? is it better?

Can you add “Close #603” in your reply, append in the first line or manually add #603 to Development section to better fit in the Github workflow? Thanks

IMG_0142

mportiz08:
Hm, this didn't seem to resolve the issue for me in Chrome.

dobromir-hristov:
Here is how it behaves on my end in Chrome

https://user-images.githubusercontent.com/9863944/236750458-358e98e0-2974-4a09-9381-d6be07deccd2.mov

hqhhuang:
I think the issue is we don't just want to see part of the tag, we want to see the entire tag without scrolling if there is enough space in the filter bar

Kyle-Ye:
> I think the issue is we don't just want to see part of the tag, we want to see the entire tag without scrolling if there is enough space in the filter bar

+1. Yes that's exactly what I'd like to say. Maybe we can align the Chrome behavior with Safari and Firefox.

dobromir-hristov:
From what I got, the issue is this happens when there is not enough space. If we dont scroll the input when you pick a tag, you basically will just see the tag and you have to scroll the input because you wont be able to click into it to search.

Kyle-Ye:
> From what I got, the issue is this happens when there is not enough space. If we dont scroll the input when you pick a tag, you basically will just see the tag and you have to scroll the input because you wont be able to click into it to search.

No. I disagree with it. Check the screen record below. If it was the issue that there is not enough space, the 2 behavior should always be the same.

Chrome:
1. Narrow sidebar -> add tag -> show part of the tag.
2. Narrow sidebar -> add tag -> Expand the sidebar and narrow it back to the original size -> show full tag

Safari and Firefox:
1. Narrow sidebar -> add tag -> show full tag.
2. Narrow sidebar -> add tag -> Expand the sidebar and narrow it back to the original size -> show full tag

https://user-images.githubusercontent.com/43724855/236894296-e6227f63-874a-40e5-976f-3ae7b03c4729.mov

dobromir-hristov:
@Kyle-Ye I saw what you mean, you would like the component to not be scrolled and reveal more of the input area, but rather keep the selected tag visible, even if that means you need to scroll a bit to see your input.

Ofc on a very narrow size this is more of a problem imho, but I guess as long as you have the input focused you can just start typing...

https://user-images.githubusercontent.com/9863944/237017435-5b9a4b0e-6292-4dc0-b035-fb568d349352.mov

Kyle-Ye:
> This PR reduces that space a bit, so its never more than 100% of the available width, aka there is always a bit of tag poking out 😆
> 
> Currently I think we have a hard min of 130px.

Tried the PR on slothcreator json which will have a "Hide Deprecated" label.

Even we have enough space, it will still truncate the label.

image

dobromir-hristov:
I will take another look :)

dobromir-hristov:
@Kyle-Ye what about now? Had to force push, because gh gave me some weird errors.

Just set a smaller hard limit, which I think is better than the constant 80%

Kyle-Ye:
> @Kyle-Ye what about now? Had to force push, because gh gave me some weird errors.
> 
> Just set a smaller hard limit, which I think is better than the constant 80%

Much better now to me. Thanks

mportiz08 (approved):

Issue #632, opened 2023-04-25
Symbol links with custom authored titles are rendered in code voice


ethan-kusters:
### Description

Symbol links with custom authored titles are rendered in code voice

If I write a doc link with a title like

```
Check out [this great function](doc:Sloth/eat(_:quantity:)).
```

the rendered output is

```
Check out `this great function`.
```

With `this great function` in code voice. The custom title should respect the formatting that was provided in the link title.

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc/issues).

### Swift-DocC Version Information

88d0fd6676f2a2ee2d169d813420645cf6bfd52d

mportiz08:
Sounds reasonable, although I worry a bit about there being an expectation that a link to a symbol would be in code voice by default. I think this is probably the right way to go though since it allows for that flexibility.

I didn't even realize you could provide a custom title for a DocC symbol link now 😮

Kyle-Ye:
> Sounds reasonable, although I worry a bit about there being an expectation that a link to a symbol would be in code voice by default. I think this is probably the right way to go though since it allows for that flexibility.
> 
> I didn't even realize you could provide a custom title for a DocC symbol link now 😮

This feature is tracked as a bug via https://github.com/apple/swift-docc/issues/271 and implemented by https://github.com/apple/swift-docc/pull/376 @mportiz08

marinaaisa:
Hey @ethan-kusters @Kyle-Ye @mportiz08 I opened a PR for this issue: https://github.com/apple/swift-docc-render/pull/645
Symbol links now respect the formatting provided in `overridingTitleInlineContent` but this means that if overridingTitleInlineContent's type is other than `symbol` or `codeVoice` it won't be using CodeVoice component at all.

Is this expected as it is?

PR #645, opened 2023-05-08
Symbol links should respect the formatting provided in `overridingTitleInlineContent`

marinaaisa:
Bug/issue #108515663, if applicable: 

## Summary

Symbol links should respect the formatting provided in `overridingTitleInlineContent`

- resolves https://github.com/apple/swift-docc-render/issues/632

### Before
Screenshot 2023-05-08 at 19 32 41

### After

Screenshot 2023-05-08 at 19 32 26

## Dependencies

NA

## Testing

[eat(_:quantity:).json.zip](https://github.com/apple/swift-docc-render/files/11423647/eat._.quantity.json.zip)

Steps:
1. Use json attached in your doccarchive and run `VUE_APP_DEV_SERVER_PROXY=[path to doccarchive] npm run serve`
2. Go to `http://localhost:8080/documentation/slothcreator/Sloth/eat(_:quantity:)`
3. Assert that "energy" inside Discussion section does not use the CodeVoice component

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode/Reference.vue
+++ b/src/components/ContentNode/Reference.vue
@@ -43,7 +43,7 @@ export default {
       return name !== notFoundRouteName;
     },
     isSymbolReference() {
-      return this.kind === 'symbol'
+      return (this.kind === 'symbol' || this.kind === 'codeVoice')
         && (this.role === TopicRole.symbol || this.role === TopicRole.dictionarySymbol);
     },
     isDisplaySymbol({ isSymbolReference, titleStyle, ideTitle }) {
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -430,10 +430,12 @@ function renderNode(createElement, references) {
       const titleInlineContent = node.overridingTitleInlineContent
         || reference.titleInlineContent;
       const titlePlainText = node.overridingTitle || reference.title;
+      const kind = node.overridingTitleInlineContent
+        ? node.overridingTitleInlineContent[0].type : reference.kind;
       return createElement(Reference, {
         props: {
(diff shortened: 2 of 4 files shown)


src/components/ContentNode/Reference.vue
```diff
@@ -43,7 +43,7 @@ export default {
       return name !== notFoundRouteName;
     },
     isSymbolReference() {
-      return this.kind === 'symbol'
+      return (this.kind === 'symbol' || this.kind === 'codeVoice')
```
mportiz08, line 46: I'm not sure that this is the right approach since it mixes the already well defined values that `kind` can be with our own custom values like this.

In this specific situation, you'll actually end up with a doubly nested `<code>` element, which is not ideal, although it may look right in your example. One comes from the reference component that this logic determines and another would would come through the content itself.

I think what we actually want is just to short-circuit this logic in the case that we have a reference with an `overridingTitleInlineContent` value, which is really close to the approach you have here, but not quite the same. Basically, in the case where a developer has provided custom text, we want to defer to their inline formatting and not apply any extra.

mportiz08 (changes requested):

src/components/ContentNode/Reference.vue
mportiz08: I went ahead and opened an alternative PR based on this feedback if you'd like to review that: #647

dobromir-hristov (commented):
Looked at both this and #647, and I think that one has a better approach.

marinaaisa:
Closing this PR in favor of https://github.com/apple/swift-docc-render/pull/647

PR #646, opened 2023-05-08
Wrap Declarations in Minimized and IDE mode

hqhhuang:
Bug/issue #, if applicable: rdar://106567892

## Summary
Wrap Declarations in minimized and IDE mode.

## Testing
Steps:
1. Manual testing

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -149,6 +149,7 @@
 import Language from 'docc-render/constants/Language';
 import metadata from 'theme/mixins/metadata.js';
 import { buildUrl } from 'docc-render/utils/url-helper';
+import { normalizeRelativePath } from 'docc-render/utils/assets';
 
 import Aside from 'docc-render/components/ContentNode/Aside.vue';
 import BetaLegalText from 'theme/components/DocumentationTopic/BetaLegalText.vue';
@@ -382,14 +383,13 @@ export default {
     };
   },
   computed: {
-    normalizedSwiftPath: ({ normalizePath, swiftPath }) => (normalizePath(swiftPath)),
+    normalizedSwiftPath: ({ swiftPath }) => (normalizeRelativePath(swiftPath)),
     normalizedObjcPath: ({
-      normalizePath,
       objcPath,
       swiftPath,
     }) => (
       // do not append language query parameter if no swiftPath exists
-      normalizePath((objcPath && swiftPath) ? buildUrl(objcPath, {
+      normalizeRelativePath((objcPath && swiftPath) ? buildUrl(objcPath, {
         language: Language.objectiveC.key.url,
       }) : objcPath)
     ),
@@ -517,12 +517,6 @@ export default {
     },
   },
   methods: {
-    normalizePath(path) {
-      // Sometimes `paths` data from `variants` are prefixed with a leading
-      // slash and sometimes they aren't
-      if (!path) return path;
-      return path.startsWith('/') ? path : `/${path}`;
-    },
     extractProps(json) {
       const {
         abstract,
@@ -618,7 +612,7 @@ export default {
 
       this.$nextTick().then(() => {
         this.$router.replace({
-          path: this.normalizePath(this.objcPath),
+          path: normalizeRelativePath(this.objcPath),
           query: {
             ...query,
             language: Language.objectiveC.key.url,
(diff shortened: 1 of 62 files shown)


src/components/DocumentationTopic.vue
```diff
@@ -732,6 +732,12 @@ export default {
 
     .source {
       border-radius: var(--code-border-radius);
+
+      > code {
+        @include inTargetIde() {
+          white-space: pre-wrap;
```
mportiz08, line 738: I think I've already expressed my opinion on this before, but I generally don't think it's a great idea to try and mix our current hard wrapping approach with a soft wrapping approach for displaying symbol declarations.

Just as a quick example, here is a comparison of how an example Swift API with multiple params would look with the current approach vs the `pre-wrap` here:

(current)
Screenshot 2023-05-08 at 2 05 02 PM

(`white-space: pre-wrap`)
Screenshot 2023-05-08 at 2 04 54 PM

This is just my subjective opinion, but I think we should go with either hard-wrap or soft-wrap and not try to mix—and my preference would be to keep the hard-wrapping since it breaks up parameters in a clearer way.

dobromir-hristov: Unless this is a user specific option, aka per IDE, I dont like it either. If a user prefers wrapping, then we could add it for them, but to all by default I dont think its great.

Another thing, wrapping should probably keep the indentation it's line, which i am not sure how to do...

hqhhuang: FWIW, this PR only allows the declaration to wrap when it's in both IDE and minimized mode. To Ethan's point, this is a really specific use case where we don't want the users to have to use the mouse to see the full content of the declaration.

mportiz08: > > Another thing, wrapping should probably keep the indentation it's line, which i am not sure how to do...
> 
> Any ideas on how to achieve this? This is my main feedback as well. From discussions with Hanqing it seems like that would be a much larger effort.

Assuming that you mean a single line by "it's line", we would probably need to update the [logic](https://github.com/apple/swift-docc-render/blob/864361eb64e16de58e4257fe9b86fb8e70d96156/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue#L60) to _not_ use the formatted tokens and instead use the original tokens in this specific scenario—that way we would just be pivoting to soft wrapping and not mixing the soft wrapping with the hard wrapping we do today. I still think it would be a worse experience though, but that's just my subjective opinion.

dobromir-hristov (approved):
Since this will only wrap single lines in IDEs that use the minimised view, I dont see issues with the change.

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -173,6 +171,13 @@ export default {
 
       return newTokens;
     },
+    hasMultipleLines({ formattedSwiftTokens }) {
```
dobromir-hristov, line 174: What if this is an objc declaration? This will only detect swift tokens.

dobromir-hristov (commented):
Why the more complicated solution? Did the old multiline detection cause issues?

hqhhuang:
> Why the more complicated solution? Did the old multiline detection cause issues?
Yes, the old way only detects if the declaration is displayed on multiple lines.
For the wrapping logic, we only want it to wrap if the declaration is singled lined, not formatted with our smart wrapping logic by adding new lines. So I added a new logic for essentially detecting whether we are using smart wrapping. 

We still need the old method for deciding which border radius to use. I also fixed a bug with that in this PR. The old logic only computes once on page load, so if the window resizes, causing the code to display on a single line instead of multiple lines (vice versa), the border radius is not updated.

src/components/DocumentationTopic.vue
```diff
@@ -737,6 +737,16 @@ export default {
       border-radius: var(--code-border-radius);
     }
 
+    .source:not(.has-multiple-lines) > code {
```
dobromir-hristov, line 740: I would add a comment why we do this.

src/components/DocumentationTopic.vue
```diff
@@ -737,6 +737,16 @@ export default {
...
+      @include inTargetIde() {
+        white-space: pre-wrap;
+
+        .token-identifier {
```
dobromir-hristov, line 744: Why are we only breaking the `token-identifier`?

hqhhuang: The thought is that we don't want to break up `keywords`, or break at any character for `attribute` or `typeIdentifier` tokens when they already have word breaks in them.

hqhhuang:
@swift-ci test

hqhhuang:
@swift-ci test

hqhhuang:
@swift-ci test

PR #655, opened 2023-05-10
Remove accidental leading whitespace for some captions

mportiz08:
Bug/issue #, if applicable: 109060293

## Summary

Fixes an issue where captions without titles have an unnecessary leading whitespace character, which may slightly impact the alignment of the text.

## Testing

Steps:
1. Verify that captions without titles don't have leading whitespace anymore
2. Verify that captions with titles still have a whitespace between the caption title and content.

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/ContentNode/FigureCaption.vue
+++ b/src/components/ContentNode/FigureCaption.vue
@@ -10,7 +10,12 @@
 
 <template>
   <figcaption class="caption" :class="{ centered }">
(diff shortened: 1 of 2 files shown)


dobromir-hristov (approved):
We could also move the `nbsp` into the `<strong>`.
Ideally it would add the space only if a slot AND a title is present, but I dont think its worth the work.

mportiz08:
Makes sense, but I think the existence of the title should imply the existence of the slot content in the expected usage.

PR #656, opened 2023-05-10
Fix param formatting with complex attributes

mportiz08:
Bug/issue #, if applicable: 108167163

## Summary

Fixes an issue where attributes are being incorrectly indented.

This happens because the existing logic for indenting parameters simply looks for the first opening `(` character to indent the first param of a multi-param symbol—this doesn't work when a symbol is prefixed with an attribute that takes arguments, since it also uses parenthesis.

To fix it—the logic has been updated to look for the first `(` character _that follows the first keyword token_.

**Example:**

```swift
// original declaration
@objc(quxA:B:) func quxqux(a: Int, b: Int) -> Int

// formatted declaration before this fix
@objc(quxA:B:
)
    func quxqux(a: Int,
    b: Int) -> Int

// formatted declaration after this fix
@objc(quxA:B:)
func quxqux(
    a: Int,
    b: Int
) -> Int
```

**Notes:**
The diff might be a little hard to read, but I basically just added a variable to track the index of the first keyword token and then wrapped an existing for-loop in an if-check that makes sure that variable is not null.

## Testing

Steps:
1. Build this branch with `npm run build`
2. Download/unzip example Swift package [AttributeExamples.zip](https://github.com/apple/swift-docc-render/files/11456549/AttributeExamples.zip)
3. Run `env DOCC_HTML_DIR=/path/to/dist swift package --disable-sandbox preview-documentation --target AttributeExamples`
4. Open http://localhost:8080/documentation/attributeexamples/exampleusingcomplexattributes/quxqux(a:b:) and verify that the formatted declaration matches the expectation from the example above.
5. Open http://localhost:8080/documentation/attributeexamples/exampleusingcomplexattributes/init(a:b:) and verify that the formatted initializer looks right as well.
6. Check out other declaration content to verify that there are no regressions with the formatting.

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -89,7 +89,7 @@ export default {
       let closeParenTokenIndex = null;
       let closeParenCharIndex = null;
       let numUnclosedParens = 0;
-      let identifierTokenIndex = null;
+      let firstKeywordTokenIndex = null;
 
       // loop through every declaration token
       while (i < tokens.length) {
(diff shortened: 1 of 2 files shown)


mportiz08:
Lol @ the :b:

I think the link is still right

mportiz08:
Actually wait, I don't think this will work with initializers, since they don't have an identifier :(

Marking as draft until I figure out a more complete solution...

mportiz08:
Maybe checking for the first keyword instead of the identifier would work more completely for different kinds of symbols (functions/macros/initializers/subscripts/etc) 🤔

mportiz08:
Changed it to look for the first keyword instead of the first identifier, which should work better for various kinds of multi-param symbols (like functions/initializers/subscripts/macros/etc). I also updated the fixture to include an initializer example.

Ready to review now :)

dobromir-hristov (approved):
Looks good to me.

PR #658, opened 2023-05-11
Enable screen readers to skip 'main' in IDE mode

hqhhuang:
Bug/issue #, if applicable: rdar://107573378 

## Summary
In IDE mode, the content in the `<main>` tag is the only content visible. VO would read "you are currently on a main", which creates an extra layer for users to step into before reading the actual content of the page.
This PR allows screen readers to not announce a `main` in IDE mode.

## Testing
Steps:
1. Manual testing

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -13,7 +13,11 @@
     class="doc-topic"
     :class="{ 'with-on-this-page': enableOnThisPageNav && isOnThisPageNavVisible }"
   >
-    <main class="main" id="main" role="main" tabindex="0">
+    <component
+      :is="isTargetIDE ? 'div' : 'main'"
+      :role="isTargetIDE? 'none' : 'main'"
+      class="main" id="main" tabindex="0"
+    >
       <DocumentationHero
         :role="role"
         :enhanceBackground="enhanceBackground"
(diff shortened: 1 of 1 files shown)


dobromir-hristov (commented):
Change looks good, but I wonder if we:

1. Need the `role: main` at all
2. Need the `tabindex=0`

I checked GH, Vue docs and Stackoverflow, and none of them have such attributes on the `main` element.

marinaaisa:
> Change looks good, but I wonder if we:
> 
> 1. Need the `role: main` at all
> 2. Need the `tabindex=0`
> 
> I checked GH, Vue docs and Stackoverflow, and none of them have such attributes on the `main` element.

Good points, @dobromir-hristov 
We don't need `role: main` at all, because we already have the `main` tag itself, it's a duplication. Let's delete it.
About tabindex, I checked how https://webaim.org/techniques/skipnav/#maincontent do it and they don't use it when skipping the navigation so we should be fine deleting it too.

Thanks @hqhhuang ! :)

hqhhuang:
Thanks for the suggestions! I removed the `role: main`. 
However, during my testing I found that removing `tabindex=0` breaks the `skip-navigation` feature. 
@marinaaisa Could you test it as well?

marinaaisa:
> Thanks for the suggestions! I removed the role: main.
However, during my testing I found that removing tabindex=0 breaks the skip-navigation feature.
@marinaaisa Could you test it as well?

The skip navigation feature is fine without the tabindex=0 because it will focus in the next focusable item, skipping the whole navigation. We are good to delete it.

hqhhuang:
@swift-ci test

hqhhuang:
@swift-ci test

marinaaisa (approved):

hqhhuang:
@swift-ci test

PR #661, opened 2023-05-12
It prevents breaking words inside header cell in an HTML tables.

marinaaisa:
Bug/issue #107874086, if applicable: 

## Summary

It prevents breaking words inside header cell in an HTML tables.
In Japanese alphabet, words were being broken making reading difficult.

### Before
Screenshot 2023-05-12 at 19 23 43

### After
Screenshot 2023-05-12 at 19 23 30

## Dependencies

NA

## Testing

[SlothCreatori18n.doccarchive.zip](https://github.com/apple/swift-docc-render/files/11466531/SlothCreatori18n.doccarchive.zip)

Steps:
1. Use the doccarchive attached above
2. Go to http://localhost:8080/ja-JP/documentation/slothcreator
3. Assert that the table displayed at Overview in Japanese don't break their words 

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode/Table.vue
+++ b/src/components/ContentNode/Table.vue
@@ -44,6 +44,10 @@ table {
 /deep/ {
   th {
     font-weight: $font-weight-semibold;
+
+    > p {
+      word-break: keep-all;
+    }
   }
 
   td,
--- a/src/components/ContentNode/Table.vue
+++ b/src/components/ContentNode/Table.vue
@@ -45,6 +45,7 @@ table {
   th {
     font-weight: $font-weight-semibold;
     word-break: keep-all;
+    hyphens: auto;
   }
 
   td,
--- a/src/components/ContentNode/Table.vue
+++ b/src/components/ContentNode/Table.vue
@@ -44,10 +44,7 @@ table {
 /deep/ {
   th {
     font-weight: $font-weight-semibold;
-
-    > p {
-      word-break: keep-all;
-    }
+    word-break: keep-all;
   }
 
   td,


hqhhuang (approved):

dobromir-hristov:
I know this looks simple but for tables with very long titles on mobile devices, this may be a problem. @SamLanier what do you think?

SamLanier:
Yes, good point @dobromir-hristov. @marinaaisa Have you tested this on mobile (and with English content in all viewport sizes)?

marinaaisa:
> Yes, good point @dobromir-hristov. @marinaaisa Have you tested this on mobile (and with English content in all viewport sizes)?

Yes, I've been testing out different viewports and pages and I couldn't find regressions.

dobromir-hristov (approved):
Did some reviews and I think this is good to have. Maybe we can add `hyphens: auto;` as well, so the browser can split where it sees fit?

dobromir-hristov:
I think it will, if the word is long enough, container is short enough and the browser seers an opportunity based on the language to split it.

marinaaisa:
> I think it will, if the word is long enough, container is short enough and the browser seers an opportunity based on the language to split it.

I understand but we will be in the same problem of Japanese words being cut, I suggest keeping it like it is. At the same time, all tables have scroll in them when they are displayed in small viewport, so the container shouldn't be short enough either (I think)

dobromir-hristov:
Do japanese characters get cutoff by hyphens? I doubt that works like that 🤔

marinaaisa:
@dobromir-hristov I've been investigating a bit deeper and you are right about the hyphens:
Hyphenations don't work for CJK languages. Source: https://github.com/w3c/csswg-drafts/issues/785

I was also curious to know if hyphens would apply even when `word-break: keep-all;` is set, but it seems to work:


Example of:
```css
word-break: keep-all;
hyphens: auto;
```

Screenshot 2023-05-19 at 18 07 14

vs:
```css
word-break: keep-all;
hyphens: initial;
```
Screenshot 2023-05-19 at 18 07 21

Because the scroll in tables, I'm not sure when these hyphens will be applied in non-CJK languages but it's good to have them just in case. I added it. Thanks for the suggestion!

marinaaisa:
@swift-ci test

PR #622, opened 2023-04-27
Disable `skip-nav` VO link in IDE

hqhhuang:
Bug/issue #, if applicable: rdar://108634054

## Summary
Disable `skip-nav` VO link in IDE

Issue #449, opened 2022-10-05
Add a 'copy code to clipboard' button for code blocks


byaruhaf:
The swift book has a number of code blocks to complement the explanations.
To improve the experience for the reader we should add a “copy code” button to make it simple for them to copy and paste the code.

natikgadzhi:
Related: https://github.com/apple/swift-docc/issues/737

Both in code blocks — if someone wanted to make two improvements with one PR ;)

mportiz08:
Thanks for the suggestion @byaruhaf. I believe we are already tracking this enhancement request in #274

PR #502, opened 2022-12-21
Fix issues with new lines not rendering, when multiline blocks are used

dobromir-hristov:
Bug/issue #, if applicable: 103087953

## Summary

Fixes an issue where empty new lines are not rendered properly.

image

## Dependencies

NA

## Testing

[codesample.json.zip](https://github.com/apple/swift-docc-render/files/10277532/codesample.json.zip)

Using the example JSON above, try to copy the text into your editor.

1. Assert there is a single new line when you paste
2. Assert this works across browsers
3. Assert it does not regress

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #663, opened 2023-05-16
Fix extra empty line issue on tutorial

### LINKED ISSUE #662: Tutorial extra empty line bug
### Description

Extra line is render for code in tutorial.

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc-render/issues).

### Expected Behavior

Show the correct highlighting for Line 25-29

image

image

```
{
  "content": [...],
  "fileName": "CustomizedSlothView.swift",
  "fileType": "swift",
  "highlights": [
    {
      "line": 25
    },
    {
      "line": 26
    },
    {
      "line": 27
    },
    {
      "line": 28
    },
    {
      "line": 29
    }
  ],
  "identifier": "01-creating-code-02-08.swift",
  "syntax": "swift",
  "type": "file"
}
```

### Actual behavior

Show the wrong highlighting (Due to extra empty line on Line 4 and Line 5).

There should be only 1 empty line (Line 3). Currently there will be 3 empty line(Line 3 - Line 5)

image

### Steps To Reproduce

The tutorial resource can be downloaded from https://developer.apple.com/documentation/xcode/slothcreator_building_docc_documentation_in_xcode

_No response_

Kyle-Ye:
Bug/issue #, if applicable: 

Close #662

## Summary

See the full context on #622

Another way to fix the problem is changing the logic here `src/components/ContentNode/CodeListing.vue`

> Note:
> After this change when we try to copy the content of `CodeListing`, we will miss the "\n"

image

```
// Expected copy result
/*
S

A
*/

// Actual copy result
/*
S
A
*/
```

A later fix to implement a clipboard button may help lift the issue.

https://github.com/apple/swift-docc-render/issues/449

## Checklist

- [x] Added tests

### DIFF
--- a/src/utils/syntax-highlight.js
+++ b/src/utils/syntax-highlight.js
@@ -1,7 +1,7 @@
 /**
  * This source file is part of the Swift.org open source project
  *
- * Copyright (c) 2021 Apple Inc. and the Swift project authors
+ * Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
  * Licensed under Apache License v2.0 with Runtime Library Exception
  *
  * See https://swift.org/LICENSE.txt for license information
@@ -192,7 +192,7 @@ function duplicateMultilineNode(element) {
 
   // wrap each new line with the current element's class
   const result = getLines(element.innerHTML)
-    .reduce((all, lineText) => `${all}<span class="${className}">${lineText || '\n\n'}</span>\n`, '');
+    .reduce((all, lineText) => `${all}<span class="${className}">${lineText || ''}</span>\n`, '');
 
   // return a list of newly wrapped HTML elements
   return htmlToElements(result.trim());
--- a/tests/unit/utils/syntax-highlight.spec.js
+++ b/tests/unit/utils/syntax-highlight.spec.js
@@ -107,17 +107,11 @@ describe("syntax-highlight", () => {
     expect(sanitizedCode).toMatchInlineSnapshot(`
       <span class="syntax-keyword">let</span> multiline = <span class="syntax-string">""</span><span class="syntax-string">"</span>
       <span class="syntax-string">Needs</span>
-      <span class="syntax-string">
-
-      </span>
+      <span class="syntax-string"></span>
       <span class="syntax-string">Spaces</span>
-      <span class="syntax-string">
-
-      </span>
+      <span class="syntax-string"></span>
       <span class="syntax-string">Between</span>
-      <span class="syntax-string">
-
-      </span>
+      <span class="syntax-string"></span>
       <span class="syntax-string">Lines</span>
       <span class="syntax-string">"</span><span class="syntax-string">""</span>
     `);


Kyle-Ye:
@swift-ci please test

src/utils/syntax-highlight.js
```diff
@@ -192,7 +192,7 @@ function duplicateMultilineNode(element) {
 
   // wrap each new line with the current element's class
   const result = getLines(element.innerHTML)
-    .reduce((all, lineText) => `${all}<span class="${className}">${lineText || '\n\n'}</span>\n`, '');
+    .reduce((all, lineText) => `${all}<span class="${className}">${lineText || ''}</span>\n`, '');
```
mportiz08, line 195: Hm. This change likely fixes one bug but it unfortunately re-introduces another that we previously had. These newlines were added to fix a bug where empty lines in a multiline Swift string were accidentally getting stripped.

With this code listing as an example:
```swift
let multiline = """
Needs

Spaces

Between

Lines
"""
```

On main, this gets rendered how it looks there.

On this branch, this gets rendered as follows:
```swift
let multiline = """
Needs
Spaces
Between
Lines
"""
```

Ideally we could find a solution that fixes #662 without re-introducing this other bug, although I know it's tricky because of the complex logic we have to handle line numbers and syntax highlighting.

mportiz08:
\cc @dobromir-hristov

src/utils/syntax-highlight.js
mportiz08: I think #502 was the PR for the bug fix that this change might accidentally re-introduce for context.

Kyle-Ye: No. This PR does not re-introduces the bug you mention. It passed the existing test and I tested this form and confirm it locally.

> Because the original use nothing. #502 add `|| "\n\n"`. But actually add `|| ""` is enough.

dobromir-hristov: Unfortunately it does re-introduce the issue:

In your docs, of you have this:

```swift
let multiline = """
Needs

Spaces

Between

Lines
"""
```

It will generate this:

image

dobromir-hristov (commented):
We should probably re-visit the way we render line numbers and the css styling around that. 

We have a display:flex with a column wrap on the `<code>` element, which is probably not a good appraoch, since we are messing up the white space handling entirely, hence why I had to add that double `/n/n` while looping over the multi-lines. 

There are two good ways to do this from my research:
1. The traditional table rendering each line approach
2. Move line numbers to a list after the code and leave some space for them.

Each has it's PROs and CONs.

src/utils/syntax-highlight.js
Kyle-Ye: > Unfortunately it does re-introduce the issue:
> 
> In your docs, of you have this:
> 
> ```swift
> let multiline = """
> Needs
> 
> Spaces
> 
> Between
> 
> Lines
> """
> ```
> 
> It will generate this:
> 
> image

Got it. I thought it occur on the tutorial side.😂 And it was OK in tutorial steps.

Kyle-Ye:
Close. Prefer to use #667

PR #564, opened 2023-03-31
Improve AX for code blocks

marinaaisa:
Bug/issue #105394505, if applicable: 

## Summary

Improve AX for code blocks:

For screen readers: it adds text telling when a code block starts and ends
For keyboard navigation users: it adds tabindex to code blocks to be able to scroll horizontally when focusing on them

## Dependencies

NA

## Testing

Steps:
1. Go to a page containing a code block, like in tutorials or declarations
2. Turn on VoiceOver and assert that it reads when a code block starts and ends
3. Turn off VoiceOver
4. Start tabbing until focusing on a code block that has a long line. Assert that by clicking on key arrows you can move to the right and left

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #670, opened 2023-05-22
Fix regression with Objective-C api whitespace formatting

mportiz08:
Bug/issue #, if applicable: 109681482

## Summary

Fixes a regression that was introduced in #564 where multi-parameter Objective-C functions/initializers are not being formatted onto newlines anymore.

This happened because the code relies on a reference to a DOM node, and that PR introduced an additional layer of indirection where the reference is now a Vue component, requiring the need to also call `.$el` on that updated ref.

\cc @ethan-kusters 

## Testing

Steps:
1. Find any Objective-C symbol that has multiple parameters
2. Verify that the symbol is now being broken onto multiple lines instead of being on a single line

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
@@ -196,7 +196,7 @@ export default {
   async mounted() {
     if (this.language === Language.objectiveC.key.api) {
       await this.$nextTick();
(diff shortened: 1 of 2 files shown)


mportiz08:
@swift-ci test

dobromir-hristov (approved):
🤦

marinaaisa (approved):
Oops 🫣

mportiz08:
@swift-ci test

PR #677, opened 2023-05-25
Add support for localized font families

marinaaisa:
Bug/issue #108129110, if applicable: 

## Summary

Add support for localized font families

## Dependencies

NA

## Testing

Steps:
1. Make sure current fonts look correct 

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #682, opened 2023-05-31
Don't override monospaced font with localized body text font variant

mportiz08:
Bug/issue #, if applicable: 110021706

## Summary

This is a slight update to #677 to fix an issue where localized variants of normal body text fonts are mistakenly overriding the monospaced font variant, as we don't support localized monospaced variants (yet).

## Testing

Steps:
1. Configure a localized font for body text
2. View a page with both body text and code-voice text
3. Verify that only the body text gets the localized font and not the code-voice text

## Checklist

- [ ] ~~Added tests~~ (sass-only change, no unit infrastructure for sass code)

### DIFF
--- a/src/styles/core/typography/_font-style-utils.scss
+++ b/src/styles/core/typography/_font-style-utils.scss
@@ -57,18 +57,21 @@ $font-weight-bold: 700 !default;
       }
       $prevAttributes: $attributes;
     }
-  } @else {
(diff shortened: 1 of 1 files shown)


dobromir-hristov (approved):
Not the prettiest thing but I think we can go with it for now.

mportiz08:
@swift-ci test

PR #687, opened 2023-06-01
It allows set a list of available locales per page coming from the metadata object in the JSON file

marinaaisa:
Bug/issue #109166355, if applicable: 

## Summary

It allows set a list of available locales per page coming from the metadata object in the JSON file

## Dependencies

NA

## Testing

[SlothCreatori18n.doccarchive.zip](https://github.com/apple/swift-docc-render/files/11628329/SlothCreatori18n.doccarchive.zip)

Steps:
1. Add the following to `theme-settings.json` file at public folder
```json
{
  "features": {
    "docs": {
      "i18n": {
        "enable": true
      }
    }
  }
}
```
2. Run attached doccarchive
3. Go to http://localhost:8080/documentation/slothcreator
4. Assert that you can see on the bottom of the page the language selector with 2 languages on it
5. Go to any other page
6. Assert that you cannot see the language selector anymore

## Checklist

- [x] Added tests


src/App.vue
```diff
@@ -24,9 +24,18 @@
     <!-- The nav sticky anchor has to always be between the Header and the Content -->
     <div :id="baseNavStickyAnchorId" />
     <slot :isTargetIDE="isTargetIDE">
-      <router-view class="router-content" />
+      <router-view
+        @availableLocales="availableLocales = $event"
```
dobromir-hristov, line 28: Not a fan of router-views emitting events, as that is a very broad and generic component. 

Implementing a store so the Footer can have access to the available locales, just for this is probably too much work.

dobromir-hristov (commented):
Left a few small comments. Overall I think this should work.

src/App.vue
marinaaisa: First I started working implementing a store, I would set the availableLocales in the AppStore since preferredLocale is also there but this data is not the best place for the AppStore because it's a property depending on each page, not the user's preferences across the app...

That's why I change it to use emitting event + props

src/App.vue
```diff
@@ -101,6 +115,10 @@ export default {
         this.attachStylesToRoot(CSSCustomProperties);
       },
     },
+    $route() {
```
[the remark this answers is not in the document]
marinaaisa: I fixed it https://github.com/apple/swift-docc-render/pull/687/commits/3f92bcb8d93c8cf78c109988f9315b2065c0fea9

marinaaisa:
@swift-ci test

dobromir-hristov (approved):
@mportiz08 can you give a look as well?

marinaaisa:
@swift-ci test

src/App.vue
mportiz08: Hm. I don't really see a clear reason to handle this in a different way than we handle the references data, which is also data that changes from page to page. It's possible I'm missing something subtle that makes it different though.

src/views/NotFound.vue
```diff
@@ -20,5 +20,8 @@ import GenericError from 'theme/components/GenericError.vue';
 export default {
   name: 'NotFound',
   components: { GenericError },
+  created() {
+    this.$emit('availableLocales', []);
```
mportiz08, line 24: Hm. Wouldn't we want the full range of supported "app" locales for pages like this which have no user content but still can have localized app-produced error messages?

src/App.vue
marinaaisa: I refactored it to use the AppStore instead https://github.com/apple/swift-docc-render/pull/687/commits/bc8589bb5772a9bd6b381bb72c1830e886a14a44 let see what you think of this other option!

src/views/NotFound.vue
marinaaisa: True. I decided that these pages shouldn't change the previous availableLocales, so I deleted this change

marinaaisa:
@swift-ci test

src/stores/AppStore.js
```diff
@@ -43,6 +44,9 @@ export default {
     this.state.preferredColorScheme = value;
     Settings.preferredColorScheme = value;
   },
+  setAvailableLocales(locales) {
```
dobromir-hristov, line 47: maybe default to empty array? So we dont fallback all the time when we pass params?

```suggestion
  setAvailableLocales(locales = []) {
```

dobromir-hristov (commented):
Store implementation does look a bit cleaner. 

What would happen if you go another page type, that does not set the `availableLocales`? Would it use the old locales, from the previous page type?

dobromir-hristov:
Makes sense to me.

dobromir-hristov (approved):
LGTM. @mportiz08 check it again too.

mportiz08:
@swift-ci test

mportiz08 (approved):

PR #683, opened 2023-05-31
Update reference to renamed font style

mportiz08:
Bug/issue #, if applicable: 110085109

## Summary

Fixes a minor typography issue introduced with #40 where a font-style was renamed, but the place where it was used was not updated to use the new name.

## Testing

Steps:
1. Run `npm run serve` and verify that no sass warning messages are emitted
2. Find example content with a captioned figure and verify that it now has the smaller font-size that it did previously

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [ ] ~~Added tests~~ (CSS-only change)
- [X] Ran `npm test`, and it succeeded
- [X] Updated documentation if necessary

PR #40, opened 2022-01-03
[AX] <table> should be taken out of <figure> and the <figcaption> should become the table’s <caption>

marinaaisa:
Bug/issue #30234820, if applicable: 

## Summary

Tables were contained within a `<figure>` element. However, there is no need for this and it was making the tables much less accessible. Tables should stand on their own with the previous `<figcaption>` as the table’s `<caption>`. 

## Dependencies

NA

## Testing

Use the provided fixture folder:
[manual-fixtures.zip](https://github.com/apple/swift-docc-render/files/7803705/manual-fixtures.zip)

Steps:
1. Run `VUE_APP_DEV_SERVER_PROXY=/path/to/manual-fixtures/ npm run serve`
2. Go to http://localhost:8080/documentation/framework/sloth
3. Assert that tables' markup is as described in the Summary.

## Checklist

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #690, opened 2023-06-02
Revert #40 and #683

mportiz08:
This PR reverts #40 and #683

#40 introduces a regression that prevents certain pages from rendering (and #683 was only introduced to fix a separate regression that #40 also introduced.)

\cc @marinaaisa @dobromir-hristov


mportiz08:
@swift-ci test

eango (approved):

PR #692, opened 2023-06-09
[AX] <table> should be taken out of <figure> and the <figcaption> should become the table’s <caption>

marinaaisa:
Bug/issue #30234820, if applicable: 

## Summary

Tables were contained within a `<figure>` element. However, there is no need for this and it was making the tables much less accessible. Tables should stand on their own with the previous `<figcaption>` as the table’s `<caption>`. 

This is a improved implementation of https://github.com/apple/swift-docc-render/pull/40 that was not taking into account https://github.com/apple/swift-docc-render/pull/404 and had to be reverted.

## Dependencies

NA

## Testing

[manual-fixtures.zip](https://github.com/apple/swift-docc-render/files/7803705/manual-fixtures.zip)

Steps:
1. Run `VUE_APP_DEV_SERVER_PROXY=/path/to/manual-fixtures/ npm run serve`
2. Go to http://localhost:8080/documentation/framework/sloth
3. Assert that tables' markup is as described in the Summary.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -213,9 +213,21 @@ function renderNode(createElement, references) {
     }
   };
 
+  const placeCaption = (rawContent, { abstract = [], title, ...others }) => {
+    const content = [rawContent];
+    if (!abstract.length) return content;
+
+    // if there is a `title`, it should be above, otherwise below
+    content.splice(title ? 0 : 1, 0,
+      createElement(Caption, {
+        props: { title, centered: !title, ...others },
+      }, renderChildren(abstract)));
+    return content;
+  };
+
   const renderFigure = ({
     metadata: {
-      abstract = [],
+      abstract,
       anchor,
       title,
       ...metadata
@@ -226,14 +238,12 @@ function renderNode(createElement, references) {
       ...rest,
       metadata,
     };
-    const figureContent = [renderChildren([node])];
-    if (abstract.length) {
-      // if there is a `title`, it should be above, otherwise below
-      figureContent.splice(title ? 0 : 1, 0,
-        createElement(Caption, {
-          props: { title, tag: 'figcaption', centered: !title },
-        }, renderChildren(abstract)));
-    }
+    const figureContent = placeCaption(renderChildren([node]), {
+      title,
+      abstract,
+      tag: 'figcaption',
+    });
+
     return createElement(Figure, { props: { anchor } }, figureContent);
   };
 
@@ -298,18 +308,19 @@ function renderNode(createElement, references) {
       ));
     }
     case BlockType.table: {
-      const tableContent = [renderTableChildren(
+      let tableContent = [renderTableChildren(
         node.rows, node.header, node.extendedData, node.alignments,
       )];
+
       if (node.metadata && node.metadata.anchor) {
-        const { title, abstract = [] } = node.metadata;
-        if (abstract.length) {
-          tableContent.splice(title ? 0 : 1, 0,
-            createElement(Caption,
-              { props: { title, centered: !title } },
-              renderChildren(abstract)));
-        }
+        tableContent = placeCaption(renderTableChildren(
+          node.rows, node.header, node.extendedData, node.alignments,
+        ), {
+          title: node.metadata.title,
+          abstract: node.metadata.abstract,
+        });
       }
+
       return createElement(
         Table,
         {
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -17,7 +17,7 @@ import CodeVoice from './ContentNode/CodeVoice.vue';
 import DictionaryExample from './ContentNode/DictionaryExample.vue';
 import EndpointExample from './ContentNode/EndpointExample.vue';
 import Figure from './ContentNode/Figure.vue';
-import FigureCaption from './ContentNode/FigureCaption.vue';
+import Caption from './ContentNode/Caption.vue';
 import InlineImage from './ContentNode/InlineImage.vue';
 import Reference from './ContentNode/Reference.vue';
 import Table from './ContentNode/Table.vue';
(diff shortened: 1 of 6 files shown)


marinaaisa:
@swift-ci test

src/components/ContentNode.vue
```diff
@@ -297,18 +307,29 @@ function renderNode(createElement, references) {
         renderChildren(node.inlineContent)
       ));
     }
-    case BlockType.table:
+    case BlockType.table: {
+      let tableContent = [renderTableChildren(
```
dobromir-hristov, line 311: I dont think you need to wrap this in an array 🤔

src/components/ContentNode.vue
```diff
@@ -297,18 +307,29 @@ function renderNode(createElement, references) {
...
+        node.rows, node.header, node.extendedData, node.alignments,
+      )];
+
       if (node.metadata && node.metadata.anchor) {
-        return renderFigure(node);
+        tableContent = placeCaption(renderTableChildren(
+          node.rows, node.header, node.extendedData, node.alignments,
+        ), {
+          title: node.metadata.title,
+          abstract: node.metadata.abstract,
```
dobromir-hristov, line 320: I would probably pass the `tag` here explicitly, just so its obvious.

dobromir-hristov (commented):
Left a few minor comments, but I think we are in a good shape here.

src/components/ContentNode.vue
```diff
@@ -226,14 +238,12 @@ function renderNode(createElement, references) {
       ...rest,
       metadata,
     };
-    const figureContent = [renderChildren([node])];
-    if ((title && abstract.length) || abstract.length) {
```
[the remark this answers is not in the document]
marinaaisa: In case there is no `title` the `or` conditional makes it valid because there is `abstract.length` anyways.
I didn't understand why the `title` condicional was included in the first place. 🤔

src/components/ContentNode.vue
marinaaisa: True, I left it behind after a refactor. I deleted it now

src/components/ContentNode.vue
marinaaisa: okay, I did it here https://github.com/apple/swift-docc-render/pull/692/commits/2d2864bdedd0c0e9a6eec699493e1d91bd1c919d

dobromir-hristov (approved):
LGTM

marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

src/styles/core/typography/_font-styles.scss
```diff
@@ -153,7 +153,7 @@ $font-styles: (
   documentation-code-listing-number: (
     large: 12_18_normal_compact_mono,
   ),
-  documentation-figcaption: (
+  documentation-caption: (
```
mportiz08, line 156: I think I might have missed this in the original PR too, but you'll also need to update the place where this is being used with the old name in `Caption`

src/components/ContentNode/Caption.vue
```diff
@@ -31,6 +31,11 @@ export default {
       type: Boolean,
       default: false,
     },
+    tag: {
+      type: String,
+      required: false,
+      default: () => 'caption',
```
mportiz08, line 37: We should add a validator to make clear that this can only ever be one of `'caption'` or `'figcaption'` and not any arbitrary HTML tag.

src/components/ContentNode/Caption.vue
```diff
@@ -53,4 +58,8 @@ export default {
 /deep/ p {
   display: inline-block;
 }
+
+caption {
```
mportiz08, line 62: I think we only want a top margin to be applied if it is preceded by some other content, right?

It might be worth looking at the way this is/was done for the figcaption/figure spacing in [`Figure.vue`][1]. It sounds like we would want to do something similar for the `caption` element and the contents within a table.

[1]: https://github.com/marinaaisa/swift-docc-render/blob/f81e35c7391f25cce99ff7385dc446ea1804aa76/src/components/ContentNode/Figure.vue#L28

tests/unit/components/ContentNode.spec.js
```diff
@@ -112,29 +112,6 @@ describe('ContentNode', () => {
       expect(codeListing.props('content')).toEqual(listing.code);
       expect(codeListing.isEmpty()).toBe(true);
     });
-
-    it('renders a `Figure`/`Figcaption` with metadata', () => {
```
mportiz08, line 116: Why was this test removed?

mportiz08 (changes requested):
I think this resolves the previous issue with crashing regression for certain content.

I did find a couple of very minor styling regressions with the font and spacing that we should fix before merging.

tests/unit/components/ContentNode.spec.js
marinaaisa: I thought it was duplicated but in fact it wasn't 😅 I put it back!

src/components/ContentNode/Caption.vue
marinaaisa: Because we are adding the caption inside the table, I couldn't apply the margin to its siblings, it won't work in a display table, so I applied it to the component itself checking if there are siblings, using the `has` pseudo-class. It's widely supported now (only Firefox is missing) https://caniuse.com/css-has https://github.com/apple/swift-docc-render/pull/692/commits/d814cd80923aba1bde7e919e7817c6c9d70b1415

Can we use `has`?

src/components/ContentNode/Caption.vue
marinaaisa: Done https://github.com/apple/swift-docc-render/pull/692/commits/d814cd80923aba1bde7e919e7817c6c9d70b1415

src/components/ContentNode/Caption.vue
mportiz08: I started reviewing this again, but I found out that the `<caption>` element has to be the first element of the `<table>`, regardless of whether or not it is displayed on the top or the bottom. 🤦‍♂️🥲

I was going to make some suggestions to this PR, but it turned out to just create a new PR. Hopefully we finally have it right now! https://github.com/apple/swift-docc-render/pull/706

I also ran into issues you probably did where our normal block-level spacing helpers don't work due to specific issues with certain elements, but I was able to address them without `:has()`.

marinaaisa:
Closing in favor of https://github.com/apple/swift-docc-render/pull/706#pullrequestreview-1511328061

PR #693, opened 2023-06-13
Move the references setting back to the DocTopic component

dobromir-hristov:
Bug/issue #, if applicable: 109856608

## Summary

Moves the setting of the references back into the `components/DocumentationTopic` instead of the `view`.

This should fix potential rendering issues in Safari.

## Dependencies

NA

## Testing

Ensure all links and interact-able elements on pages are rendered properly and console has no errors.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -622,6 +622,7 @@ export default {
     }
 
     this.store.reset();
+    this.store.setReferences(this.references);
   },
 };
 </script>
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -624,6 +624,12 @@ export default {
     this.store.reset();
     this.store.setReferences(this.references);
   },
+  watch: {
+    // update the references in the store, in case they update, but the component is not re-created
+    references(references) {
+      this.store.setReferences(references);
+    },
+  },
 };
 </script>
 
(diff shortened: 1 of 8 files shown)


src/components/DocumentationTopic.vue
```diff
@@ -622,6 +622,7 @@ export default {
     }
 
     this.store.reset();
+    this.store.setReferences(this.references);
```
mportiz08, line 625: Sent you more details offline, but we may need to still consider having some way of also doing this as part of the update cycle and not just the initial created hook to support situations where the references may change during the lifetime of the component.

This does seem to fix the rendering issue that we noticed with references in bulleted lists though! Seems like you may have found that there is a race condition with the original way we're setting these refs causing that.

dobromir-hristov:
@swift-ci test

src/components/DocumentationTopic.vue
```diff
@@ -622,6 +622,13 @@ export default {
...
+  },
+  watch: {
+    // update the references in the store, in case they update, but the component is not re-created
+    references(references) {
```
mportiz08, line 629: Should we also add this watcher to other top-level components?

mportiz08 (approved):
Seems to be working well in both the initial and updating contexts while still resolving the race condition impacting bulleted lists with links.

dobromir-hristov:
@swift-ci test

dobromir-hristov:
@swift-ci test

PR #695, opened 2023-06-15
Keep scroll position on page nav when there is a filter applied

dobromir-hristov:
Bug/issue #, if applicable: 110300795

## Summary

Changes the scroll logic in the Navigator, to keep current scroll position, if you are filtering and the current page is in the results. Also keeps position if you are filtering and navigate to a page.

## Dependencies

NA

## Testing

Steps:
1. Scroll to the bottom of the nav
2. Click on a page that is at the bottom
3. Filter so that the current page is in the results. Assert its in the viewport (nav is not scrolled to top intentionally)
4. Navigate a few pages and assert the nav keeps every page in viewport
5. Filter so that the current page is NOT in the results. Assert nav is scrolled to top.
6. Move back/forward in navigation, and assert pages that are in the nav are scrolled to, and those that arent cause the nav to scroll to top.

## Checklist

- [ ] Added tests


dobromir-hristov:
@swift-ci test

marinaaisa (approved):
I've been testing and I assert that the issue is fixed now

PR #706, opened 2023-06-30
Use <caption> for table captions instead of <figcaption>

mportiz08:
Bug/issue #, if applicable: 30234820

## Summary

Use `<caption>` for tables instead of `<figcaption>`.
    
Tables with captions will no longer be embedded in a `<figure>` with a `<figcaption>`. Instead, the `<table>` will simply have a `<caption>` as its first element when needed.
    
For table captions, the position of the caption must be done in CSS since the `<caption>` is always required to be the first element of the `<table>` and cannot be moved to the end like they can with figure captions.

## Testing

Steps:
1. Find a page with a captioned table
2. Verify that the DOM structure now looks like `<table><caption>` instead of `<figure><figcaption><table>`
3. Verify that there are no visual regressions with figures, captions, or tables

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/ContentNode/Caption.vue
+++ b/src/components/ContentNode/Caption.vue
@@ -26,7 +26,7 @@ const CaptionTag = {
 };
 
 export default {
-  name: 'FigureCaption',
+  name: 'Caption',
   constants: { CaptionTag },
   props: {
     title: {
@@ -50,7 +50,7 @@ export default {
 @import 'docc-render/styles/_core.scss';
 
 .caption {
-  @include font-styles(documentation-figcaption);
+  @include font-styles(documentation-caption);
 
   &.centered {
     text-align: center;
--- a/src/components/ContentNode/Caption.vue
+++ b/src/components/ContentNode/Caption.vue
@@ -9,7 +9,7 @@
 -->
 
 <template>
-  <component class="caption" :class="{ centered }" :is="tag">
+  <component class="caption" :class="{ trailing }" :is="tag">
     <template v-if="title">
       <strong>{{ title }}</strong>&nbsp;<slot />
     </template>
@@ -25,23 +25,35 @@ const CaptionTag = {
   figcaption: 'figcaption',
 };
 
+const CaptionPosition = {
+  leading: 'leading', // before element and left aligned
+  trailing: 'trailing', // after element and center aligned
+};
+
 export default {
   name: 'Caption',
-  constants: { CaptionTag },
+  constants: {
+    CaptionPosition,
+    CaptionTag,
+  },
   props: {
     title: {
       type: String,
       required: false,
-    },
-    centered: {
-      type: Boolean,
-      default: false,
     },
     tag: {
       type: String,
       required: true,
       validator: v => Object.hasOwnProperty.call(CaptionTag, v),
     },
+    position: {
+      type: String,
+      default: () => CaptionPosition.leading,
+      validator: v => Object.hasOwnProperty.call(CaptionPosition, v),
+    },
+  },
+  computed: {
+    trailing: ({ position }) => position === CaptionPosition.trailing,
   },
 };
 </script>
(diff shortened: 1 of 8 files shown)


src/components/ContentNode/Caption.vue
```diff
@@ -0,0 +1,87 @@
...
+  margin: 0 0 var(--spacing-stacked-margin-large) 0;
+
+  // `caption-side` must be used for the table version of <caption> to appear
+  // underneath the table since the element must always be the first element
+  // within the <table> in the DOM according to the spec
+  &.centered {
+    caption-side: bottom;
+    margin: var(--spacing-stacked-margin-large) 0 0 0;
+  }
+}
```
marinaaisa, line 82: I believe we could simplify all these styles into:

```scss
.caption {
  @include font-styles(documentation-caption);
  margin: 0 0 var(--spacing-stacked-margin-large) 0;

  &.centered {
    text-align: center;
    margin: var(--spacing-stacked-margin-large) 0 0 0;
    // caption-side only applies for tables
    caption-side: bottom;
  }
}
```
because in both `figcaption` and `caption` cases, we have 2 styles possible: on top-left (default) or bottom-centered (when we add the .centered class).

if we are worried that an extra margin would be add in case there is no siblings, we could add:
```scss
&:only-child {
  margin: 0;
}
```
but I don't think this use case will ever happen.

marinaaisa (changes requested):
Thanks Marcus for reviewing https://github.com/apple/swift-docc-render/pull/692 and submitting this new PR!

Since <caption> element has to be the first element of the <table>, this makes much more sense! 

I reflected on the `centered` prop and I proposed a refactor in the CSS. Let me know if it makes sense or I'm missing some use case. (Hopefully not!)

src/components/ContentNode/Caption.vue
mportiz08: Hm, I'll see if I can simplify it more, but I'm not sure we would be able to use that suggestion as-is. The `caption-side` is supposed to only apply to `<caption>` elements and not `<figcaption>` ones.

Also, it would be nice to have the margin rules dependent on siblings just to make it a little more flexible if someone wanted to use the component elsewhere, but maybe that's not a big deal since it should always be used with siblings anyway.

I'll see if it makes sense to simplify more though.

src/components/ContentNode/Caption.vue
```diff
@@ -0,0 +1,87 @@
...
+
+export default {
+  name: 'Caption',
+  constants: { CaptionTag },
+  props: {
+    title: {
+      type: String,
+      required: false,
+    },
+    centered: {
```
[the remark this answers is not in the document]
mportiz08: Maybe. It is more clear but also kind of an odd property name. I'll think on something better.

src/components/ContentNode/Caption.vue
mportiz08: I think this is the best I can come up with at the moment: https://github.com/apple/swift-docc-render/pull/706/commits/54a6137a9bb80c3789cac7bb35cf52caa40c498f

src/components/ContentNode/Caption.vue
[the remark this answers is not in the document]
mportiz08: https://github.com/apple/swift-docc-render/pull/706/commits/54a6137a9bb80c3789cac7bb35cf52caa40c498f

I think the position makes more sense as a prop and it's just an implementation detail that we center the bottom positioned one. I considered leaving out the prop entirely since it is always derived from the presence of the title, but I wanted it to be slightly more flexible in case we wanted to use it anywhere else.

dobromir-hristov (approved):
Tested in various scenarios, looks great!

src/components/ContentNode/Caption.vue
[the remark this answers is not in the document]
marinaaisa: Looks good!

src/components/ContentNode/Caption.vue
marinaaisa: Looks good!

marinaaisa (approved):
👏

mportiz08:
@swift-ci test

PR #708, opened 2023-07-03
Code blocks should not be tabbable 

hqhhuang:
Bug/issue #, if applicable: rdar://111580884

## Summary
Code blocks should not be tubbable unless there are links within the code block.

### DIFF
--- a/src/components/CodeBlock.vue
+++ b/src/components/CodeBlock.vue
@@ -10,7 +10,6 @@
 
 <template>
   <code
-    tabindex="0"
     :data-before-code="$t('accessibility.code.start')"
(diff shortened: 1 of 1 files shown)


dobromir-hristov (commented):
I guess we need to asses two things:

1. If the blue outline is the problem, we could hide it for IDE mode.
2. If just declarations are to not be focusable, we could remove the tabindex for them.
3. We could prevent focusable codeblocks in specific situations, like minified view.
4. We need to see why the `code` is not being scrolled with arrows when focused. I tried in Chrome and it worked fine for me.

PR #710, opened 2023-07-05
[AX] Prevented from adding #main anchor into the URL when skipping navigation

marinaaisa:
Bug/issue #109460523, if applicable: 

## Summary

Skip navigation only works once on a page. Because when we trigger the skip, we insert the `#main` anchor in the URL of the current page. Then, when we try to skip again, it does not work properly.

To fix this issue, I prevented it from adding main anchor into the URL.

## Dependencies

NA

## Testing

Steps:
1. Go to any page
2. Start tabbing and you will see "Skip navigation" text
3. Assert that when you click on it, it focuses the main element and then you can continue tabbing from there. But it doesn't add "#main" into the URL

## Checklist

- [x] Added tests


marinaaisa:
@swift-ci test

marinaaisa:
Closing this PR since it doesn't provide enough value to the AX.

PR #440, opened 2022-09-23
Update Vue@2.7

dobromir-hristov:
Bug/issue #, if applicable: 100963792

## Summary

Updates Vue to 2.7

## Dependencies

- https://github.com/dobromir-hristov/swift-docc-render/pull/15 - this PR upgrades on top of the current one

## Testing

Make sure it builds/serves as normal without warnings.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #711, opened 2023-07-06
[5.9] Update Vue to v2.7, VueCLI to v5, Node to v18, NPM to v9

dobromir-hristov:
- **Explanation:** Updates major prod and dev deps, like Vue, VueCLI, Node to latest stable and supported versions. Removes critical dep issues.
- **Scope:** Entire app, as the dev-tools got updated
- **Issue:** rdar://100963792
- **Risk:** High, updates dev dependencies and Vue version, with slight css changes
- **Testing:** Manually tested app in lots of scenarios
- **Reviewer:** @marinaaisa, @hqhhuang, @mportiz08 
- **Original PR:** #440


dobromir-hristov:
@swift-ci test

mportiz08:
I would personally consider this as potentially "High" risk simply due to the number of files affected and the very wide impact, although everything seems to working well from my testing.

src/components/CallToActionButton.vue
```diff
@@ -9,9 +9,8 @@
 -->
 
 <template>
-  <DestinationDataProvider v-if="action" :destination="action">
+  <DestinationDataProvider v-if="action" :destination="action" v-slot="{ url, title }">
```
eango, line 12: Has the Call to Action been tested with this change? I recall we had an issue with this in the past

dobromir-hristov: This is just a syntax change for the slot. The tests would have failed if its wrong :)

franklinsch:
@dobromir-hristov Can you elaborate on the benefit of including this in the 5.9 release? Is it worth the risk you've outlined?

dobromir-hristov:
We wanted to get this out of the door as soon as possible, as docc-render has lots of critical dependency vulnerabilities that we cannot fix in other ways. 

Ofc if you think its risky to do it now, we can wait for a more appropriate time.

mportiz08:
We discussed this and feel that it is probably too risky to take for 5.9 due to the large number of changes and limited end-user benefit.

We will continue with the updated dependencies on main going forward though.

PR #717, opened 2023-07-11
don't display declaration platforms if they're all the same

QuietMisdreavus:
Bug/issue #, if applicable: None (yet)

## Summary

This is an experiment to accompany https://github.com/apple/swift-docc/pull/654 - That PR and its SymbolKit dependency change the Declarations section to potentially include multiple declarations from the same platform. This PR updates the Declaration component to drop the platform caption if all the declaration groups have the same platform sets.

Before this change, this is what this situation would look like:

image

With this PR, the symbol looks like this: (this screenshot is using a local Swift-DocC-Render build on my machine, but the above screenshot is using the Swift-DocC-Render that shipped with Xcode 14.3.1, which hopefully explains the padding and heading differences)

image

## Dependencies

This can land without it, but it's only useful with https://github.com/apple/swift-docc/pull/654.

## Testing

Use the branch from https://github.com/apple/swift-docc/pull/654, then, in the Swift-DocC repo:

Steps:
1. `DOCC_HTML_DIR=/path/to/swift-docc-render/dist swift run docc preview 'Tests/SwiftDocCTests/Test Bundles/AlternateDeclarations.docc'`
2. Ensure that the `MyClass/present(completion:)` page doesn't print the "macOS" caption on its declarations.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic/PrimaryContent/Declaration.vue
+++ b/src/components/DocumentationTopic/PrimaryContent/Declaration.vue
@@ -53,6 +53,7 @@ import DeclarationSourceLink
 
 import { ChangeTypes } from 'docc-render/constants/Changes';
 import { multipleLinesClass } from 'docc-render/constants/multipleLines';
+import { isEqual } from 'docc-render/utils/arrays';
 
 export default {
   name: 'Declaration',
@@ -85,11 +86,18 @@ export default {
   computed: {
     /**
      * Show the captions of DeclarationGroup without changes
-     * when there are more than one declarations
+     * when there are multiple sets of platforms among the
+     * declarations.
      * @returns {boolean}
      */
     hasPlatformVariants() {
-      return this.declarations.length > 1;
+      const platforms = [];
+      for (let i = 0; i < this.declarations.length; i += 1) {
+        if (!platforms.some(platform => isEqual(platform, this.declarations[i].platforms))) {
+          platforms.push(this.declarations[i].platforms);
+        }
+      }
+      return platforms.length > 1;
     },
     /**
      * Returns whether there are declaration changes.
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
@swift-ci Please test

src/components/DocumentationTopic/PrimaryContent/Declaration.vue
```diff
@@ -89,11 +90,18 @@ export default {
...
     /**
      * Show the captions of DeclarationGroup without changes
-     * when there are more than one declarations
+     * when there are multiple sets of platforms among the
+     * declarations.
      * @returns {boolean}
      */
     hasPlatformVariants() {
-      return this.declarations.length > 1;
+      const platforms = [];
```
mportiz08, line 98: Could we possibly just create a computed property that is an object where the keys are the stringified list of platforms and the values are the actual lists of declarations mapped to those platforms? Likely the reason we weren't doing this before is because it was always expected to have a unique list of platforms per-declaration, and it sounds like that is no longer the case.

marinaaisa (approved):
I cleaned up a bit of the code here https://github.com/swiftlang/swift-docc-render/pull/717/commits/969fa50f287406697b1cda08a6e6f2fd54fba008 but everything else looks good to me!

Thank you!

src/components/DocumentationTopic/PrimaryContent/Declaration.vue
marinaaisa: After speaking with @hqhhuang I agree that since we need `platforms` and `tokens` in the same object for `DeclarationList` it may be easier and safer to keep it as it is.

PR #723, opened 2023-07-31
Fix issue where backslash newline delimiters in multiline Swift string literals break syntax highlighting

mportiz08:
Bug/issue #, if applicable: 102993199

## Summary

Fixes an issue demonstrated in the following screenshots where the syntax highlighting looks broken for Swift code listings with multi-line string literals that have backslash newline delimiters.

Before:
before

After:
after

### Details

This happens because the HTML `<span>` for the "subst" token representing the backslash also includes the newline whitespace, and that causes issue when our line-by-line logic for multi-line syntax encounters this span that is broken onto multiple lines.

To fix it, we are simply overriding the default sub-mode from highlight.js for this token to _not_ include the newline characters in the resulting `<span>`.

## Testing

Steps:
1. Checkout this branch and run the dev server with `VUE_APP_DEV_SERVER_PROXY=https://docs.swift.org/swift-book npm run serve`
2. Open http://localhost:8080/documentation/the-swift-programming-language/stringsandcharacters and verify that the broken code listings with backslash delimiters look right now
3. Compare with https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters and verify that there are no other regressions related to Swift string literal syntax highlighting

## Checklist

- [X] Added tests

### DIFF
--- a/src/utils/custom-highlight-lang/swift.js
+++ b/src/utils/custom-highlight-lang/swift.js
@@ -65,7 +65,7 @@ export default function swiftOverride(hljs) {
       ...variant,
       contains: variant.contains.map(mode => (isEscapedNewlineMode(mode) ? ({
         className: 'subst',
-        begin: /\\/,
+        begin: /\\#{0,3}/,
         end: /[\t ]*(?:[\r\n]|\r\n)/,
         // same match as the original one but with an explicit start/end match so
         // that the end one can be excluded from the resulting span
--- a/src/utils/custom-highlight-lang/swift.js
+++ b/src/utils/custom-highlight-lang/swift.js
@@ -41,5 +41,20 @@ export default function swiftOverride(hljs) {
     };
   }
 
+  const str = language.contains.find(({ className }) => className === 'string');
+  str.variants = str.variants.map(variant => ({
+    ...variant,
+    // omit the "ESCAPED_NEWLINE"/("subst") highlight.js tokenizers that
+    // highlight.js uses to syntax highlight backslash delimiters in multi-line
+    // string literals to prevent an issue where they interrupt the multi-line
+    // string literal tokenizer
+    contains: variant.contains.reduce((acc, c) => {
+      if (c.className === 'subst' && c.match) {
+        return acc;
+      }
+      return acc.concat(c);
+    }, []),
+  }));
+
   return language;
 }
--- a/src/utils/custom-highlight-lang/swift.js
+++ b/src/utils/custom-highlight-lang/swift.js
@@ -41,19 +41,36 @@ export default function swiftOverride(hljs) {
     };
   }
 
+  // Checks if a given language sub-mode matches the "ESCAPED_NEWLINE" from the
+  // built-in Swift parser from hljs
+  const isEscapedNewlineMode = (mode) => {
+    const { className, match } = mode;
+    if (className !== 'subst' || !match) {
+      return false;
+    }
+
+    const matchStr = match.toString();
+    return matchStr.startsWith('\\') && matchStr.endsWith('[\\t ]*(?:[\\r\\n]|\\r\\n)');
+  };
+
+  // replace the "ESCAPED_NEWLINE" sub-mode in the multiline string literal mode
+  // variants so that it doesn't include the actual newline characters in the
+  // span token that it generates, because this causes issues with our
+  // line-number + multi-line string literal logic when the span for the
+  // backslash token is split across multiple lines
   const str = language.contains.find(({ className }) => className === 'string');
   str.variants = str.variants.map(variant => ({
     ...variant,
-    // omit the "ESCAPED_NEWLINE"/("subst") highlight.js tokenizers that
-    // highlight.js uses to syntax highlight backslash delimiters in multi-line
-    // string literals to prevent an issue where they interrupt the multi-line
-    // string literal tokenizer
-    contains: variant.contains.reduce((acc, c) => {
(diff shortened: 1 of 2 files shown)


mportiz08:
For reference, the default sub-mode from highlight.js being overwritten here can be seen at https://github.com/highlightjs/highlight.js/blob/2a972d86589f65f4e79f67a02c3bf389a187afa0/src/languages/swift.js#L163C2-L163C2

mportiz08:
\cc @amartini51

src/utils/custom-highlight-lang/swift.js
```diff
@@ -41,5 +41,37 @@ export default function swiftOverride(hljs) {
...
+    return matchStr.startsWith('\\') && matchStr.endsWith('[\\t ]*(?:[\\r\\n]|\\r\\n)');
+  };
+
+  // replace the "ESCAPED_NEWLINE" sub-mode in the multiline string literal mode
+  // variants so that it doesn't include the actual newline characters in the
+  // span token that it generates, because this causes issues with our
+  // line-number + multi-line string literal logic when the span for the
+  // backslash token is split across multiple lines
+  const str = language.contains.find(({ className }) => className === 'string');
+  str.variants = str.variants.map(variant => ({
```
dobromir-hristov, line 62: This is relying changing the `variants` inside a single object in the `contains` array, purely by reference? 🤣 pure 🪄 🦄

src/utils/custom-highlight-lang/swift.js
```diff
@@ -41,5 +41,37 @@ export default function swiftOverride(hljs) {
...
+    ...variant,
+    contains: variant.contains.map(mode => (isEscapedNewlineMode(mode) ? ({
+      className: 'subst',
+      begin: /\\/,
+      end: /[\t ]*(?:[\r\n]|\r\n)/,
```
dobromir-hristov, line 67: I guess this is the actual fix, right? Giving it a `start` and an `end`, instead of an open `match`?

dobromir-hristov (approved):
Great job Marcus, this is quite a tricky one.

src/utils/custom-highlight-lang/swift.js
mportiz08: Yeah, I don't love it tbh. Maybe I could refactor it to not rely on changing it by reference, but I worry about how much more code I need to write 😿

mportiz08: Here's a less 🪄🦄 version: eaa410127784b057baa8b03db9287a5ab172d2bf 😆

src/utils/custom-highlight-lang/swift.js
mportiz08: Also quickly added support for the optional extended `#` delimiters with 685c910e89d5671f8cbfa07edefdc96518b4f748

src/utils/custom-highlight-lang/swift.js
```diff
@@ -41,5 +41,40 @@ export default function swiftOverride(hljs) {
...
+  // line-number + multi-line string literal logic when the span for the
+  // backslash token is split across multiple lines
+  const strIndex = language.contains.findIndex(({ className }) => className === 'string');
+  language.contains[strIndex] = {
+    ...language.contains[strIndex],
+    variants: language.contains[strIndex].variants.map(variant => ({
+      ...variant,
+      contains: variant.contains.map(mode => (isEscapedNewlineMode(mode) ? ({
+        className: 'subst',
+        begin: /\\#{0,3}/,
```
dobromir-hristov, line 68: clever 🤔 should we add a test?

dobromir-hristov (approved):
🚀

dobromir-hristov:
@swift-ci test

src/utils/custom-highlight-lang/swift.js
mportiz08: I would, but I'm not familiar enough with the intricacies of this extended syntax to write a realistic sample 😅

jameschensmith:
Great fix! This still seems to be an issue in the current version of the Swift Book (5.9.2 beta). I'm not sure which version of `swift-docc-render` Swift Book is using, though, so the fix might not be there just yet.

image

amartini51:
@jameschensmith Thanks for the feedback about the book.  Maybe we can continue this discussion in an issue on [that repo](https://github.com/apple/swift-book/issues), or in the forums?

I think this is correct as-is.  The `\` at the end of these lines prevents the string literal from containing a newline, even though the source code starts a new line there.

jameschensmith:
> @jameschensmith Thanks for the feedback about the book. Maybe we can continue this discussion in an issue on [that repo](https://github.com/apple/swift-book/issues), or in the forums?
> 
> I think this is correct as-is. The `\` at the end of these lines prevents the string literal from containing a newline, even though the source code starts a new line there.

Hey @amartini51. I think this is the correct place to put this (unless the swift book is not up-to-date with the latest `swift-docc-render`). I'm aware of the usage of `\`. The problem is the syntax highlighting (for reference, I'm referring to [this section in the book](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters/#Multiline-String-Literals)). If you look at the before / after in the [OP](https://github.com/apple/swift-docc-render/pull/723#issue-1830217298), you'll notice the images are also probably taken from the Swift book (it's the same snippet I used in my reference image). I could create an issue in [swift-book](https://github.com/apple/swift-book) if you'd still like to have it tracked there. Let me know if there's still any confusion 👍

PR #724, opened 2023-08-01
Ensure highlight.js builds only the files that are imported

dobromir-hristov:
Bug/issue #, if applicable: 113252221

## Summary

Fixes and issue with webpack pulling in more highlight.js files than it should + fixes the node modules path

## Dependencies

NA

## Testing

Steps:
1. Ensure when building, we build the correct set of language files
2. Ensure the new import path does not cause errors when referenced by extending docc-render

## Checklist

- [ ] Added tests

### DIFF
--- a/src/setup-utils/vue-config-utils.js
+++ b/src/setup-utils/vue-config-utils.js
@@ -53,6 +53,7 @@ function baseChainWebpack(config) {
   // Add theme fallback resolver
   config.resolve.alias.set('theme', themeUtils.getThemePaths());
   config.resolve.alias.set('docc-render', path.resolve(__dirname, '../'));
+  config.resolve.alias.set('highlight-js-alias', path.dirname(require.resolve('highlight.js/package.json')));
 
   // Add license header to built files
   config
@@ -68,7 +69,7 @@ function baseChainWebpack(config) {
     .plugin('LanguagesPlugin')
     .use(webpack.ContextReplacementPlugin, [
       /highlight\.js\/lib\/languages$/,
-      new RegExp(`/(${[builtinLanguages, envLanguages].join('|')})$`),
+      new RegExp(`^./(${[builtinLanguages, envLanguages].filter(Boolean).join('|')}).js$`),
     ]);
 }
 
--- a/src/setup-utils/vue-config-utils.js
+++ b/src/setup-utils/vue-config-utils.js
@@ -69,7 +69,7 @@ function baseChainWebpack(config) {
     .plugin('LanguagesPlugin')
     .use(webpack.ContextReplacementPlugin, [
       /highlight\.js\/lib\/languages$/,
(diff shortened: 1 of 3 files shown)


src/setup-utils/vue-config-utils.js
```diff
@@ -68,7 +69,7 @@ function baseChainWebpack(config) {
     .plugin('LanguagesPlugin')
     .use(webpack.ContextReplacementPlugin, [
       /highlight\.js\/lib\/languages$/,
-      new RegExp(`/(${[builtinLanguages, envLanguages].join('|')})$`),
+      new RegExp(`^./(${[builtinLanguages, envLanguages].filter(Boolean).join('|')}).js$`),
```
mportiz08, line 72: Was the double building of languages files caused by the addition of `envLanguages` because this pattern was too inclusive in the common case where `envLanguages` is an empty string?

Just making sure I understand what went wrong and how you fixed it.

mportiz08: (If so, it sounds like that was introduced back with #551 and we just didn't notice it—whoops!)

mportiz08, line 72:
```suggestion
      new RegExp(`^./(${[builtinLanguages, envLanguages].filter(Boolean).join('|')})\.js$`),
```

src/utils/syntax-highlight.js
```diff
@@ -90,7 +90,7 @@ async function importHighlightFileForLanguage(language) {
         languageFile = await import(
           // See bug https://github.com/webpack/webpack/issues/13865
           /* webpackChunkName: "highlight-js-[request]" */
-          `@/../node_modules/highlight.js/lib/languages/${file}`
+          `highlight-js-alias/lib/languages/${file}`
```
mportiz08, line 93: Should we be using the new alias in all places where we import files from this package or is this the only import that needed to be explicitly changed to use the alias? Just curious.

I'm guessing just this one since it's the only one using the `@`.

mportiz08 (approved):
Looks good. Just have a few questions mostly for my own curiosity.

src/setup-utils/vue-config-utils.js
dobromir-hristov: There are a bunch of `css.js.js` files which are old deprecated file-naming convention files. Those were however being pulled into the build as well.

The extra `envLanguages` I just noticed can be empty and fixed it. I think the more restrictive regex is what is fixing it now.

src/utils/syntax-highlight.js
dobromir-hristov: Nah. the alias is just for here, because there is a webpack bug that we cannot dynamically import those lang files, so it needs an absolute path.

dobromir-hristov:
@swift-ci test

Issue #722, opened 2023-07-23
Blank articles when hosting DocC archives if library name is "Tutorials"


jiachenyee:
### Description

When hosting a DocC archive with a library named "Tutorials" (case insensitive) online, it results in a blank screen when trying to view articles. 

When an article is in the library, it hosts the article on a URL like this: "example.com/tutorials/articlename"

When trying to open it, it shows a blank screen.
![image](https://github.com/apple/swift-docc/assets/36725840/ec857874-ebe0-42b3-8ef7-4da665d248c7)

Here is a demo site: https://bug.jiachen.app/documentation/tutorials 
GitHub repository: https://github.com/jiachenyee/DemoDocCIssue/

I assume this somehow caused an issue with the way DocC hosts tutorials with a similar URL pattern, having tutorials on a URL like "example.com/tutorials/tutorialname".

I should point out that the article displays perfectly fine on Xcode, the only issue is when hosted online.

I have only had time to test it out with GitHub Pages and Xcode 14 Beta 4, but I assume the behaviour should be consistent regardless.

I have tested on multiple browsers, including Safari (Mac/iPhone), Chrome (Mac), and Firefox (Mac). The issue exists on all of them.

### Checklist

- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc/issues).

### Expected Behavior

It should display the contents of the article file. Here is a screenshot of what is displayed when the DocC Archive is opened within Xcode.

![Sample Article  - Xcode  2023-07-23 at 10 20 44 PM](https://github.com/apple/swift-docc/assets/36725840/7a1a3556-3dc8-4533-b384-4ef8cb3cf6f5)

### Actual behavior

This is what it looks like when viewing the hosted version.

![image](https://github.com/apple/swift-docc/assets/36725840/102175f2-a981-49a8-84dc-cbdbf0e0c08b)

View it for yourself: https://bug.jiachen.app/documentation/tutorials/article

### Steps To Reproduce

1. Create a new Swift Package in Xcode
2. Name the Package "Tutorials"
3. Create a Documentation.docc Documentation Catalog
4. Create an article file 
5. Retrieve the DocC archive 
6. When previewing the .doccarchive from Xcode, it looks perfectly fine
[...]

PR #726, opened 2023-08-02
Ensure articles with the name "Tutorials" render with the right component

mportiz08:
Bug/issue #, if applicable: 113289632 (#722)

## Summary

This fixes an issue where the wrong component is used to render a documentation article if it has the name "Tutorials".

This happens because the locale-specific tutorials-overview route matches the URL for this article before the expected route for documentation topics matches it. To fix it, I simply have re-ordered the routes definition so that the locale-specific versions come after the normal ones instead of before.

\cc @jiachenyee

## Testing

Steps:
1. Checkout this branch and run the dev server with `VUE_APP_DEV_SERVER_PROXY=https://bug.jiachen.app npm run serve`
2. Open http://localhost:8080/documentation/tutorials and verify that it gets rendered as a normal documentation article and doesn't appear to be a blank tutorials-overview page anymore
3. Check that other existing routes still work as expected, including locale-specific ones if possible

## Checklist

- [ ] ~~Added tests~~ (no setup to test route config)

### DIFF
--- a/src/setup-utils/SwiftDocCRenderRouter.js
+++ b/src/setup-utils/SwiftDocCRenderRouter.js
@@ -19,7 +19,7 @@ import { baseUrl } from 'docc-render/utils/theme-settings';
 import { addPrefixedRoutes } from 'docc-render/utils/route-utils';
 import { notFoundRouteName } from 'docc-render/constants/router';
 
-const defaultRoutes = [...addPrefixedRoutes(routes, [notFoundRouteName]), ...routes];
+const defaultRoutes = [...routes, ...addPrefixedRoutes(routes, [notFoundRouteName])];
 
 export default function createRouterInstance(routerConfig = {}) {
   const router = new Router({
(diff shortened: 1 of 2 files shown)


Kyle-Ye (changes requested):
I'd like to request adding a `tests/unit/routes.spec.js` file add some test cases.

For example, https://github.com/vuejs/vue-router/blob/e53a549f04e9e02d293d8a4c2537a046ae2381dc/test/unit/specs/create-matcher.spec.js#L51-L54

`expect(matcher.match('/documentation/tutorials/article').name).toBe('xx')`

jiachenyee:
Tested it on my end, LGTM!

dobromir-hristov (approved):
We dont know the languages upfront, to limit the list of matches, so I guess this is a better option.

Kyle-Ye (approved):

mportiz08:
> I'd like to request adding a `tests/unit/routes.spec.js` file and some test cases.
> 
> For example, https://github.com/vuejs/vue-router/blob/e53a549f04e9e02d293d8a4c2537a046ae2381dc/test/unit/specs/create-matcher.spec.js#L51-L54
> 
> `expect(matcher.match('/documentation/tutorials/article').name).toBe('xx')`

Normally I don't think we would want to be testing the vue-router behavior directly since the routes are effectively its configuration, but we do have some logic that impacts the order of the routes, and I agree that it would be helpful to have some tests to make sure we don't regress with this issue in the future.

I added some tests here: 70bff17785b049bdfe95ce6741f8866f9075b530

Thanks for reviewing!

Kyle-Ye:
@swift-ci test

PR #698, opened 2023-06-20
Aria-label that indicates scrolling behaviour interferes in headings content for screen readers

marinaaisa:
Bug/issue #111006953, if applicable: 

## Summary

Aria-label that indicates scrolling behaviour interferes in headings content for screen readers

It's been fixed by restructuring how the anchor link is set and adding AX description that should read: Content of heading + "in page link" to describe that the link is within the page.

Screenshot 2023-06-26 at 20 30 31

## Dependencies

NA

## Testing

Steps:
1. Run any doccarchive
2. Go to section title
3. Assert that you can read its content using VoiceOver

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #728, opened 2023-09-11
Disallow selection of invisible text for linkable headings

mportiz08:
Bug/issue #, if applicable: 115140459

## Summary

Prevents an issue where the text _"in page link"_ is unexpectedly included when selecting and copy/pasting a linkable heading.

## Testing

Steps:
1. Run `VUE_APP_DEV_SERVER_PROXY=https://mportiz08.github.io/swift-docc npm run serve` and open http://localhost:8080/documentation/docc (or any content with linkable subheadings)
2. Use the mouse to select and copy/paste the text of a linkable heading, like [this one](http://localhost:8080/documentation/docc#structure-and-formatting)
3. Verify that the pasted text matches the heading and does not include the text "in page link" at the end anymore

## Checklist

### DIFF
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -95,4 +95,8 @@ $icon-margin: 7px;
     }
   }
 }
+
+.visuallyhidden {
+  user-select: none;
+}
 </style>
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -16,11 +16,11 @@
     <router-link
       v-if="shouldLink"
       :to="{ hash: `#${anchor}` }"
+      :data-after-text="$t('accessibility.in-page-link')"
       class="header-anchor"
       @click="handleFocusAndScroll(anchor)"
     >
       <slot />
-      <span class="visuallyhidden" >{{ $t('accessibility.in-page-link') }}</span>
       <LinkIcon class="icon" aria-hidden="true" />
     </router-link>
     <template v-else>
@@ -79,6 +79,11 @@ $icon-margin: 7px;
   position: relative;
   padding-right: $icon-size-default + $icon-margin;
   display: inline-block;
+
+  &::after {
+    @include visuallyhidden;
+    content: attr(data-after-text);
+  }
 
   .icon {
     position: absolute;
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -100,8 +100,4 @@ $icon-margin: 7px;
     }
   }
 }
-
-.visuallyhidden {
-  user-select: none;
-}
 </style>
--- a/tests/unit/components/ContentNode/LinkableHeading.spec.js
+++ b/tests/unit/components/ContentNode/LinkableHeading.spec.js
@@ -48,7 +48,8 @@ describe('LinkableHeading', () => {
     expect(wrapper.attributes('id')).toBe('title');
     const headerAnchor = wrapper.find('.header-anchor');
     expect(headerAnchor.props('to')).toEqual({ hash: '#title' });
-    expect(headerAnchor.text()).toBe('Title accessibility.in-page-link');
+    expect(headerAnchor.text()).toBe('Title');
+    expect(headerAnchor.attributes('data-after-text')).toBe('accessibility.in-page-link');
   });
 
   it('does not render anchor if there is no anchor', () => {


mportiz08:
@swift-ci test

src/components/ContentNode/LinkableHeading.vue
```diff
@@ -95,4 +95,8 @@ $icon-margin: 7px;
     }
   }
 }
+
+.visuallyhidden {
```
mportiz08, line 99: @marinaaisa as written, I'm only doing this for the linkable heading scenario, but do you think it might make sense to lift this rule to _all_ visually hidden elements? I wasn't sure, so I just did it this way for now, but I wouldn't mind making the change for everything if that makes sense.

dobromir-hristov: Good question 🤔 I doubt those elements need to be selected.

marinaaisa: I agree that adding `user-select: none;` to `visually-hidden` would be useful.
But there is other problem: this CSS property only solves the case you only copy text of the title, if you start copying from some other text and continue with the title, the "in page link" is still there. 🤔

marinaaisa: One solution that comes to my mind is adding an event listener to `copy` or `select` that adds `display: none` (this prevents from copying but also prevents screen readers to read the content) to class `visually-hidden`. Then we will need to delete the `display` property when we consider that the action is done with another event listener (not sure in which)

hqhhuang: I think the "in page link" is still there because although the `span` is visually hidden and can't be selected, it's still part of the name of the link. When we copy the heading by selecting it, it's copying the entire link, not just the text (if you paste into Slack, it will format the result into a link automatically instead of plain text). 

This is what I see when I click & drag on the text portion of the link
image

hqhhuang: Looks like we were using aria-labels for accessibility for the link, but changed it [here](https://github.com/apple/swift-docc-render/pull/698). I don't completely understand what was the issue with the aria-label in the original implementation? Is it just preventing scrolling to the correct position?

mportiz08: Ah, I see what I'm missing now—this still happens on Chrome when using paste-with-formatting—not just a regular paste (without formatting).

dobromir-hristov: We could use this mechanism, its commonly used for line-numbers in code listings, which is how we use it too afaik.

```css
.link::before {
  content: attr(data-accessible-text);
}

Comment (marinaaisa) [reply src/components/ContentNode/LinkableHeading.vue:99]: > Looks like we were using aria-labels for accessibility for the link, but changed it here. I don't completely understand what was the issue with the aria-label in the original implementation? Is it just preventing scrolling to the correct position?

@hqhhuang with the aria-label we were replacing the content of the headers for screen readers, so unfortunately it's not an option :/

Comment (marinaaisa) [reply src/components/ContentNode/LinkableHeading.vue:99]: > We could use this mechanism, its commonly used for line-numbers in code listings, which is how we use it too afaik.
> 
> ```css
> .link::before {
>   content: attr(data-accessible-text);
> }
> ```

We do this on the `StrikeThrough.vue` too, it could be a solution. But it won't fix the `visuallyhidden` class issue in other cases of the web. We use this class to provide AX information in other places across the web and I guess we have the same issue with copy and pasting.

Comment (marinaaisa) [reply src/components/ContentNode/LinkableHeading.vue:99]: Another alternative is using `display: none` instead of `visuallyhidden`. So it won't appear for copying, screen readers or visual users but we can reference it with an `id` that we can use in `aria-labelledby` after the id of the router-link itself, to contain the content of it too. I tried it and it works.

```vue
<router-link
  v-if="shouldLink"
  :to="{ hash: `#${anchor}` }"
  id="router-link"
  :aria-labelledby="`router-link ax-in-page-link-${_uid}`"
  class="header-anchor"
  @click="handleFocusAndScroll(anchor)"
>
  <slot />
  <span hidden :id="`ax-in-page-link-${_uid}`" >{{ $t('accessibility.in-page-link') }}</span>
  <LinkIcon class="icon" aria-hidden="true" />
</router-link>
```

It would be even better if we apply
```vue
 <span hidden id="ax-in-page-link" >{{ $t('accessibility.in-page-link') }}</span>
 ```
only once in the document, so we don't repeat that HTML node every time and we just make a reference to it.
Note: I used `_uid` in the one that it's inside the component because ids need to be unique.

mportiz08: Thanks for all the help and ideas everyone!

I decided to go with the pseudo element approach for now, since that is what I was most familiar with from working on the strikethrough implementation before, but I'd be open to exploring the other approach in the future and seeing if it makes sense to refactor other usages of "visuallyhidden" with similar copy/paste issues.

mportiz08:
@swift-ci test

src/components/ContentNode/LinkableHeading.vue
hqhhuang: > > Looks like we were using aria-labels for accessibility for the link, but changed it here. I don't completely understand what was the issue with the aria-label in the original implementation? Is it just preventing scrolling to the correct position?
> 
> @hqhhuang with the aria-label we were replacing the content of the headers for screen readers, so unfortunately it's not an option :/

Could we resolve this by including the content of the header as part of the aria-label? Like setting "In page link: Overview" as a whole to be the aria label?

hqhhuang (approved):
Tested in Safari and Chrome, LGTM!

dobromir-hristov (approved):
LGTM

src/components/ContentNode/LinkableHeading.vue
marinaaisa: > Could we resolve this by including the content of the header as part of the aria-label? Like setting "In page link: Overview" as a whole to be the aria label?

@hqhhuang The issue is that we use a `slot` for the content of the header so it's not a prop that we can attach in the aria-label. That's why I wrote a solution before using `aria-labelledby` instead, because it allow us to use the ids to reference the content inside the slot.

marinaaisa (approved):
LGTM! :)

PR #729, opened 2023-09-12
Enhancing User Experience with Smooth Scrolling 📜

sanjaiyan-dev:
This PR adds `scroll-behavior: smooth` to provide a smoother and more user-friendly scrolling experience, enhancing navigation, aesthetics, and engagement while maintaining compatibility with modern browsers.

Extremely sorry, if I made any mistakes :(

### DIFF
--- a/src/styles/base/_reset.scss
+++ b/src/styles/base/_reset.scss
@@ -18,6 +18,7 @@ html {
   // enough that it does not actually exceed the breakpoint height.
   background-color: var(--colors-text-background, var(--color-text-background));
   height: 100%;
+  scroll-behavior: smooth;
 }
 
 //============================================================


mportiz08 (changes requested):
Unfortunately, I'm finding that this change has a negative impact on the experience of scrolling to linked subsections within example article content.

Example:

https://github.com/apple/swift-docc-render/assets/212918/f04c2f07-5284-4274-aa0e-340f742bce08

Notice how the page is no longer scrolling so that the linked subsection appears at the top of the page (under the nav bar).

You can compare this with the expected behavior on main:

https://github.com/apple/swift-docc-render/assets/212918/5657fad1-790b-49d8-a435-b2ee00263c7c

I'm not sure _why_ this is happening with your simple change, but there must be something with our usage of the scrolling APIs that is messing with the timing/location or something when combined with this CSS addition.

If you're able to resolve this, I'd be happy to re-consider this change—smooth scrolling in general sounds like a good addition if we can get it working properly. Thanks for the contribution!

sanjaiyan-dev:
> Unfortunately, I'm finding that this change has a negative impact on the experience of scrolling to linked subsections within example article content.
> 
> Example:
> 
> https://github.com/apple/swift-docc-render/assets/212918/f04c2f07-5284-4274-aa0e-340f742bce08
> 
> Notice how the page is no longer scrolling so that the linked subsection appears at the top of the page (under the nav bar).
> 
> You can compare this with the expected behavior on main:
> 
> https://github.com/apple/swift-docc-render/assets/212918/5657fad1-790b-49d8-a435-b2ee00263c7c
> 
> I'm not sure _why_ this is happening with your simple change, but there must be something with our usage of the scrolling APIs that is messing with the timing/location or something when combined with this CSS addition.
> 
> If you're able to resolve this, I'd be happy to re-consider this change—smooth scrolling in general sounds like a good addition if we can get it working properly. Thanks for the contribution!
> 

Sorry, I will try to resolve this issue (not sure) 🤞🏽

sanjaiyan-dev:
> Unfortunately, I'm finding that this change has a negative impact on the experience of scrolling to linked subsections within example article content.
> 
> Example:
> 
>  sanjaiyan-improve-css.mov 
> Notice how the page is no longer scrolling so that the linked subsection appears at the top of the page (under the nav bar).
> 
> You can compare this with the expected behavior on main:
> 
>  main.mov 
> I'm not sure _why_ this is happening with your simple change, but there must be something with our usage of the scrolling APIs that is messing with the timing/location or something when combined with this CSS addition.
> 
> If you're able to resolve this, I'd be happy to re-consider this change—smooth scrolling in general sounds like a good addition if we can get it working properly. Thanks for the contribution!

Hi, I tried to implement the smooth scrolling using JavsScript, hope it works 🤞🏽

mportiz08 (changes requested):
Nice—seems to be working now!

There are a number of failing unit tests to look into now, but I think you should be able to easily update them based on the changes you made. You can see them with `npm test`.

src/utils/scroll-lock.js
```diff
@@ -160,7 +160,11 @@ export default {
       document.body.style.removeProperty('width');
 
       // restore scrolled Y position after resetting the position property
-      window.scrollTo(0, Math.abs(scrolledClientY));
+      window.scrollTo({
+        top: Math.abs(scrolledClientY),
+        left: 0,
+        behavior: 'smooth',
```
dobromir-hristov, line 166: Scroll lock should be instant, this is more of a "hack" to return the scrollbar back to where it was, before locking the scroll.

src/components/Navigator/NavigatorCard.vue
```diff
@@ -933,6 +933,7 @@ export default {
       this.$refs.scroller.$el.scrollBy({
         top: offsetHeight * positionIndex,
         left: 0,
+        behavior: 'smooth',
```
dobromir-hristov, line 936: I think this should not have smooth scrolling. Its causing weird jumps as we are sampling the item if its in the viewport or not, too fast.

dobromir-hristov (changes requested):
On Chrome this is not behaving well at all.

In general, have to do a refactor on the way we handle scrolling. We sometimes call a scroll and then after a frame or two nudge by an extra few pixels to compensate for the sticky navbar for example. This is not visible when instant scrolling, but very much so when smooth scrolling.

This causes jumping and stuttering, especially on Chrome which is not OK.

Things I tried:

1. Clicking on headings
2. Refreshing after clicking the headings
3. Clicking on items in the OnThisPage right floating nav

sanjaiyan-dev:
> On Chrome this is not behaving well at all.
> 
> In general, have to do a refactor on the way we handle scrolling. We sometimes call a scroll and then after a frame or two nudge by an extra few pixels to compensate for the sticky navbar for example. This is not visible when instant scrolling, but very much so when smooth scrolling.
> 
> This causes jumping and stuttering, especially on Chrome which is not OK.
> 
> Things I tried:
> 
> 1. Clicking on headings
> 2. Refreshing after clicking the headings
> 3. Clicking on items in the OnThisPage right floating nav

Hi @dobromir-hristov , I made the your suggested changes :) , Hope it works correctly now  !

dobromir-hristov:
> Hi @dobromir-hristov , I made the your suggested changes :) , Hope it works correctly now !

Yes those are better, yet my last point still stands. When you land on a page with an anchor it does not smoothly scroll to that part of the page, but rather it starts jumping up/down.

Here is an example:

https://github.com/apple/swift-docc-render/assets/9863944/8f7fa61e-bbdf-44a9-80d1-38910fb2111e

PR #733, opened 2023-09-19
Underline inline links

marinaaisa:
Bug/issue #115116961, if applicable: 

## Summary

Inline links need to be underline under WCAG 2.2 compliance
More info: https://www.w3.org/WAI/WCAG22/Techniques/failures/F73

## Dependencies

NA

## Testing

Steps:
1. Go to a page with inline links
2. Assert that they are underline
3. Go to their hover and focus states and assert that underline thickness is greater than its normal state

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -449,7 +449,12 @@ function renderNode(createElement, references) {
     }
     case InlineType.link:
       // Note: `InlineType.link` has been deprecated, but may still be found in old JSON.
-      return createElement('a', { attrs: { href: node.destination } }, (
+      return createElement('a', {
+        attrs: { href: node.destination },
+        // Inline links need to be underline under WCAG 2.2 compliance
+        // More info: https://www.w3.org/WAI/WCAG22/Techniques/failures/F73
+        class: 'underline',
+      }, (
         node.title
       ));
     case InlineType.reference: {
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -451,9 +451,7 @@ function renderNode(createElement, references) {
       // Note: `InlineType.link` has been deprecated, but may still be found in old JSON.
       return createElement('a', {
         attrs: { href: node.destination },
-        // Inline links need to be underline under WCAG 2.2 compliance
-        // More info: https://www.w3.org/WAI/WCAG22/Techniques/failures/F73
-        class: 'underline',
+        class: 'inline-link',
       }, (
         node.title
       ));
@@ -473,9 +471,7 @@ function renderNode(createElement, references) {
           titleStyle: reference.titleStyle,
           hasInlineFormatting: !!titleInlineContent,
         },
-        // Inline links need to be underline under WCAG 2.2 compliance
-        // More info: https://www.w3.org/WAI/WCAG22/Techniques/failures/F73
-        class: 'underline',
+        class: 'inline-link',
       }, (
         titleInlineContent ? renderChildren(titleInlineContent) : titlePlainText
       ));
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -468,6 +468,9 @@ function renderNode(createElement, references) {
           titleStyle: reference.titleStyle,
           hasInlineFormatting: !!titleInlineContent,
         },
+        // Inline links need to be underline under WCAG 2.2 compliance
+        // More info: https://www.w3.org/WAI/WCAG22/Techniques/failures/F73
+        class: 'underline',
       }, (
         titleInlineContent ? renderChildren(titleInlineContent) : titlePlainText
       ));
--- a/src/components/Navigator/BaseNavigatorCard.vue
+++ b/src/components/Navigator/BaseNavigatorCard.vue
@@ -146,7 +146,6 @@ $close-icon-padding: 5px;
     align-items: center;
     height: $nav-height;
     white-space: nowrap;
-    text-decoration: none;
 
     .card-link {
       color: var(--color-text);
--- a/src/components/Navigator/BaseNavigatorCard.vue
+++ b/src/components/Navigator/BaseNavigatorCard.vue
@@ -14,15 +14,6 @@
       <div class="navigator-card-inner">
         <div class="head-wrapper">
           <div class="head-inner">
-            <button
-              :id="SIDEBAR_HIDE_BUTTON_ID"
-              class="close-card"
-              :class="{ 'hide-on-large': !allowHiding }"
-              :aria-label="$t('navigator.close-navigator')"
-              @click="handleHideClick"
-            >
-              <SidenavIcon class="icon-inline close-icon" />
-            </button>
             <Reference
               :id="INDEX_ROOT_KEY"
               :url="technologyPath"
@@ -34,6 +25,15 @@
               </h2>
               <Badge v-if="isTechnologyBeta" variant="beta" />
             </Reference>
+            <button
+              :id="SIDEBAR_HIDE_BUTTON_ID"
+              class="close-card"
+              :class="{ 'hide-on-large': !allowHiding }"
+              :aria-label="$t('navigator.close-navigator')"
+              @click="handleHideClick"
+            >
+              <SidenavIcon class="icon-inline close-icon" />
+            </button>
           </div>
         </div>
         <slot name="body" className="card-body"/>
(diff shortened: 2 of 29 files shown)


src/components/ContentNode.vue
```diff
@@ -468,6 +468,9 @@ function renderNode(createElement, references) {
           titleStyle: reference.titleStyle,
           hasInlineFormatting: !!titleInlineContent,
         },
+        // Inline links need to be underline under WCAG 2.2 compliance
+        // More info: https://www.w3.org/WAI/WCAG22/Techniques/failures/F73
+        class: 'underline',
```
mportiz08, line 473: Could we just add a `:deep(a)` rule for any anchors within a `ContentNode` since we always want to apply the underline for links here? That would also catch the missed `InlineType.link` case from above which I believe you would also want to apply an underline.

mportiz08 (commented):
Haven't tried it yet, but it might also be interesting to do the calculation in the lightness difference between our black/gray plaintext and blue links to see if an underline or other decoration is strictly necessary (probably is).

marinaaisa:
> Haven't tried it yet, but it might also be interesting to do the calculation in the lightness difference between our black/gray plaintext and blue links to see if an underline or other decoration is strictly necessary (probably is).

@mportiz08 I investigated a bit more and I found that our dark mode doesn't have enough lightness contrast:
Light mode: link text 60%, normal text 0% -> ratio of lightness contrast 1201:1 ✅
Dark mode: link text 50%, normal text 100% -> ratio of lightness contrast 1:2 🚫

To pass, the lightness contrast needs to be greater than 3:1, based on https://www.w3.org/WAI/WCAG22/Techniques/failures/F73

Note: to calculate contrast ratio use: Contrast Ratio (CR) = (L1 + 0.05) / (L2 + 0.05)

One option is to apply underline only to dark mode. Would that be better?

marinaaisa:
> Haven't tried it yet, but it might also be interesting to do the calculation in the lightness difference between our black/gray plaintext and blue links to see if an underline or other decoration is strictly necessary (probably is).

@mportiz08 I investigated a bit more and I found that we have enough lightness difference (which it's the same as contrast):
Light mode: link text #3366ff, normal text #000000 -> contrast ratio 4.49:1 ✅
Dark mode: link text #0099ff, normal text #ffffff -> contrast ratio 3.00:1 ✅

To pass, the contrast needs to be greater than 3:1, based on https://www.w3.org/WAI/WCAG22/Techniques/failures/F73

Note: to calculate contrast ratio I use this tool https://colourcontrast.cc/

src/styles/base/_typography.scss
```diff
@@ -137,10 +137,16 @@ a {
...
   }
 
-  &:hover {
+  &:hover, &:focus, &.underline {
     text-decoration: underline;
   }
 
+  &.underline {
+    &:hover, &:focus {
+      text-decoration-thickness: 2px;
```
hqhhuang, line 146: I tested this and the change in thickness is a little jarring to me. 
I looked in the accessibility guide and can't seem to find information on how to handle hover state of a link that's already underlined. But I found that [mdn docs ]( https://developer.mozilla.org/en-US/docs/Learn/CSS/First_steps)disable the underline on hover, which is pretty interesting and maybe worth trying?

src/components/ContentNode.vue
marinaaisa: https://github.com/apple/swift-docc-render/pull/733/commits/8f20497b471af7444639cadc23888eaaa8f93665

dobromir-hristov (commented):
Did a quick test and things look alright. Left a few small comments in the issue

hqhhuang (approved):
LGTM

hqhhuang:
Actually, could you see if we could move the underline slightly lower? It's currently covering up the existing underline in the link.

marinaaisa:
> Actually, could you see if we could move the underline slightly lower? It's currently covering up the existing underline in the link.

My latest change should fix it, could you review again @hqhhuang ? thanks!

src/styles/base/_typography.scss
```diff
@@ -137,8 +137,10 @@ a {
     text-decoration: none;
   }
 
-  &:hover {
-    text-decoration: underline;
+  // Inline links need to be underline under WCAG 2.2 compliance
+  // More info: https://www.w3.org/WAI/WCAG22/Techniques/failures/F73
+  &:hover, &:focus, &.inline-link {
```
hqhhuang, line 142: I think `:focus` affects the navigator, which I think is unintended?

hqhhuang (changes requested):

src/styles/base/_typography.scss
marinaaisa: Good catch! I've been thinking about what to do with the Navigator links and I believe in this case it's better to keep them underline to differentiate from group maker items.

I reviewed other elements on the UI and I prevented underline in OnThisPage, Navigator Head and Skip Nav. Also I fixed the focus style for Navigator Head. https://github.com/apple/swift-docc-render/pull/733/commits/f756425094147510469b5471580f756d7066d7ee

Can you review again? Thanks

marinaaisa:
@swift-ci test

hqhhuang (approved):
Looks great!!

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -10,7 +10,6 @@
 
 <template>
   <BaseNavigatorCardItem
-    class="navigator-card-item"
```
hqhhuang, line 13: Why are we removing this?

hqhhuang (approved):
Looks great!! I like the keyboard focus mimicking the hover behavior

src/components/Navigator/BaseNavigatorCard.vue
```diff
@@ -34,6 +25,15 @@
               </h2>
               <Badge v-if="isTechnologyBeta" variant="beta" />
             </Reference>
+            <button
```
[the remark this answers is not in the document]
dobromir-hristov: This no, but its probably better when you navigate across stuff. Right @marinaaisa ?

dobromir-hristov (commented):
Here is what I found, not sure all of them are actual issues

1. Instances of CallToActionButton are getting underlines on focus. This is visible in the Tutorials
2. Navigator links without without underline on hover might be a problem. Check with @SamLanier, we recently had this discussion. 
3. Tutorial Overview left OnThisPage nav gets underline on Focus. Is that expected?
4. The "ViewMore" in Tutorial Overview resources also gets an underline, but the items above it dont.
5. On Focus, the Time Estimates on TutorialOverview also get underlined 😱 
6. For the Documentation breadcrumbs, we should make sure the `.nav-title-link` also gets the same treatment as the rest of the breadcrumbs.
7.

src/components/Navigator/NavigatorCardItem.vue
marinaaisa: The class is duplicated on BaseNavigatorCardItem that's why I deleted it

src/components/Navigator/BaseNavigatorCard.vue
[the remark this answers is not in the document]
marinaaisa: Yep, it's to have a better keyboard navigation experience from left to right.

marinaaisa:
> Here is what I found, not sure all of them are actual issues
> 
> 1. Instances of CallToActionButton are getting underlines on focus. This is visible in the Tutorials
> 2. Navigator links without without underline on hover might be a problem. Check with @SamLanier, we recently had this discussion.
> 3. Tutorial Overview left OnThisPage nav gets underline on Focus. Is that expected?
> 4. The "ViewMore" in Tutorial Overview resources also gets an underline, but the items above it dont.
> 5. On Focus, the Time Estimates on TutorialOverview also get underlined 😱
> 6. For the Documentation breadcrumbs, we should make sure the `.nav-title-link` also gets the same treatment as the rest of the breadcrumbs.

@dobromir-hristov After seeing so many issues with adding underline to general focus state, I decided to delete it and leave it only for inline links as the original spec. https://github.com/apple/swift-docc-render/pull/733/commits/8d51bca45c7f8eeea145d20b5ed7783107b048a3
Can you check it again? I'm only keeping the hover state in focus state for navigator links.

I'll check with @SamLanier about the Navigator links on hover state, but it seems material for a new radar since the focus state is not involved.

dobromir-hristov (approved):
Yes, now I dont see those issues. I think we are good here.

marinaaisa:
@swift-ci test

PR #746, opened 2023-10-26
Start documenting components with DocC using in-source comments

mportiz08:
Bug/issue #, if applicable: 117215034

## Summary

This adds a script to generate a "symbol graph" JSON file, which can be used to document Swift-DocC-Render using Swift-DocC itself.

Specifically, this script utilizes [tree-sitter][tree-sitter] parsers to inspect each `.vue` file in the `src/components` directory and emit a "symbol" node in the graph corresponding to each Vue component. By default, a title and declaration are created for every component. If a component defines its props, a "Parameters" section will be automatically generated to list the type and optionality of each prop. Additionally, you can add abstract/overview/parameter documentation by using a multi-line comment immediately preceding the export statement for the component, using the normal DocC [structure][docc-structure] and [formatting][docc-formatting].

With the following command, the symbol graph is generated as part of invoking Swift-DocC to build the documentation, and it can be previewed at http://localhost:8000/documentation/swiftdoccrender.

```shell
npm run docs
```

Now, this will include pages for every component in addition to the articles in `SwiftDocCRender.docc`. The existing catalog can also be used to document components using [extension files][docc-extension-files] like normal Swift symbols. As an example, the `Reference` component is documented using a mix of an extension file and its in-source documentation comment.

**Example screenshot:**
Example screenshot of component documentation page

This example page is generated from the in-source documentation for the [`WordBreak`][component-word-break] component.

### Future directions and other notes

I think the ability to document Swift-DocC-Render using Swift-DocC is something that should motivate us to write better documentation for components going forward.

I have done very minimal documentation/curation of existing components in this PR, but I think this is something that will naturally grow and evolve over time as we work on the app.

I think it makes sense to utilize a GitHub action to eventually publish the output of `npm run docs:build` as a GitHub Pages website for this repo and have it automatically update as documentation changes.

The way tree-sitter is used here is something that could potentially be done in a more generic way that allows for documenting code in other languages and not just Vue projects in a future world, but there will likely need to be changes made to the symbolgraph format to make it work in a more language-agnostic way.

[tree-sitter]: https://tree-sitter.github.io/tree-sitter/
[docc-structure]: https://www.swift.org/documentation/docc/writing-symbol-documentation-in-your-source-files
[docc-formatting]: https://www.swift.org/documentation/docc/formatting-your-documentation-content
[docc-extension-files]: https://www.swift.org/documentation/docc/adding-supplemental-content-to-a-documentation-catalog#Add-Extension-Files-to-Append-to-or-Override-Source-Documentation-Comments
[component-word-break]: https://github.com/mportiz08/swift-docc-render/blob/9680b2170951e682807ec3ab785d52da62544369/src/components/WordBreak.vue#L12

## Dependencies

To run the `npm run docs` command, you will need to have a version of Swift-DocC installed and available on your system, either from Xcode or a Swift toolchain.

## Testing

Steps:
1. Run `npm run docs` and preview the docs
2. Verify that you can make changes to component comments and/or markdown files and run `npm run docs` again to see those changes reflected in the resulting web pages.

## Checklist

- [ ] Added tests


SwiftDocCRender.docc/SwiftDocCRender.md
```diff
@@ -1,8 +1,4 @@
-# Swift-DocC-Render
-
-@Metadata {
```
dobromir-hristov, line 3: This is no longer needed?

dobromir-hristov (commented):
From a brief look over things, this is pretty well made, nice job :)

Are we keeping the rundocc bin file?

SwiftDocCRender.docc/SwiftDocCRender.md
mportiz08: Not strictly necessary anymore, although we could add it back if we wanted to change the main display name.

mportiz08:
> From a brief look over things, this is pretty well made, nice job :)
> 
> Are we keeping the rundocc bin file?

It's still being used in `bin/build-docs`.

marinaaisa (approved):
Very interesting how you used async generator functions for this case and the parser of tree sitter. 👏  Amazing job!

I found a minor detail: when sloth is set 2 times in the context of a `if` `else` conditional template, `<slot />`  gets rendered in the documentation 2 times, which doesn't make much sense. It should only be one.

Let me know if you prefer to work on this in the PR or in a new radar. 
Example: http://localhost:8000/documentation/swiftdoccrender/navmenulink

mportiz08:
> > I found a minor detail: when sloth is set 2 times in the context of a `v-if` `v-else` conditional template, `<slot />` gets rendered in the documentation 2 times, which doesn't make much sense. It should only be one.
> > Let me know if you prefer to work on this in the PR or in a new radar. Example: http://localhost:8000/documentation/swiftdoccrender/navmenulink
> 
> Nice catch! Yeah, that's definitely a bug—I'll look into it, but I expect it might be tricky to fix, so maybe it will be good to consider a separate PR in the future.

Actually, I was overthinking the complexity of this. Since you can only provide unique slots as the user of a component, it's simple enough to just keep track of the unique slots to fix this issue. Fixed with 41b4272e279ecd9010ccd67fa2971a637df7284b :)

I also added a note about one known area where we _aren't_ showing a slot in the declaration when we should (ones that are expressed in a render function instead of a template). Addressing that is complex enough that I think we should look at it later though. 8ed373de472ad5127e5c81117f3e182fcd1b8bfe

marinaaisa (dismissed):
> Actually, I was overthinking the complexity of this. Since you can only provide unique slots as the user of a component, it's simple enough to just keep track of the unique slots to fix this issue. Fixed with https://github.com/apple/swift-docc-render/commit/41b4272e279ecd9010ccd67fa2971a637df7284b :)

Nice!

> I also added a note about one known area where we aren't showing a slot in the declaration when we should (ones that are expressed in a render function instead of a template). Addressing that is complex enough that I think we should look at it later though. https://github.com/apple/swift-docc-render/commit/8ed373de472ad5127e5c81117f3e182fcd1b8bfe

It totally seems more difficult to solve.

I noticed a bigger issue when components are within a folder which name doesn't match any component name, their docs don't get rendered.
For example, all components inside `Filter` folder: `FilterInput`, `Tag`, `TagList` aren't in the docs because of this issue.

mportiz08:
> I noticed a bigger issue when components are within a folder which name doesn't match any component name, their docs don't get rendered. For example, all components inside `Filter` folder: `FilterInput`, `Tag`, `TagList` aren't in the docs because of this issue.

I think this might just be a limitation of the way that DocC builds the index that is used for populating the sidebar—since those components have "Filter" as part of their path components but don't have a page for "Filter" (since no component), then they _are_ emitted in the symbol graph JSON and _do_ get HTML pages, but they don't show up in the navigator since we're not able to place them in the tree.

For example, you should be able to them if you access the following URLs but they won't show up as an item in the sidebar:
* http://localhost:8000/documentation/swiftdoccrender/filter/filterinput/
* http://localhost:8000/documentation/swiftdoccrender/filter/tag/
* http://localhost:8000/documentation/swiftdoccrender/filter/taglist/

Not sure yet but maybe I could try to workaround this by tweaking the path components in the symbol graph for these situations or something.

\cc @franklinsch

mportiz08:
> Not sure yet but maybe I could try to workaround this by tweaking the path components in the symbol graph for these situations or something.

@marinaaisa I believe I have resolved this. I updated the `pathComponents` for each symbol to only include items that map to real components, which makes sense in terms of the path hierarchy since we don't have a symbol or page for "Filter" in this scenario. This has a big impact on all of the icons, which also don't have an `Icons` component, so they all turn into top-level components, which matches reality since the directory is just for our own convenience.

In the future, we can add markdown pages to better curate things, but now all of the components that have pages should show up in the navigator.

mportiz08:
The URLs will also change based on the path components change, so `FilterInput` would be a top-level /documentation/swiftdoccrender/filterinput, etc

dobromir-hristov (approved):

mportiz08:
@swift-ci test

PR #750, opened 2023-11-06
Remove rounding from device frame calculations

dobromir-hristov:
Bug/issue #, if applicable: 115480995

## Summary

Removes the rounding from the DeviceFrames screen calcs, because they are causing weird rounding issues in browsers, while resizing. 

The percentages are never visually rendered, so it does not matter how many decimals there are.

## Dependencies

NA

## Testing

Steps:
1. Load a page with device frames
2. Ensure frames from leak background colors, are off positioned, etc.

## Checklist

- [x] Added tests


marinaaisa (approved):

dobromir-hristov:
@swift-ci test

PR #751, opened 2023-11-07
Decorated topic titles links aren't visually identified as links because its content has different colors

marinaaisa:
Bug/issue #116521488, if applicable: 

## Summary
Decorated topic titles links aren't visually identified as links because its content has different colors.
The reason is that the link text is code and by decorating it with code highlighting we interfere in how the element is identified as a link. This is AX problem and I solved it by underlining the link

### Before:
Screenshot 2023-11-07 at 19 24 29

### After:
Screenshot 2023-11-07 at 19 22 40

## Dependencies

NA

## Testing

Steps:
1. Go to any documentation page with decorated titles
2. Assert that they are underlined 

## Checklist

- [x] Added tests


SamLanier:
@marinaaisa this underline looks to be too far away from the text. I believe it should be closer to the text like our current underline-on-hover style is. 

Also, did we investigate making the full title link-blue instead? I think the underline is distracting, and when we link to symbols inline, the full link is blue (not gray and blue like the Topics title shown here).

marinaaisa:
@SamLanier thanks for the review! I add some comments:

> @marinaaisa this underline looks to be too far away from the text. I believe it should be closer to the text like our current underline-on-hover style is.

Underline is a bit further away to prevent underscore characters `_` to be covered by the underline.
This was implemented in this [previous PR](https://github.com/apple/swift-docc-render/pull/733#issuecomment-1772114219) but we can look into it if it's not the best approach to fix that issue.

> Also, did we investigate making the full title link-blue instead? I think the underline is distracting, and when we link to symbols inline, the full link is blue (not gray and blue like the Topics title shown here).

The issue with having the full link in blue is that we would be missing the current coding highlighting we have on the links.
Example of the current coding highlighting:
Screenshot 2023-11-08 at 16 03 45

mportiz08:
> > Also, did we investigate making the full title link-blue instead? I think the underline is distracting, and when we link to symbols inline, the full link is blue (not gray and blue like the Topics title shown here).
> 
> The issue with having the full link in blue is that we would be missing the current coding highlighting we have on the links. Example of the current coding highlighting: Screenshot 2023-11-08 at 16 03 45

Specifically, the gray text is additional decoration to add context to the symbol name in blue, which is helpful for distinguishing overloads that may share the same symbol name (as one example scenario).

For example:
```
// these links
init(foo: Int, bar: Int)
init(foo: Int, bar: String)

// both have the same name (blue text)
init(foo:bar)
```

The blue part is the official symbol name and the gray decoration text helps add additional context that may be useful at a glance.

SamLanier (approved):
Design approved. 

Note: the distance between the line and the text should still be adjusted.

mportiz08 (approved):
The code looks fine on a technical level.

On a more subjective level, I don't think the default underlines look very appealing since it looks like the hover state is stuck, and it's not clear to me why we would add this for decorated symbol links and not other links like articles, etc.

marinaaisa:
@swift-ci test

PR #760, opened 2023-11-30
Ensure dark colors are always used in tutorial hero

mportiz08:
Bug/issue #, if applicable: 

## Summary

Since this component is meant to always be rendered with a dark color scheme regardless of the user chosen setting, there may be some elements (like asides) that may look broken in light mode since they are trying to use light color scheme colors on the dark background.

### Example

**Before**
screenshot showing the issue on the main branch

**After**
screenshot showing the resolved issue on this branch

## Testing

Steps:
1. Create or update a tutorial with a warning aside in the tutorial intro section
2. Using this branch, open that tutorial page and verify that the warning aside uses the dark colors, _even when you choose the "Light" color scheme option_
3. Ensure there are no regressions with other colors in this element

## Checklist

### DIFF
--- a/src/components/Tutorial/Hero.vue
+++ b/src/components/Tutorial/Hero.vue
@@ -193,6 +193,20 @@ export default {
   background-color: var(--color-tutorial-hero-background);
   color: var(--color-tutorial-hero-text);
   position: relative;
+
+  &.dark {
+    @media screen {
+      // ensure dark colors are always used, regardless of the selected
+      // light/dark/auto color scheme preference
+      //
+      // unfortunately the order of the property declaration matters here due
+      // to the way that some properties refer to others, which is why both the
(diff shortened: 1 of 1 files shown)


dobromir-hristov (approved):
A bit barbaric 🤣  but we dont have the notion of "themed" components, where we can toggle the light or dark theme on demand, so this will suffice.

mportiz08:
> A bit barbaric 🤣 but we dont have the notion of "themed" components, where we can toggle the light or dark theme on demand, so this will suffice.

Yeah, I kind of hate it lol, but I don't know of a better alternative without explicitly overriding specific colors...which would mean we would end up missing others at some point. Luckily there doesn't seem to be any noticeable performance impact with just redefining all the vars in this way 😅

mportiz08:
@swift-ci test

PR #762, opened 2023-12-01
Adds dropdown UI to display other declarations

hqhhuang:
Bug/issue #, if applicable: rdar://101257666

## Summary
Adds a dropdown UI in `DocumentationTopic` to display other declarations if there are any.

## Dependencies
DocC: https://github.com/apple/swift-docc/pull/740
Without this dependency, this PR should not change how things behavior and should not introduce regressions.

## Testing
Test manually and with unit test

Steps:
1. Manually test that this PR does not introduce regressions
2. Manually test that the dropdown is collapsed by default, clicking on a pill takes user to the expected page

## Checklist

- [x] Added tests


src/components/DocumentationTopic.vue
```diff
@@ -80,10 +81,14 @@
...
-            v-if="hasPrimaryContent"
+            v-if="hasPrimaryContent || hasOtherDeclarations"
             :class="['container', { 'minimized-container': enableMinimized }]"
           >
-            <div class="description" :class="{ 'after-enhanced-hero': enhanceBackground }">
+            <div
+              class="description"
+              :class="{ 'after-enhanced-hero': enhanceBackground }"
+              v-if="!declListExpanded"
+            >
```
dobromir-hristov, line 91: i'd put the `v-if` on top, as per Vue guidelines

```suggestion
              v-if="!declListExpanded"
              class="description"
              :class="{ 'after-enhanced-hero': enhanceBackground }"
            >
```

dobromir-hristov (commented):
I have pushed a few minor css fixes. Here are a few extra things:

1. I am missing hover effects on the extra declarations
2. I am missing a hover state on the expand button
3. Is this the finalised design for the expand button? I saw some other variants around.

hqhhuang:
I confirmed with @biamx3 that we don't need hover state for the extra declarations and the botton.

> Is this the finalised design for the expand button? 

Yes!

dobromir-hristov (approved):

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -36,6 +61,12 @@ export default {
   name: 'DeclarationGroup',
   components: {
     Source: DeclarationSource,
+    TransitionExpand,
+  },
+  data() {
+    return {
+      selectedIdentifier: this.identifier,
```
marinaaisa, line 68: Do we need the `selectedIdentifier` at all? Why not using the `identifier` directly? Aren't they the same?

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -70,8 +111,35 @@ export default {
       type: String,
       required: false,
     },
+    declListExpanded: {
```
marinaaisa, line 114: Could you change the name to something like "declarationsExpanded" ?
"decl" is a bit difficult to understand

marinaaisa (approved):
Nice job! Left minor comments.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
hqhhuang: This is for the purpose of animating the dropdown.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
hqhhuang: I wanted to make it clear that this is for a list of declarations, but naming this `declarationsListExpanded` is a bit too long. I think `decl` is a pretty common abbreviation.

marinaaisa: sounds good

marinaaisa (approved):

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
hqhhuang: To elaborate a bit: to animate the dropdown collapsing without a jump, we visually select the pill users clicked on first and run the animation before we actually push the new route. So there is a split second where the `selectedIdentifier` is different from the identifier of the page

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -80,6 +147,32 @@ export default {
...
+  },
+  methods: {
+    async selectDeclaration(identifier) {
+      if (identifier === this.identifier) return;
+      this.selectedIdentifier = identifier;
+      // enough time to update the just selected item
+      await waitFor(100);
+      this.isExpanded = false; // collapse the list
+      // await animation to finish
+      await waitFor(500);
```
mportiz08, line 166: Is this to match the duration of a CSS transition? I imagine it's not trivial, but it would be nice if we made were able to make this programmatic in the future instead of hardcoded to match the duration.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -80,6 +147,32 @@ export default {
...
+      this.$router.push(this.references[identifier].url);
```
mportiz08, line 167: I might be overlooking something, but I don't know if this is working as expected. When I click an alternate overload, it does change the selected UI, but it doesn't seem to update the URL and actually navigate to a new page.

Another issue I'm seeing is that I can't navigate back to the original symbol after clicking a subsequent overload. I might just be misunderstanding the expected behavior here, but not sure.

**Example**

https://github.com/apple/swift-docc-render/assets/212918/b4b7eb0e-501f-4b89-a02e-47066634b47f

src/components/DocumentationTopic.vue
```diff
@@ -42,10 +42,10 @@
         >
           <component :is="titleBreakComponent">{{ title }}</component>
           <template #after v-if="isSymbolDeprecated || isSymbolBeta">
-          <small
-          :class="tagName"
-          :data-tag-name="tagName"
-          />
+            <small
```
mportiz08, line 45: Not important but maybe you meant to only indent the prop assignments here?

src/components/DocumentationTopic.vue
```diff
@@ -80,10 +81,14 @@
           :class="{ 'no-primary-content': !hasPrimaryContent && enhanceBackground }"
         >
           <div
-            v-if="hasPrimaryContent"
+            v-if="hasPrimaryContent || hasOtherDeclarations"
```
mportiz08, line 84: Not very important but should having these "other declarations" just be considered in the computation of `hasPrimaryContent` itself?

src/components/DocumentationTopic.vue
```diff
@@ -520,6 +539,13 @@ export default {
     declarations({ primaryContentSections = [] }) {
       return primaryContentSections.filter(({ kind }) => kind === SectionKind.declarations);
     },
+    hasOtherDeclarations({ declarations = [] }) {
+      // there's always only 1 `declaration` at this level
+      return declarations.length ? declarations[0].declarations.some(decl => Object.prototype.hasOwnProperty.call(decl, 'otherDeclarations')) : false;
```
mportiz08, line 544: Since you already have a default empty array, I'm not sure you need the ternary expression here.

src/components/DocumentationTopic.vue
```diff
@@ -520,6 +539,13 @@ export default {
...
+    },
+    declListToggleText({ declListExpanded }) {
+      return declListExpanded ? 'Hide other declarations' : 'Show all declarations';
```
mportiz08, line 547: We should probably use the localized string JSON for this text, even if we don't yet have the non-English translations.

src/components/DocumentationTopic.vue
```diff
@@ -647,6 +676,47 @@ export default {
 <style scoped lang="scss">
 @import 'docc-render/styles/_core.scss';
 
+$space-size: 15px;
+
+.declaration-list-menu {
+  position: relative;
+  margin: 0 !important;
```
mportiz08, line 683: It would be awesome if we could clean this up to not require `!important`.

src/components/DocumentationTopic.vue
```diff
@@ -647,6 +676,47 @@ export default {
...
+  .declaration-list-toggle {
+    display: flex;
+    flex-direction: row;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    background: var(--color-fill); // cover border line
+    padding: 5px 15px;
+    color: var(--colors-link, var(--color-tutorials-overview-link));
```
mportiz08, line 695: Is there a reason why we can't just use `var(--color-link)` here? It seems a little odd to reference a tutorial-specific color here.

src/components/DocumentationTopic.vue
```diff
@@ -647,6 +676,47 @@ export default {
...
+  width: 100%;
+
+  .declaration-list-toggle {
+    display: flex;
+    flex-direction: row;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    background: var(--color-fill); // cover border line
```
mportiz08, line 693: If this is just meant to match the background color of the page, could we just set it to none or transparent?

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -80,6 +147,32 @@ export default {
...
+    },
+    getWrapperComponent(decl) {
+      return (!this.isExpanded || decl.identifier === this.identifier)
+        ? 'div' : 'button';
```
mportiz08, line 171: Not important, but if this is meant to swap between a button and a div for active vs inactive states, it might be worth exploring if we can just use the disabled attribute on the button instead of changing the element.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -102,8 +195,48 @@ export default {
...
+  transition-timing-function: linear;
+  transition-property: opacity, height;
+  $docs-declaration-source-border-width: 1px;
+
+  .source {
+    border-width: $docs-declaration-source-border-width;
+
+    // ensure links are not clickable, when expanded
+    :deep(a) {
+      pointer-events: none;
```
mportiz08, line 208: Which scenario are we trying to prevent links from being clickable? Just curious

mportiz08 (commented):
Nice work!

It's cool to see this working with the corresponding DocC branch + feature flag, and it seems like things are mostly isolated in terms of the default behavior—haven't noticed any visual regressions there yet.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -80,6 +147,32 @@ export default {
...
+        this.$emit('update:declListExpanded', value);
+      },
+    },
+  },
+  methods: {
+    async selectDeclaration(identifier) {
+      if (identifier === this.identifier) return;
+      this.selectedIdentifier = identifier;
+      // enough time to update the just selected item
+      await waitFor(100);
```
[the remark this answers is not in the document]
dobromir-hristov: I couldnt get it to work with a next tick... Probably should investigate again.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
dobromir-hristov: Ideally it should be possible to wait for a transition, its just that `transitionend` is a bit whacky and fires for child transitions as well.

src/components/DocumentationTopic.vue
dobromir-hristov: Accessing `declarations[0].declarations` would be an error.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
dobromir-hristov: We could render it as a disabled button instead, ye 🤔

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
dobromir-hristov: When the whole thing is expanded, you could by accident click on a link to a symbol or something, in the declaration. This prevents that, until you collapse the declarations.

mportiz08: Oh, got it—we want clicking the a declaration in the expanded list to switch the page and then have the normal clicking links within the selected declaration. That makes sense, thanks.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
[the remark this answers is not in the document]
hqhhuang: I just tried waiting on next tick. Let me know what you think.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
hqhhuang: It's longer than the duration of the transition. We use `waitFor` for the sidebar animation as well. Would be nice to make both of these programmatic in the future.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
hqhhuang: > Another issue I'm seeing is that I can't navigate back to the original symbol after clicking a subsequent overload. I might just be misunderstanding the expected behavior here, but not sure.

You can navigate back to the "generic page" with the browser back button or clicking on the navigator. It is expected that if you click on the pill, it will navigate to the "hashed" specific page.

src/components/DocumentationTopic.vue
hqhhuang: I meant to indent the whole thing because it is within the `template`. I'm not sure if we just forgot to indent it originally or it's meant to be on the same level as the `template`.

src/components/DocumentationTopic.vue
hqhhuang: I wanted to track it as an improvement in a separate bug if that's ok?

src/components/DocumentationTopic.vue
hqhhuang: Kept it separate because I still want the dividing line to appear when there are other declarations, but no primary content (for showing the button on the line). So don't want to apply the `no-primary-content` class above.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
hqhhuang: I didn't want all the "regular" declaration pills, which are the majority case, to be rendered as a disabled button. What do you think?

src/components/DocumentationTopic.vue
hqhhuang: Actually I don't even think setting margin here is needed. I removed it.

src/components/DocumentationTopic.vue
mportiz08: For the translations specifically? I think it will make adding those later much easier since it will just be a configuration change at that point.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
mportiz08: I think it's fine as-is considering the button is only meant to be used for pages w/ overloaded symbols and this code impacts all pages, regardless of overloads. Was just curious.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
mportiz08: That sounds like what I was expecting. This works as expected with the fixture data you're using to test, but it's not working that way with the real DocC branch supporting this—you may want to work with @mayaepps to make sure that it's a known bug or incomplete functionality there and not an issue with any of the logic here—seems to be working with the ideal fixture data though. Thanks for clarifying!

src/components/DocumentationTopic.vue
dobromir-hristov: Just move the strings to the lang files, ye. We can localise them later.

hqhhuang:
@swift-ci test

hqhhuang:
@swift-ci test

Issue #765, opened 2023-12-06
Support SSG


pepicrft:
### Feature Name

Support SSG

### Description

We are migrating [Tuist](https://tuist.io)'s documentation to Swift DocC to integrate it more closely into the development workflows (we were using Docusuarus previously). To my surprise, Swift DocC renders client-side, and therefore, search engines, which rely on HTML markup, can't index the content of the documentation, which is not ideal because we'd like the project's documentation to be indexed by Google.

### Motivation

Projects might want their documentation sites to be indexed by search engines. If I'm not wrong, Vue, the UI technology that's being used by Swift DocC, has SSR APIs to generate the markup and output JavaScript that can hydrate the dynamic elements once the page has been loaded. Alternatively, has the team considered using plain HTML with a templating solution and sprinkles of JavaScript?

### Importance

It unlocks making Swift projects' documentation more discoverable, which I think is very desirable for open source maintainers who would love their projects to show up on Google results. 

### Alternatives Considered

I considered other alternatives like Vite Press, Vue Press, or Astro. However, as I said earlier, Tuist contributors love to stay as close as possible to Xcode and Swift, and using those other technologies adds friction.

j-f1:
Hi! I’m actually working on implementing this. My plan is to have a separate script that renders each page in a Node.js environment and then snapshots the HTML once the rendering process is finished (including a few async tasks like syntax highlighting). Ultimately, the output is the same as the `transform-for-static-hosting` `docc` command, except that it fills the content into the generated `index.html` files rather than simply copying a template file.

There are also a couple of improvements that could be made to a few components to better handle the lack of scripting in browsers (for example, the sidebar needs virtual scrolling to work so that would have to be hidden, and the breadcrumbs in the navigation bar would need to be permanently expanded).

I reviewed a bunch of the existing SSR packages for Vue but they seemed to either require upgrading to Vue 3 (which I tried to do but discovered there are a ton of difficult-to-resolve breaking changes, got ~75% of the way through after 10 or so hours) or substantially changing the architecture of how the Vue app works (which I wasn’t comfortable doing).

j-f1:
> Can we provide any support to get this to the finish line?

Take a look at https://github.com/apple/swift-docc-render/pull/766 — it doesn’t have any of the actual prerendering logic but if you spot any issues with the no-js experience there I would love to hear about them :)

[...]

PR #766, opened 2023-12-06
Improve rendering behavior in JS-free environments

j-f1:
Bug/issue #, if applicable: Part of #765

## Summary

This PR adds some `@media (scripting: none)` media queries to some parts of the app to better improve the flow when pre-rendering to a static site, then loading with JS disabled.

Note that the `scripting: none` media query is [not widely supported](https://caniuse.com/mdn-css_at-rules_media_scripting), so it might be wise to swap this with another approach for detecting the presence of JS. The first thing that comes to mind is adding a class like `no-js` to the `<html>` tag, then removing it once JS runs. I’m interested to hear any thoughts on this!

## Dependencies

No formal dependencies.

## Testing

Unfortunately the prerendering script is not yet available, but you can simulate it:

1. Load a doc page up in a browser with JS available
2. Wait for the page to fully load
3. Save the HTML out to the appropriate location in the `.doccarchive` (making sure to name it `index.html`
4. Serve the `doccarchive` as a static site, then load that page with JS disabled.

## Checklist

- [ ] Added tests (I’m not sure there’s a good way to do automated tests for this)
- [x] Updated documentation if necessary (no changes needed)


app/index.html
```diff
@@ -20,7 +20,7 @@
     <script type="application/javascript">var baseUrl = "<%= BASE_URL %>"</script>
 </head>
 <body data-color-scheme="auto">
-<noscript><%= require('@/assets/global-elements/noscript.html') %></noscript>
+<noscript><%= require('@/assets/global-elements/noscript.html').default %></noscript>
```
mportiz08, line 23: Thanks for the `.default` fix! That must be a fairly recent regression since it wasn't necessary in the past 😅

On another note, I assume that your other PR that does the pre-rendering would actually need to remove this `<noscript>` element entirely, right? Otherwise, we end up with something like this where it is included at the top of the page instead of the current experience where it is the full page (without JS enabled).

Screenshot 2023-12-07 at 2 03 24 PM

src/components/ColorSchemeToggle.vue
```diff
@@ -81,6 +81,10 @@ export default {
   @media print {
     display: none;
   }
+
+  @media (scripting: none) {
```
mportiz08, line 85: Unfortunately, it looks like your suspicion is correct about this media feature not being widely supported enough, so we should probably look into an alternate approach for this. The global percentage of support for this is roughly about 8% according to [caniuse](https://caniuse.com/mdn-css_at-rules_media_scripting), which doesn't seem good enough.

Here's my current thinking on a possible alternate approach to solve this without that media query:

1. Update the current `<noscript>` element from the main index file (or add a new one) to include a reference to a new global CSS asset, which will hold global CSS styles that _only_ apply when JS is disabled.
```html
  <!-- this stylesheet is only included when JS is disabled -->
  <noscript><link rel="stylesheet" href="noscript.css" /></noscript>
```
2. Update that JS-disabled `noscript.css` file to include the following css rule:
```css
/* hide any element with the `data-js-required` attribute when JS is disabled */
*[data-js-required] {
  display: none;
}
```
3. For any component that you want to hide when JS is disabled, simply add the `data-js-required` attribute to the top level HTML element in its component.
```html
<template>
  <fieldset
    class="color-scheme-toggle"
    role="radiogroup"
    data-js-required
  >
```

I haven't tried any of this out, but it seems like it might be a fairly clean approach to doing this in a more compatible way.

There are other less clean ways like setting a default `display: none;` to things you want to hide without JS and then unhiding them in a mounted hook or something.

This media query would be the best solution once it's more compatible in the future though!

mportiz08 (changes requested):

app/index.html
j-f1: Yep! It will remove that `<noscript>` — although maybe there still should be a bar at the top of the page that warns folks they might not get the full experience if JS is off.

src/components/ColorSchemeToggle.vue
j-f1: I like that idea! There are a couple of things (like the breadcrumbs) where I want to do something other than hiding, so I might use the `no-js` class approach I mentioned in the intial comment to handle those, but the data attribute is a really cool idea too!

mportiz08: For sure. You could use distinct attributes to handle different types of styling changes, but I'll let you decide which approach fits best.

app/main.js
```diff
@@ -26,3 +26,5 @@ new Vue({
   render: h => h(App),
   i18n: SwiftDocCRenderi18n(),
 }).$mount('#app');
+
+document.documentElement.classList.remove('no-js');
```
mportiz08, line 30: Can we actually move this slightly to be before the `$mount` call above to better guarantee that the vue app never sees this class?

app/index.html
```diff
@@ -9,7 +9,7 @@
 -->
 
 <!DOCTYPE html>
-<html lang="en-US">
+<html lang="en-US" class="no-js">
```
mportiz08, line 12: To test this, I would have to manually add this class back to the rendered HTML that I'm copying to the archive as part of simulating the pre-rendered output. That means that the actual prerender script would need to do that as well, right?

j-f1: Yes!

mportiz08 (approved):

mportiz08:
@swift-ci test

PR #770, opened 2023-12-14
Use localized string format for "other declaration" button

hqhhuang:
**Explanation:** Use localized string format for "other declaration" button.
**Bug/Issue #**: rdar://119615630
**Scope:** Small
**Risk:** No risk, just 1 string config change for other declarations
**Testing:** Updated unit test and manual testing
**Reviewed By:** @mportiz08

## Dependencies
NA

## Testing
Verify the english version of the string is still rendered correctly, follow testing instructions [here](https://github.com/apple/swift-docc-render/pull/762)

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -544,7 +544,8 @@ export default {
       return declarations.length ? declarations[0].declarations.some(decl => Object.prototype.hasOwnProperty.call(decl, 'otherDeclarations')) : false;
     },
     declListToggleText({ declListExpanded }) {
-      return declListExpanded ? 'Hide other declarations' : 'Show all declarations';
+      return declListExpanded ? `${this.$t('verbs.hide')} ${this.$t('declarations.other-declarations')}`
+        : `${this.$t('verbs.show')} ${this.$t('declarations.all-declarations')}`;
     },
   },
   methods: {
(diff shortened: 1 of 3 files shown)


hqhhuang:
@swift-ci test

src/components/DocumentationTopic.vue
```diff
@@ -544,7 +544,8 @@ export default {
       return declarations.length ? declarations[0].declarations.some(decl => Object.prototype.hasOwnProperty.call(decl, 'otherDeclarations')) : false;
     },
     declListToggleText({ declListExpanded }) {
-      return declListExpanded ? 'Hide other declarations' : 'Show all declarations';
+      return declListExpanded ? `${this.$t('verbs.hide')} ${this.$t('declarations.other-declarations')}`
```
mportiz08, line 547: I'm not an expert, but we may want to use a single string for this instead of splitting it up into 2 strings for the noun and verbs and concatenating them. The reason is that we can't guarantee that the verb would always come before the noun in all languages. (@marinaaisa please correct me if this is wrong)

dobromir-hristov: Yep, it should be a single sentence.

hqhhuang:
@swift-ci test

mportiz08 (approved):

eango (approved):

PR #774, opened 2024-01-26
Remove ability to override icons in "Topics" section

mportiz08:
Bug/issue #, if applicable: 120670171

## Summary

Due to limitations with remote SVG assets, there can be an issue where the color of a custom icon may be present in multiple locations using the same color, which results in a less than ideal experience.

Current known workarounds for this issue would necessitate server side headers for cross origin resource sharing, which DocC-Render can't control since it is a client-side only tool—with that in mind, it makes more sense to remove the ability for overriding the icon here and leaving it as something to customize in the hero element only for now.

## Testing

Steps:
1. Add the following snippet to an existing documentation catalog, with the referenced icon asset.
```
@Metadata {
  @PageImage(purpose: icon, source: "example-custom-icon.svg", alt: "icon")
}
```
2. Verify that the custom icon still shows up in the hero element for the page, but it no longer appears in the "Topics" section from a parent page.

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/DocumentationTopic/TopicLinkBlockIcon.vue
+++ b/src/components/DocumentationTopic/TopicLinkBlockIcon.vue
@@ -45,10 +45,6 @@ export default {
       type: String,
       required: true,
     },
-    imageOverride: {
-      type: Object,
-      default: null,
-    },
   },
 
   computed: {
--- a/src/components/DocumentationTopic/TopicLinkBlockIcon.vue
+++ b/src/components/DocumentationTopic/TopicLinkBlockIcon.vue
@@ -9,9 +9,8 @@
 -->
 
 <template>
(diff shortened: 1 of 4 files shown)


src/components/DocumentationTopic/TopicLinkBlockIcon.vue
```diff
@@ -41,7 +39,7 @@ const TopicRoleIcons = {
 };
 
 export default {
-  components: { OverridableAsset, SVGIcon },
+  components: { SVGIcon },
   props: {
```
hqhhuang, line 43: I think we could remove the `imageOverride` prop here and no need to pass it in the `TopicsLinkBlock` component.

mportiz08:
@swift-ci test

PR #776, opened 2024-02-02
Mentioned In primary content section

bitjammer:
Add a new "Mentioned In" section, which renders up to 5 links to articles that mention a symbol by symbol link.

rdar://121899580

- [x] Added tests


bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
```diff
@@ -0,0 +1,56 @@
...
+  Copyright (c) 2024 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <section v-if="topics.length > 0" class="mentions">
+    <h2 class="mentions" id="mentions">Mentioned in</h2>
```
mportiz08, line 13: Also, instead of using the raw `<h2>` here, we should probably use the `LinkableHeading` component, which will give us the behavior where you can click the heading and get a link to the subsection of the page.

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
```diff
@@ -0,0 +1,56 @@
...
+    <Mention v-for="(topic, i) in topics"
+      :key="i"
```
mportiz08, line 15: Small thing, but Vue warns about using index values as "keys" since they are order-dependent. A better key might be the title or URL, which should uniquely identify each child node.

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
```diff
@@ -0,0 +1,56 @@
...
+
+  Copyright (c) 2024 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <section v-if="topics.length > 0" class="mentions">
```
mportiz08, line 12: Minor thing, but you can shorten this to `topics.length` because anything over 0 is considered "truthy" in JavaScript (and 0 is "falsy").

It may also be better to move this conditional to the thing rendering this component instead of the component itself.

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
```diff
@@ -0,0 +1,56 @@
...
+export default {
+  name: 'MentionedIn',
+  components: {
+    Mention,
+  },
+  mixins: [referencesProvider],
+  props: {
+    mentions: {
+      type: Array,
+      required: false,
```
mportiz08, line 37:
```suggestion
      required: true,
```

I think it makes sense for this to be a requirement since there's no need to render this component if there are no mentioned items.

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
```diff
@@ -0,0 +1,56 @@
...
+    },
+  },
+  computed: {
+    topics() {
+      if (this.mentions && this.mentions.length) {
+        return this.mentions
+          .slice(0, 5)
```
mportiz08, line 44: Does the compiler ever provide more than this many items? Just curious if this maximum is only being enforced here or at both levels. (It's fine leaving in either way)

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
```diff
@@ -0,0 +1,56 @@
...
+          .flatMap(mention => this.references[mention]);
+      }
+      return [];
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
```
mportiz08, line 53: You can delete the whole `<style>` block if there's no CSS being added.

src/mixins/onThisPageRegistrator.js
```diff
@@ -57,6 +57,11 @@ export default {
       if (primaryContentSections) {
         primaryContentSections.forEach((section) => {
           switch (section.kind) {
+          case SectionKind.mentions:
+            this.store.addOnThisPageSection({
+              title: 'Mentioned in', anchor: anchorize('mentions'), level: 2,
```
mportiz08, line 62: We should use the localization helper here as well once you add an entry for this string.

src/components/DocumentationTopic/PrimaryContent/Mention.vue
```diff
@@ -0,0 +1,90 @@
...
+<template>
+  <div class="link-block">
+    <Reference
+      :url="url"
+      :role="role"
+      :kind="kind"
+      class="link"
+    >
+      <TopicLinkBlockIcon class="mention-icon" :role="role" />
+      <WordBreak tag="span">{{ title }}</WordBreak>
```
mportiz08, line 19: It's probably safe to leave out the `WordBreak` component since these mentioned in links will always be article titles. The word break helper is more useful when something can contain a long single word, like some symbol names.

tests/unit/components/DocumentationTopic/PrimaryContent/Mentions.spec.js
```diff
@@ -0,0 +1,91 @@
...
+*/
+
+import { shallowMount } from '@vue/test-utils';
+import Mentions from 'docc-render/components/DocumentationTopic/PrimaryContent/Mentions.vue';
+import Mention from 'docc-render/components/DocumentationTopic/PrimaryContent/Mention.vue';
+
+function createReferences(names) {
+  return names.reduce((refs, name) => {
+    const id = 'doc://' + name;
+    refs[id] = {
```
mportiz08, line 18: The linter is likely complaining about this (mutating a param). I don't mind if you just disable that rule for this specific instance since it's not harmful.

Otherwise, you can rewrite the reducer to appease the linter using something like the following pseudo code:
```
=> ({
  ...refs,
  [`doc://${name}`] = {...}
})
```

mportiz08 (commented):
This is looking really good—functionally it seems to be working great using the corresponding DocC branch. Great work! 🤘

I added a bunch of comments, but most of them are small details or minor things that the linter is probably complaining about—it's in great shape!

src/components/DocumentationTopic/PrimaryContent/Mention.vue
```diff
@@ -0,0 +1,90 @@
...
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
+@import 'docc-render/styles/_core.scss';
+
+.abstract,
+.link-block :deep(.badge) {
```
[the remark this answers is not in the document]
bitjammer: Probably a bit of cargo from another component I was looking at. I'll double-check but probably not needed.

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
bitjammer: Okay, might be naive understanding, I was thinking this was necessary to preserve the ordering of the mentions as they come in, since they are pre-scored according to how docc calculated relevance. What you're describing with the keys wouldn't affect the order of presentation, right?

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
bitjammer: The compiler could provide more, so it was decided that this would be handled at the presentation layer. Could change in the future. Thoughts?

src/components/DocumentationTopic/PrimaryContent/Mention.vue
bitjammer: Ah okay, I thought this was affecting word wrapping generally.

bitjammer:
Thanks for the review, @mportiz08! I'll have another pass at it next week.

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
mportiz08: Right, it shouldn't change the order as you're presenting it here.

It's mostly a hint at the uniqueness of individual items to optimize performance of patching the DOM for lists when necessary. In your case, it doesn't technically matter that much since the list is not dynamically updated after being loaded and rendered, but it's still considered a good practice to use something closer to an identifier when possible.

[Maintaining state with `key`](https://vuejs.org/guide/essentials/list.html#maintaining-state-with-key)
[Why not always use the index as the key in a vue.js for loop?](https://stackoverflow.com/questions/44531510/why-not-always-use-the-index-as-the-key-in-a-vue-js-for-loop)

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
mportiz08: Makes sense. I was just curious. Sounds good.

hqhhuang (commented):
I have a generic comment on the current design that includes the "Mentioned In" section as a part of the primary content section. Is there a particular reason for that, as opposed to creating its own top level section, similar to the Default Implementation, Relationships, and See Also section?
I view the primary content section as just the information about the symbol itself. I think this feature is an improvement (or potentially a replacement) of the See Also section. Its main purpose is to provide information on how it is connected to other documentation pages within the technology. 
We may also have design discussions later on about the ordering of the sections. Should it go above the Overview/Discussion (PrimaryContent), or after. Should it be included in minimized mode? It would be easier to make changes to these things if this section is pulled out of the Primary Content section. We can change the ordering without any compiler changes. 
We also don't need to hard code the `Mentioned In` title name in docc-render, and could just use the `title` from the json and the `TopicsTable` to render the content.

bitjammer:
@hqhhuang Agree it will need revisiting in the future, especially as overviews become shorter and explanatory content lies elsewhere, or if mentions come from outside the same bundle. I get that it still seems a bit like a "did you know" section with a heading like that, but it's meant to be much more as time goes on.

As for See Also, I think this could replace it given what most people hope the section accomplishes. At its heart, it's meant to get people back onto a learning path or to read up on the fundamental concepts that the API embodies. In that sense, it's much more important than See Also, maybe even the Overview. Hopefully that makes the placement make sense.

To your last point about API changes, yes, it's really meant to be just "articles" going forward. DocC should ensure that.

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
```diff
@@ -0,0 +1,55 @@
...
+  Copyright (c) 2024 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+<template>
+  <section v-if="topics.length" class="mentions">
+    <LinkableHeading class="mentions" id="mentions">
```
mportiz08, line 13: The linkable behavior of this doesn't seem to be working, but I think this might fix it:

```suggestion
    <LinkableHeading class="mentions" anchor="mentions">
```

mportiz08: You can probably also remove the class, since it's repeating the parent class and doesn't seem to be used in any CSS.

tests/unit/components/DocumentationTopic/PrimaryContent/Mentions.spec.js
```diff
@@ -0,0 +1,91 @@
...
+      propsData: {
+        mentions,
+      },
+      provide: {
+        store: {
+          state: { references },
+        },
+      },
+    });
+    console.log(wrapper.html({ pretty: true }));
```
mportiz08, line 69:
```suggestion
```

src/components/DocumentationTopic/PrimaryContent/Mentions.vue
mportiz08: To test the linkable behavior, you can just click on the "Mentioned in" heading, and it should link directly to that subsection and change the URL to `#mentions`

bitjammer:
@swift-ci Please test

bitjammer:
Need to update my heading test with the removal of the class.

bitjammer:
@swift-ci Please test

PR #780, opened 2024-02-19
NavigatorCard component refactor

marinaaisa:
Bug/issue #120048809, if applicable: 

## Summary

Refactor of NavigatorCard:

- [x] Move the Navigator tags to i18n strings
- [x] Separate filter logic for children into its own mixin
- [x] Separate tags logic into its own mixin

## Dependencies

NA

## Testing

Steps:
1. Assert that everything works well after the refactor

## Checklist

- [x] Added tests


marinaaisa:
@swift-ci test

src/components/Navigator/NavigatorCard.vue
```diff
@@ -98,9 +98,9 @@
           <div class="input-wrapper">
             <FilterInput
               v-model="filter"
-              :tags="availableTags"
+              :tags="suggestedTags"
               :translatableTags="translatableTags"
```
hqhhuang, line 102: Is there a scenario where we would have a tag that's not translatable?

src/mixins/navigator/tagsProvider.js
```diff
@@ -0,0 +1,94 @@
...
+        if (changeType && apiChangesTypesSet.has(changeType)) {
+          availableTags.changes.push(ChangeNames[changeType]);
+          apiChangesTypesSet.delete(changeType);
+        }
+        if (deprecated && generalTags.has(HIDE_DEPRECATED)) {
+          availableTags.other.push(HIDE_DEPRECATED);
+          generalTags.delete(HIDE_DEPRECATED);
+        }
+      }
+      return Object.values(availableTags).flat();
```
hqhhuang, line 91: Like I mentioned in the other comment, I don't think this logic needs to be this complicated (separate into categories, than flatten) if we just included all types of tags within the `FILTER_TAGS` constant.

src/mixins/navigator/filteredChildren.js
```diff
@@ -0,0 +1,59 @@
...
+      // find children that match current filters
+      return children.filter(({
+        title, path, type, deprecated, deprecatedChildrenCount, childUIDs,
+      }) => {
+        // groupMarkers know how many children they have and how many are deprecated
+        const isDeprecated = deprecated || deprecatedChildrenCount === childUIDs.length;
+        // check if `title` matches the pattern, if provided
+        const titleMatch = filterPattern ? filterPattern.test(title) : true;
+        // check if `type` matches any of the selected tags
+        let tagMatch = true;
```
hqhhuang, line 40: Could we simplify this by calling `extractTags`, and check if there's any overlap between `selectedTags`?

hqhhuang (commented):
Looks great! I left small suggestions to simplify the tags categories. Let me know if there's any particular reason we wanted to keep them separate.

src/constants/Tags.js
```diff
@@ -0,0 +1,32 @@
...
+import { TopicTypes } from 'docc-render/constants/TopicTypes';
+
+export const HIDE_DEPRECATED = 'navigator.tags.hide-deprecated';
+
+export const FILTER_TAGS = {
+  sampleCode: 'filter.tags.sample-code',
+  tutorials: 'filter.tags.tutorials',
+  articles: 'filter.tags.articles',
+  webServiceEndpoints: 'filter.tags.web-service-endpoints',
+};
```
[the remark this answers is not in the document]
marinaaisa: `extractTags` is a bit complex logic because there are 3 different types of tags that they don't work in the same way:
- Filter tags: we add them if any symbol has the same type. We should call them "Type tags" instead.
- API Changes tags: we add them when API Changes is turn on and there are tags with the change type
- Deprecated tag: we only add it if there are any deprecated symbols and we don't have API Changes tags

That's why I couldn't really simplify them into a single object, but if you find a way of doing it, please let me know!

src/mixins/navigator/filteredChildren.js
marinaaisa: `extractTags` would just give us an array of `availableTags` that we end up using as suggested tags for the filter. In the `filterChildren` method, we filter the symbols. Could you develop how `extractTags` would help us to simply this function? we always have selected tags.

src/constants/Tags.js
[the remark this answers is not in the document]
hqhhuang: https://github.com/marinaaisa/swift-docc-render/pull/10

I tried it out and wasn't able to simplify it as much as I initially thought. I added a comment to remind future developers that the categories are also created to keep the suggested tags in a particular order.

src/mixins/navigator/tagsProvider.js
hqhhuang: See followups: https://github.com/apple/swift-docc-render/pull/780#discussion_r1512119968

src/mixins/navigator/filteredChildren.js
hqhhuang: I tried it out today. Initially I thought that we could just check if there are overlaps between the `availableTags` of a given node and the `selectedTags` to know if a node should be included in the filter or not. (In this world, I thought we can just add a new function to curate that list of `availableTags` for a node. Both the filtering logic and the tags provider logic can call this helper function)
But after trying that logic I realized that I failed to consider that when the "hide deprecated" tag is available and also selected, that node should be hidden. So looking at whether any overlaps exist is not gonna work.

src/components/Navigator/NavigatorCard.vue
```diff
@@ -524,7 +384,7 @@ export default {
     apiChanges(value) {
       if (value) return;
       // if we remove APIChanges, remove all related tags as well
-      this.selectedTags = this.selectedTags.filter(t => !this.$t(ChangeNames[t]));
+      this.selectedTags = this.selectedTags.filter(t => !ChangeNameToType[t]);
```
hqhhuang, line 387:
```suggestion
      this.selectedTags = this.selectedTags.filter(t => !FILTER_TAGS[t]);
```
Maybe? `selectedTags` should be holding a list of tag types now

hqhhuang: Edit: I was wrong, `selectedTags` holds a list of tag strings. However, there's no need to convert that name to type again when we can just look up whether a tag exists from `FILTER_TAGS`.
Since `ChangeNameToType` is not used anywhere else, I also removed it from `Changes.js`

hqhhuang: Edit: I was wrong, `selectedTags` holds a list of tag names (to then be localized). However, there is no need to convert the names back to type again.

hqhhuang (approved):

hqhhuang:
@swift-ci test

PR #782, opened 2024-02-28
Add generic pager component

mportiz08:
Bug/issue #, if applicable: 123727043

## Summary

This adds a new, generic component called `Pager`, which can be used to interact with multiple horizontal "pages" of content.

For example, this is how it could be used to show a list of paragraphs, with only one being shown at a time—the user can use controls and/or scrolling to switch the currently active paragraph.

```html
<Pager :pages="['a', 'b', 'c']">
  <template #page="{ page }">
    <p>{{ page }}</p>
  </template>
</Pager>
 ```

Additionally, a new `usePager` prop has been added to `TopicsLinkCardGrid` so that this functionality can be opted-in to. We might make the decision to make this the default behavior at some point in the future if it's well received.

### Example

A quick demo with `TopicsLinkCardGrid.usePager` enabled.

demo

## Testing

Steps:
1. Checkout this branch
2. Apply [this patch](https://github.com/apple/swift-docc-render/files/14427351/test.patch) to get some example content and opt-in to the pager for `@Links` (`git apply test.patch`)
3. Build this branch with `npm run build`
4. Build the docs with `DOCC_HTML_DIR=/path/to/swift-docc-render/dist npm run docs` and open http://localhost:8000/documentation/swiftdoccrender in the browser
5. Verify that the pager functionality works as expected
* can scroll/swipe horizontally to navigate through pages
* can use controls in gutter area to navigate through pages
* can use dot indicators to see the currently active page and navigate to others
6. Change `usePager` back to `false` and verify that the grid works the same as it does today and just flows vertically with no options for paging through it horizontally

## Checklist

- [X] Added tests

### CODE REVIEW

**Comment by mportiz08** [comment]
@swift-ci test

File: src/components/Pager.vue [hunk 1]
```diff
@@ -0,0 +1,374 @@
...
+    <template v-else>
+      <div class="container">
+        <Gutter class="left">
+          <ControlPrevious
+            :disabled="!hasPreviousPage"
+            @click.native="previous"
+          />
+        </Gutter>
+        <div class="viewport" ref="viewport">
+          <div
```
marinaaisa, line 24: Pages that are not visible should be hidden, not only visually but from all users (including screen reader users). We should add a class with  `display: none` or the HTML `hidden` attribute.

src/components/Pager.vue
```diff
@@ -0,0 +1,374 @@
...
+          :disabled="!hasPreviousPage"
+          @click.native="previous"
+        />
+        <ControlNext
+          :disabled="!hasNextPage"
+          @click.native="next"
+        />
+      </div>
+      <div class="indicators">
+        <a
```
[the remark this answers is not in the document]
marinaaisa: Links without a text description need to have an aria-label describing what they are for.
For this indicator link, let's add an aria-label saying `Show slide [page] of [total length of pages]`

src/components/Pager.vue
```diff
@@ -0,0 +1,374 @@
...
+    </template>
+    <template v-else>
+      <div class="container">
+        <Gutter class="left">
+          <ControlPrevious
+            :disabled="!hasPreviousPage"
+            @click.native="previous"
+          />
+        </Gutter>
+        <div class="viewport" ref="viewport">
```
marinaaisa, line 23: Screenshot 2024-03-01 at 17 25 43

The focus outline of items gets cut because of the overflow property in this element, we may want to fix that by adding a padding to it.

src/components/PagerControl.vue
```diff
@@ -0,0 +1,70 @@
...
+  This source file is part of the Swift.org open source project
+
+  Copyright (c) 2024 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+<template>
+  <button :class="['pager-control', action]">
```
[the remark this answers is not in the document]
mportiz08: 9dbf15f60fc48837224f8bca658c72ee38a795c8

src/components/Pager.vue
```diff
@@ -0,0 +1,374 @@
...
+          </div>
+        </div>
+        <Gutter class="right">
+          <ControlNext
+            :disabled="!hasNextPage"
+            @click.native="next"
+          />
+        </Gutter>
+      </div>
+      <div class="compact-controls">
```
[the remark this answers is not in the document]
mportiz08: https://github.com/apple/swift-docc-render/pull/782/commits/d64f8c58616f6100e368cc9404e2c477d9fbc4b5

src/components/Pager.vue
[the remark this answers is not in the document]
mportiz08: https://github.com/apple/swift-docc-render/pull/782/commits/d64f8c58616f6100e368cc9404e2c477d9fbc4b5

I added labels for the indicators `aria-current` to the active label, but I don't think it's appropriate to add `aria-disabled` to the inactive indicators, since they can still be used to navigate to those pages and aren't technically "disabled" in that sense.

src/components/Pager.vue
```diff
@@ -0,0 +1,374 @@
...
+  This source file is part of the Swift.org open source project
+
+  Copyright (c) 2024 Apple Inc. and the Swift project authors
+  Licensed under Apache License v2.0 with Runtime Library Exception
+
+  See https://swift.org/LICENSE.txt for license information
+  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+<template>
+  <div class="pager">
```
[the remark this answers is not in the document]
mportiz08: https://github.com/apple/swift-docc-render/pull/782/commits/d64f8c58616f6100e368cc9404e2c477d9fbc4b5

src/components/Pager.vue
```diff
@@ -0,0 +1,374 @@
...
+    <template v-if="pages.length === 1">
+      <slot name="page" :page="pages[0]" />
```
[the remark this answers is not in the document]
mportiz08: https://github.com/apple/swift-docc-render/pull/782/commits/d64f8c58616f6100e368cc9404e2c477d9fbc4b5

src/components/Pager.vue
mportiz08: The structure is purposefully set up to use overflow and visually hiding the overflowing pages to allow for natural scrolling/swiping. Using something like `display: none` would eliminate the functionality of scrolling between pages since we wouldn't actually be using overflow anymore in that scenario.

src/components/Pager.vue
mportiz08: Interesting. Not exactly sure what might be causing this off the top of my head—will need to dig in and try to see if I can see why this happens.

mportiz08:
@swift-ci test

src/components/Pager.vue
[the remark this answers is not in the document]
marinaaisa: I mentioned it because I read it at https://www.smashingmagazine.com/2023/02/guide-building-accessible-carousels/ and I thought it made sense. But it would be okay if we don't add it too, it's a bit controversial in my opinion.

> Do not just indicate visually but also semantically which control does represent the currently displayed slide. You can use aria-disabled="true" for this purpose. In this case, aria-disabled is preferable to the HTML attribute disabled. Unlike disabled, a button with an aria-disabled attribute is still included in the tab sequence of the page. You may also use the [aria-current](https://www.w3.org/TR/wai-aria-1.1/#aria-current) property set to "true".

src/components/Pager.vue
marinaaisa: Yep, I understand the concern. Let's leave it as it is then!

src/components/Pager.vue
marinaaisa: I don't think this is a blocker at all.

mportiz08: Didn't find it yet. Agreed it's not a blocker but would be nice to look into in the future.

mportiz08:
@swift-ci test

src/components/DocumentationTopic/TopicsLinkCardGrid.vue
```diff
@@ -9,43 +9,100 @@
...
+      [TopicSectionsStyle.compactGrid]: {
+        [BreakpointName.large]: 6,
+        [BreakpointName.medium]: 4,
+        [BreakpointName.small]: 1,
+      },
+      [TopicSectionsStyle.detailedGrid]: {
+        [BreakpointName.large]: 4,
+        [BreakpointName.medium]: 2,
+        [BreakpointName.small]: 1,
+      },
```
hqhhuang, line 92: Could we add more detailed documentation here?
For example, `compactGrid` large breakpoint is 2 row by 3 columns; This information is already available, but needs to do a bit math to figure it out.

hqhhuang (approved):

PR #783, opened 2024-02-28
Update Hero Availability & Badge style

hqhhuang:
Bug/issue #, if applicable: 122130081

## Summary
Update style of Availability in the Hero and Badges.

## Dependencies
NA

## Testing
Manual testing, check:
1. badges in the navigator, nav bar, and the availability pill
2. availability pills in the hero section

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Badge.vue
+++ b/src/components/Badge.vue
@@ -60,7 +60,7 @@ export default {
   border-width: var(--badge-border-width, 1px);
   margin: auto;
   margin-left: 5px;
-  color: var(--colors-button-text, var(--color-button-text));
+  color: var(--colors-badge-text, var(--color-badge-text));
   background-color: var(--badge-color);
   @include prefers-dark {
     background-color: var(--badge-dark-color);
--- a/src/components/Badge.vue
+++ b/src/components/Badge.vue
@@ -42,6 +42,7 @@ export default {
 
 <style lang="scss" scoped>
 @import 'docc-render/styles/_core.scss';
+$border-radius: 1px;
 
 @mixin badge-variation($color) {
   --badge-color: var(--color-badge-#{$color});
@@ -53,20 +54,17 @@ export default {
 
   @include font-styles(badge);
   display: inline-block;
-  padding: 2px 10px;
+  padding: 2px 4px;
   white-space: nowrap;
-  background: none;
-  border-radius: var(--badge-border-radius, calc(#{$border-radius} - 1px));
-  border-style: var(--badge-border-style, solid);
-  border-width: var(--badge-border-width, 1px);
-  margin-left: 10px;
-  color: var(--badge-color);
-
-  .theme-dark & {
-    --badge-color: var(--badge-dark-color);
+  border-radius: var(--badge-border-radius, $border-radius);
+  border: none;
+  margin-left: 5px;
+  color: var(--colors-button-text, var(--color-button-text));
+  background-color: var(--badge-color);
+  @include prefers-dark {
+    background-color: var(--badge-dark-color);
   }
 
-  /* nav bar badge */
   &-deprecated {
     @include badge-variation(deprecated);
   }
--- a/src/components/Badge.vue
+++ b/src/components/Badge.vue
@@ -42,7 +42,6 @@ export default {
 
 <style lang="scss" scoped>
 @import 'docc-render/styles/_core.scss';
-$border-radius: 1px;
 
 @mixin badge-variation($color) {
   --badge-color: var(--color-badge-#{$color});
@@ -56,8 +55,9 @@ $border-radius: 1px;
   display: inline-block;
   padding: 2px 4px;
   white-space: nowrap;
-  border-radius: var(--badge-border-radius, $border-radius);
-  border: none;
+  border-radius: var(--badge-border-radius, 1px);
+  border-style: var(--badge-border-style, none);
+  border-width: var(--badge-border-width, 1px);
   margin: auto;
   margin-left: 5px;
   color: var(--colors-button-text, var(--color-button-text));
--- a/src/components/Badge.vue
+++ b/src/components/Badge.vue
@@ -58,6 +58,7 @@ $border-radius: 1px;
   white-space: nowrap;
   border-radius: var(--badge-border-radius, $border-radius);
   border: none;
+  margin: auto;
   margin-left: 5px;
   color: var(--colors-button-text, var(--color-button-text));
   background-color: var(--badge-color);
--- a/src/components/DocumentationTopic/Summary/Availability.vue
+++ b/src/components/DocumentationTopic/Summary/Availability.vue
@@ -30,7 +30,8 @@
         :introducedAt="platform.introducedAt"
         :platformName="platform.name"
       />
-      <Badge v-if=badgeFor(platform) :variant="badgeFor(platform)"> </Badge>
+      <Badge v-if=platform.deprecatedAt variant="deprecated"> </Badge>
+      <Badge v-else-if=platform.beta variant="beta"> </Badge>
     </div>
   </Section>
 </template>
@@ -69,9 +70,6 @@ export default {
     };
   },
   methods: {
-    badgeFor(platform) {
-      return platform.deprecatedAt ? 'deprecated' : (platform.beta && 'beta');
-    },
     changeFor(platform) {
       const { identifier, state: { apiChanges } } = this;
       const { availability = {} } = (apiChanges || {})[identifier] || {};
@@ -125,34 +123,7 @@ export default {
 .tech-icon {
   height: 12px;
   padding-right: 5px;
-  fill: var(--badge-color);
-
-  @include prefers-dark {
-    fill: var(--badge-dark-color);
-  }
-}
-
-.beta, .deprecated {
-  border-radius: 3px;
-  padding: 2px 4px;
-  border: none;
-  color: var(--colors-button-text, var(--color-button-text));
-}
-
-.beta {
-  background-color: var(--color-badge-beta);
-
-  @include prefers-dark {
-    background-color: var(--color-badge-dark-beta);
-  }
-}
-
-.deprecated {
-  background-color:var(--color-badge-deprecated);
-
-  @include prefers-dark {
-    background-color: var(--color-badge-dark-deprecated);
-  }
+  fill: var(--color-figure-gray);
 }
 
 .changed {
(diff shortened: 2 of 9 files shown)


hqhhuang:
@swift-ci test

src/components/Badge.vue
```diff
@@ -53,20 +54,18 @@ export default {
 
   @include font-styles(badge);
   display: inline-block;
-  padding: 2px 10px;
+  padding: 2px 4px;
   white-space: nowrap;
-  background: none;
-  border-radius: var(--badge-border-radius, calc(#{$border-radius} - 1px));
-  border-style: var(--badge-border-style, solid);
```
mportiz08, line 60: Should we keep these and just change the defaults so that users can continue to theme the style/width of the badge border?

src/components/DocumentationTopic/Summary/Availability.vue
```diff
@@ -99,16 +97,23 @@ export default {
...
   gap: 10px;
-  margin-top: 25px;
-}
-
-.badge {
-  margin: 0;
+  margin-top: rem(15px);
+  @include font-styles(body-reduced);
+
+  div:not(:last-child) {
```
mportiz08, line 109: These selectors are a little messy and hard to follow. I have one idea on a possible way to make it clearer that may or may not work, but it might be worth looking into.

1. Add the same class to every span that you want to have the separator (can be different components)
2. Change the rule to simply be `.that-class + .that-class` (every one preceded by another and apply the left border—no need for using not or last-child, etc)

I may be misunderstanding what this is doing, so feel free to ignore if that's the case.

src/components/DocumentationTopic/Summary/Availability.vue
```diff
@@ -119,72 +124,43 @@ export default {
...
-  }
-}
-
-.deprecated {
-  color: var(--color-badge-deprecated);
-
-  @include prefers-dark {
-    color: var(--color-badge-dark-deprecated);
-  }
+  fill: var(--color-figure-gray) !important;
```
mportiz08, line 127: Can we do this in a way that doesn't need `!important`?

Haven't tried these, but these are the workarounds I might try first:
* make sure your override has the right precedence
* possibly override `--color-svg-icon` here instead

src/components/Badge.vue
```diff
@@ -53,20 +54,18 @@ export default {
...
-  color: var(--badge-color);
-
-  .theme-dark & {
-    --badge-color: var(--badge-dark-color);
+  border-radius: var(--badge-border-radius, $border-radius);
+  border: none;
+  margin: auto;
+  margin-left: 5px;
+  color: var(--colors-button-text, var(--color-button-text));
+  background-color: var(--badge-color);
```
mportiz08, line 64: Now that we're using the badge color as a background fill instead of a border, we'll likely need to re-think the color values themselves, since I don't think the contrast ratio will be high enough in terms of accessibility.

src/components/Badge.vue
hqhhuang: Good point!

src/components/DocumentationTopic/Summary/Availability.vue
hqhhuang: Using sibling selector and apply left border would cause the dividing line to appear on items on a new line when the container wraps. I couldn't find a clean way to not apply border style to first element in each row. Having a dividing line on the right of the last item in a row looks not as broken.
But I added a new class name & cleaned this up a little. Let me know what you think.

hqhhuang:
@swift-ci test

src/components/DocumentationTopic/Summary/Availability.vue
mportiz08: Understand why you need to do the right border for wrapping 👍

How about something like this maybe?

```diff
diff --git a/src/components/DocumentationTopic/Summary/Availability.vue b/src/components/DocumentationTopic/Summary/Availability.vue
index 29f8156b..e1079bb5 100644
--- a/src/components/DocumentationTopic/Summary/Availability.vue
+++ b/src/components/DocumentationTopic/Summary/Availability.vue
@@ -10,18 +10,18 @@
 
 <template>
   <Section class="availability" role="complementary" :aria-label="$t('sections.availability')">
-    <div
+    <span
       v-for="technology in technologies"
       class="technology"
       :key="technology"
     >
       <TechnologyIcon class="tech-icon" />
-      <span class="availability-info">{{ technology }}</span>
-    </div>
+      <span>{{ technology }}</span>
+    </span>
 
-    <div
+    <span
       v-for="platform in platforms"
-      class="platform availability-info"
+      class="platform"
       :class="changesClassesFor(platform.name)"
       :key="platform.name"
     >
@@ -32,7 +32,7 @@
       />
       <Badge v-if="platform.deprecatedAt" variant="deprecated" />
       <Badge v-else-if="platform.beta" variant="beta" />
-    </div>
+    </span>
   </Section>
 </template>
 
@@ -106,11 +106,14 @@ $icon-size-default: 20px;
   margin-top: rem(15px);
   @include font-styles(body-reduced);
 
-  &:not(:last-child) .availability-info::after {
-    // apply on the span of technology class to ensure same separator height
-    content: " ";
-    border-right: 1px solid;
-    padding-left: 10px;
+  & > * {
+    border-right: 1px solid currentColor;
+    padding-right: 10px;
+
+    &:last-child {
+      border-right: none;
+      padding-right: 0;
+    }
   }
 }
 ```

Comment (mportiz08) [reply src/components/DocumentationTopic/Summary/Availability.vue]: And bonus points if you use a css variable for the `10px` since it would be used for the `gap` and also this padding.

**Comment by hqhhuang** [comment]
@swift-ci test

File: src/components/DocumentationTopic/Summary/Availability.vue (hunk 2 above)
Comment (marinaaisa) [reply src/components/DocumentationTopic/Summary/Availability.vue]: I haven't tried Marcus's proposal but it could work. I'd add a small detail: we could revert this change too if we don't use a `content` property https://github.com/apple/swift-docc-render/pull/783/files#diff-61f11105739958fab994f705ecdd503c38e30566ec60f517046ff58c2298d9afL150

File: src/components/DocumentationTopic/Summary/Availability.vue [hunk 5]
```diff
@@ -10,18 +10,18 @@
 
 <template>
   <Section class="availability" role="complementary" :aria-label="$t('sections.availability')">
-    <Badge
+    <div
```
marinaaisa, line 13: Instead of rendering a `div` , it could be more semantic if we render `li` elements inside a `ul`.

marinaaisa (changes requested):
I tested the new style and it looks great 🚀 

Marcus' comments to simplify the CSS were very reasonable, let's see if we can achieve it.
I left a comment to improve the AX semantics.

src/components/DocumentationTopic/Summary/Availability.vue
mportiz08: Maybe. I would definitely use a span or make sure to use inline or inline-block list items instead of a div block since these are all inline text that are meant to live on the same line(s).

src/components/DocumentationTopic/Summary/Availability.vue
```diff
@@ -99,92 +97,77 @@ export default {
...
-  &-modified {
-    border-color: var(--color-changes-modified);
+.availability {
+  display: flex;
+  flex-flow: row wrap;
+  gap: $availability-info-spacing;
+  margin-top: rem(15px);
+
+  & > * {
+    @include font-styles(body-reduced);
```
mportiz08, line 158: You can probably lift this up into the parent since it's probably redundant to be setting the font styles for every item

src/components/DocumentationTopic/Summary/Availability.vue
```diff
@@ -99,92 +97,77 @@ export default {
...
+    &::after {
+      content: '';
+      display: inline-block;
+      width: 1px;
+      height: 14px;
```
mportiz08, line 163: Is this meant to match the font size or line height? You may want to consider using em units if that's the case.

src/components/DocumentationTopic/Summary/Availability.vue
```diff
@@ -99,92 +97,77 @@ export default {
...
+      background: black;
```
mportiz08, line 164: This should be currentColor or something to switch between black/white depending on the color scheme, right? Or is it purposefully always black?

src/components/DocumentationTopic/Summary/Availability.vue
hqhhuang: I changed it to use <span> for now.
I have a bug tracking doing a cleanup of the `Summary` component/folder, since we no longer really have a summary section. Maybe as a part of that I can look into treating availability info as a `ul`.

tests/unit/components/DocumentationTopic/Summary/Availability.spec.js
```diff
@@ -74,52 +75,45 @@ describe('Availability', () => {
     expect(section.attributes('role')).toBe('complementary');
   });
 
-  it('renders a `Badge` for technologies, a `Badge` and `AvailabilityRange` for each platform', () => {
+  it('renders a `div` for technologies, a `AvailabilityRange` and a `Badge` if applicable for each platform', () => {
```
mportiz08, line 78:
```suggestion
  it('renders a `span` for technologies, a `AvailabilityRange` and a `Badge` if applicable for each platform', () => {
```

(or maybe update the whole test name to reflect how it works today in case it's different)

src/components/DocumentationTopic/Summary/Availability.vue
```diff
@@ -99,92 +97,77 @@ export default {
...
+  gap: $availability-info-spacing;
+  margin-top: rem(15px);
+  @include font-styles(body-reduced);
+
+  & > * {
+    &::after {
+      content: '';
+      display: inline-block;
+      width: 1px;
+      height: rem(14px);
```
mportiz08, line 163: Would `1em` work here for the intended effect? Not super important, just curious.

mportiz08 (approved):
LGTM now

hqhhuang:
@swift-ci test

src/components/DocumentationTopic/Summary/Availability.vue
marinaaisa: Looking forward to that refactor of the Summary component! It's okay if we apply the semantic changes then.

marinaaisa (approved):
LGTM! Thanks 👍

PR #784, opened 2024-02-29
Add alt text for screen readers to describe video content

marinaaisa:
Bug/issue #120903895, if applicable: 

## Summary

Add aria tags to improve AX in videos.

A visually hidden text is being referenced as description for the video and its custom controls using `aria-describedby` tag.
More info: https://www.w3.org/WAI/PF/HTML/wiki/Media_Alt_Technologies 

- Add `aria-describedby` to the video element: it describes the content video using a visually hidden text that is being referenced in the video element. In case an screen reader cannot read the description, it will still read the text content.
- Add `aria-label` to the video element: it indicates how the user should interact when there are custom controls. This is not added if video uses the default controls because it's not needed.
- Add `aria-roledescription` to the video element: it describes that the element is a video element, in case a screen reader don't tell.
- Add `aria-controls` to the custom controls: it links the video to custom controls when there are. This is not added if video uses the default controls because it's not needed.

## Dependencies

NA

## Testing

[videoTutorial.doccarchive.zip](https://github.com/apple/swift-docc-render/files/14448921/videoTutorial.doccarchive.zip)

Steps:
1. Download the attached .doccarchive above
2. Run: `VUE_APP_DEV_SERVER_PROXY=[path-to-doccarchive] npm run serve` 
3. Go to http://localhost:8080/tutorials/slothcreator/page-01
4. Assert that the video element has the `aria-describedby` tag referring to a span element which contains the alt text.
5. Assert that the control-button element has the `aria-describedby` tag referring to a span element which contains the alt text.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Asset.vue
+++ b/src/components/Asset.vue
@@ -1,7 +1,7 @@
 <!--
   This source file is part of the Swift.org open source project
 
-  Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
+  Copyright (c) 2021-2024 Apple Inc. and the Swift project authors
   Licensed under Apache License v2.0 with Runtime Library Exception
 
   See https://swift.org/LICENSE.txt for license information
--- a/src/components/Asset.vue
+++ b/src/components/Asset.vue
@@ -110,6 +110,8 @@ export default {
         autoplays: this.prefersReducedMotion ? false : this.videoAutoplays,
         posterVariants: this.videoPoster ? this.videoPoster.variants : [],
         deviceFrame: this.deviceFrame,
+        alt: this.asset.alt,
+        id: this.asset.alt ? this.identifier : null,
       };
     },
     assetListeners() {
--- a/src/components/Asset.vue
+++ b/src/components/Asset.vue
@@ -105,13 +105,13 @@ export default {
     videoProps() {
       return {
         variants: this.asset.variants,
-        showsControls: this.showsVideoControls,
+        showsDefaultControls: this.showsVideoControls,
         muted: this.videoMuted,
         autoplays: this.prefersReducedMotion ? false : this.videoAutoplays,
         posterVariants: this.videoPoster ? this.videoPoster.variants : [],
         deviceFrame: this.deviceFrame,
         alt: this.asset.alt,
-        id: this.asset.alt ? this.identifier : null,
+        id: this.identifier,
       };
     },
     assetListeners() {
--- a/src/components/ReplayableVideoAsset.vue
+++ b/src/components/ReplayableVideoAsset.vue
@@ -14,7 +14,7 @@
       ref="asset"
       :variants="variants"
       :autoplays="autoplays"
-      :showsControls="showsControls"
+      :showsDefaultControls="showsDefaultControls"
       :muted="muted"
       :posterVariants="posterVariants"
       :deviceFrame="deviceFrame"
@@ -25,10 +25,10 @@
       @ended="onVideoEnd"
     />
     <a
-      v-if="!showsControls"
+      v-if="!showsDefaultControls"
       class="control-button"
       href="#"
-      :aria-describedby="alt ? id : null"
+      :aria-controls="id"
       @click.prevent="togglePlayStatus"
     >
       {{ text }}
@@ -64,9 +64,9 @@ export default {
     },
     id: {
       type: String,
-      required: false,
+      required: true,
     },
-    showsControls: {
+    showsDefaultControls: {
       type: Boolean,
       default: () => false,
     },
@@ -121,7 +121,7 @@ export default {
     onPause() {
       const { video } = this.$refs.asset.$refs;
       // if the video pauses, and we are hiding the controls, show the replay button
-      if (!this.showsControls && this.isPlaying) {
+      if (!this.showsDefaultControls && this.isPlaying) {
         this.isPlaying = false;
       }
       this.videoEnded = video.ended;
--- a/src/components/ReplayableVideoAsset.vue
+++ b/src/components/ReplayableVideoAsset.vue
@@ -1,7 +1,7 @@
 <!--
   This source file is part of the Swift.org open source project
 
-  Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
+  Copyright (c) 2021-2024 Apple Inc. and the Swift project authors
   Licensed under Apache License v2.0 with Runtime Library Exception
 
   See https://swift.org/LICENSE.txt for license information
--- a/src/components/ReplayableVideoAsset.vue
+++ b/src/components/ReplayableVideoAsset.vue
@@ -28,6 +28,7 @@
       v-if="!showsControls"
       class="control-button"
       href="#"
+      :aria-describedby="id || null"
       @click.prevent="togglePlayStatus"
     >
       {{ text }}
--- a/src/components/ReplayableVideoAsset.vue
+++ b/src/components/ReplayableVideoAsset.vue
@@ -18,6 +18,8 @@
       :muted="muted"
       :posterVariants="posterVariants"
       :deviceFrame="deviceFrame"
+      :alt="alt"
+      :id="id"
       @pause="onPause"
       @playing="onVideoPlaying"
       @ended="onVideoEnd"
@@ -54,6 +56,14 @@ export default {
     variants: {
       type: Array,
       required: true,
+    },
+    alt: {
+      type: String,
+      required: false,
+    },
+    id: {
+      type: String,
+      required: false,
     },
     showsControls: {
       type: Boolean,
--- a/src/components/ReplayableVideoAsset.vue
+++ b/src/components/ReplayableVideoAsset.vue
@@ -28,7 +28,7 @@
       v-if="!showsControls"
       class="control-button"
       href="#"
-      :aria-describedby="id || null"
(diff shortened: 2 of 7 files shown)


marinaaisa:
@swift-ci test

src/components/Asset.vue
```diff
@@ -10,6 +10,13 @@
 
 <template>
   <div class="asset">
+    <span
```
mportiz08, line 13: Is there a reason why this needs to live in `Asset` instead of `VideoAsset`?

src/components/Asset.vue
```diff
@@ -110,6 +119,7 @@ export default {
         autoplays: this.prefersReducedMotion ? false : this.videoAutoplays,
         posterVariants: this.videoPoster ? this.videoPoster.variants : [],
         deviceFrame: this.deviceFrame,
+        altTextId: this.videoHasAlt ? this.altTextId : null,
```
mportiz08, line 122: Can we just pass the alt text itself here and render the video specific alt text element in the video component if it exists?

src/components/Asset.vue
```diff
@@ -63,6 +70,8 @@ export default {
     },
   },
   computed: {
+    altTextId: ({ _uid }) => `alt-video-text-${_uid}`,
```
mportiz08, line 73: Can we just use our existing `identifier` prop instead of relying on Vue's internal `_uid` value?

src/components/Asset.vue
marinaaisa: Fixed it

src/components/Asset.vue
marinaaisa: Good idea! I used the identifier as unique id :)

marinaaisa:
@swift-ci test

src/components/ReplayableVideoAsset.vue
```diff
@@ -26,6 +28,7 @@
       v-if="!showsControls"
       class="control-button"
       href="#"
+      :aria-describedby="id || null"
```
mportiz08, line 31: Did you mean to do something like `alt ? id : null` or `alt && id` here since we only add the description element with the id if there is alt text?

src/components/VideoAsset.vue
```diff
@@ -15,28 +15,38 @@
...
-    >
-      <!--
-        Many browsers do not support the `media` attribute for `<source>` tags
-        within a video specifically, so this implementation for dark theme assets
-        is handled with JavaScript media query listeners unlike the `<source>`
-        based implementation being used for image assets.
-      -->
-      <source :src="normalizePath(videoAttributes.url)">
-    </video>
+    <div>
```
mportiz08, line 18: Is the extra div needed because of the way that `ConditionalWrapper` works? Just curious. It would be nice to not add it, but I'm guessing that you already tried and found that it's necessary.

src/components/VideoAsset.vue
```diff
@@ -15,28 +15,38 @@
...
+      <span
+        v-if="alt"
+        :id="id || null"
```
mportiz08, line 21: Double check me, but I think just `:id="id"` would be the same behavior as `:id="id || null"` here.

But, I expect that the id is required if we're ever rendering this element, so maybe it should just be included in the above `v-if`, like `v-if="alt && id"`

src/components/VideoAsset.vue
```diff
@@ -15,28 +15,38 @@
...
+      </span>
+      <video
+        ref="video"
+        :controls="showsControls"
+        :data-orientation="orientation"
+        :autoplay="autoplays"
+        :poster="normalisedPosterPath"
+        :muted="muted"
+        :width="optimalWidth"
+        :aria-describedby="id || null"
```
mportiz08, line 34: Same comment from above: I think `:aria-describedby="id || null"` would be the same as `:aria-describedby="id"` since Vue will not actually populate the attribute if it is a falsy value.

src/components/ReplayableVideoAsset.vue
marinaaisa: We only pass the id if[ there is an alt property at the Asset component](https://github.com/apple/swift-docc-render/pull/784/files#diff-9267d8f28e02bad68c9715386d639fefd148b7ef49489bb469b299fe1deccad3R114), but you are right, adding `|| null` is useless here, Vue won't populate the attribute if it's an undefined value anyways.

marinaaisa: https://github.com/apple/swift-docc-render/pull/784/commits/8ffbb24898c76e4007589f624ea79a5639f0c3a9

src/components/VideoAsset.vue
marinaaisa: Yes, I tried to do:
```html
<ConditionalWrapper
    ref="wrapper"
    :tag="!!deviceFrame ? DeviceFrameComponent : 'div'"
    :should-wrap="!!deviceFrame || !!alt"
    :device="deviceFrame"
  >
```
which in theory it would allow me to render a 'div' in case that we pass an `alt` text: the `tag` prop is only used when `shouldWrap` prop is true.

But testing without an `alt` text it returns the Vue error:
`Multiple root nodes returned from render function. Render function should return a single root node.`

Somehow it counts the span text as multiple root node, even when it doesn't render. 😵‍💫

marinaaisa:
Hi @mportiz08 I did quite new improvements for this PR:

- Add `aria-describedby` to the video element: it describes the content video using a visually hidden text that is being referenced in the video element. In case an screen reader cannot read the description, it will still read the text content.
- Add `aria-label` to the video element: it indicates how the user should interact when there are custom controls. This is not added if video uses the default controls because it's not needed.
- Add `aria-roledescription` to the video element: it describes that the element is a video element, in case a screen reader don't tell.
- Add `aria-controls` to the custom controls: it links the video to custom controls when there are. This is not added if video uses the default controls because it's not needed.

src/components/VideoAsset.vue
```diff
@@ -15,28 +15,43 @@
...
-        within a video specifically, so this implementation for dark theme assets
-        is handled with JavaScript media query listeners unlike the `<source>`
-        based implementation being used for image assets.
-      -->
-      <source :src="normalizePath(videoAttributes.url)">
-    </video>
+    <div>
+      <span
+        v-if="alt"
+        :id="alt ? altTextId : null"
```
mportiz08, line 21:
```suggestion
        :id="altTextId"
```

Since you made id required and altTextId is always derived from id, it should never be null, and you will always need it for this element.

src/components/VideoAsset.vue
```diff
@@ -15,28 +15,43 @@
...
+        class="visuallyhidden"
+      >
+        {{ $t('formats.colon', {
+          content: $t('video.description')
+        }) + alt }}
```
mportiz08, line 26: You likely have more expertise in this area than me, but I'm wondering if we should be concatenating these two strings together like this as opposed to making the whole string localizable in the following way in case other languages require these phrases to be combined in different ways:

```vue
{{ $t('video.description', { alt }) }}
```

where the key in English might look like:

```json
"video": {
  "description": "Content description of this video: {alt}"
```

I'm definitely not an expert on this, so I'll defer to your judgement here.

src/components/VideoAsset.vue
```diff
@@ -15,28 +15,43 @@
...
+      </span>
+      <video
+        ref="video"
+        :id="id"
```
mportiz08, line 30: Just reminding you that when merging `main` back into this, you may get a small conflict due to my most recent PR—just remember to make sure that the `:key="videoAttributes.url"` line persists when resolving it—thanks!

src/components/VideoAsset.vue
marinaaisa: Fixed!

marinaaisa:
@swift-ci test

tests/unit/components/Asset.spec.js
```diff
@@ -102,9 +103,17 @@ describe('Asset', () => {
...
+    const videoAsset = mountAsset(identifier, { video }).find(ReplayableVideoAsset);
     expect(videoAsset.props('variants')).toBe(video.variants);
     expect(videoAsset.props('posterVariants')).toEqual([]);
+    expect(videoAsset.props('id')).toBe(identifier);
+    expect(videoAsset.props('alt')).toBe(video.alt);
+  });
+
+  it('renders a `ReplayableVideoAsset` without id if video does not provides an alt text', () => {
+    const videoAsset = mountAsset('video', { video: { ...video, alt: null } }).find(ReplayableVideoAsset);
+    expect(videoAsset.props('id')).not.toBe(video.id);
```
mportiz08, line 116: This test technically passes, but it's not doing what is described in the test (and I wouldn't expect it to). You made a change so that id is always required, and you are passing one, so we shouldn't be testing that there isn't one in this scenario—and it only doesn't have one because the `video` object here doesn't have one, so this test is saying the real "video" idea is not equal to `video.id` (`expect("video").not.toBe(undefined)`)

I don't think there's anything wrong with the code at this point to be clear, just this test.

marinaaisa: I modified this component so many times that I forgot to delete this old test 🤦‍♀️  Thanks!

marinaaisa:
@swift-ci test

Teapane (approved):

PR #785, opened 2024-03-06
Ensure video variants update in sync with a color scheme change

mportiz08:
Bug/issue #, if applicable: 124102221

## Summary

This fixes an issue where a video does not automatically switch to the expected variant (if multiple are available) when a user toggles from "Light" to "Dark" mode or vice-versa. After refreshing the page, the initial variant is always the correct one, but it does not dynamically switch as expected when using the color scheme toggle.

I'd love to hear if there is a better way to fix this or if I'm missing something obvious, because I don't think a `key` _should_ strictly be needed in this situation, but I may be overlooking something. I tried a couple of other things that seemed like more appropriate fixes, but neither worked:
* add a new `normalizedPath` variable to `data` and update it using a watcher on the `videoAttributes` computed property—it does get updated when toggle is changed, but the DOM doesn't update
* explicitly call `this.$forceUpdate()` as part of the above approach—still doesn't update

The key approach works since it forces Vue to not re-use DOM elements for videos with different URLs, but I feel like it shouldn't be necessary...so open to other ideas if I'm missing the true root issue :)

## Testing

Steps:
1. Add a video directive (example: `@Video(source: test-video)`) to a DocC article and make sure that both light/dark variants of the asset are provided in the catalog (example: `test-video.mp4` and `test-video~dark.mp4`).
2. Build this branch and use it when previewing the catalog with DocC (example: `DOCC_HTML_DIR=/path/to/swift-docc-render/dist docc preview Example.docc`)
3. Verify that the video switches from the light variant to the dark variant when switching from light to dark using the color scheme toggle (and vice-versa)
4. Verify that there are no other video-related regressions introduced by this change

## Checklist

- [ ] ~~Added tests~~ (not really a good/useful way to test the key attribute)

### DIFF
--- a/src/components/VideoAsset.vue
+++ b/src/components/VideoAsset.vue
@@ -17,6 +17,7 @@
   >
     <video
       ref="video"
+      :key="videoAttributes.url"
       :controls="showsControls"
       :data-orientation="orientation"
       :autoplay="autoplays"


marinaaisa (approved):
This seems the simplest and safest solution. If we find a better way to solve it in the future, we can iterate. Thanks!

mportiz08:
@swift-ci test

eango (approved):

PR #795, opened 2024-03-27
[6.0] Support thematic breaks

mportiz08:
- **Explanation:** Adds support for rendering thematic breaks using horizontal rules
- **Scope:** Only impacts content with thematic break syntax, compiled with supporting DocC version
- **Issue:** rdar://125507297
- **Risk:** Low, small additive change that has no impact on content without new syntax
- **Testing:** Added unit tests. Verified that new syntax gets rendered as horizontal rule when compiled with supported DocC version. Tested that no rendering changes occur without syntax and/or with unsupported DocC version.
- **Reviewer:** @marinaaisa and @emilyychenn 
- **Original PR:** #791

### DIFF
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -1,7 +1,7 @@
 <!--
   This source file is part of the Swift.org open source project
 
-  Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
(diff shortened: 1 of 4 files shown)


franklinsch (approved):

mportiz08:
@swift-ci test

PR #84, opened 2022-03-15
Allow disabling the filter clear on FilterInput tag selection

dobromir-hristov:
Bug/issue #, if applicable: 89796778

## Summary

Allows disabling the `clear on tag selection` feature in the FilterInput. This should allow the Navigator filter to apply tags, without clearing your query.

## Dependencies

NA

## Testing

Steps:
1. Go to a page with a navigator
2. Apply a title filter
3. Pick a Tag
4. Assert the filter is not removed

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #802, opened 2024-04-02
Clear filter input when a new tag is selected

hqhhuang:
Bug/issue #, if applicable: rdar://124954232

## Summary
We expect the filter input to reset when a new tag is selected. Removing the option to keep input when selecting a new tag since we never expect the filter to behave this way anymore.

## Dependencies
NA

## Testing
Manual testing & run unit test

Steps:
1. Type some input into the filter bar
2. Select a tag
3. Assert the input is cleared.

## Checklist

- [x] Added tests


mportiz08 (approved):
This is essentially just reverting #84, right? (Just checking)

PR #805, opened 2024-04-04
Links to other declarations should include query params

hqhhuang:
Bug/issue #, if applicable: rdar://124041300

## Summary
The links pushed to the router when users click on the pill for another declaration should include query parameters.

## Dependencies
NA

## Testing
Manual testing and run updated unit test 

## Checklist

- [x] Added tests


src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -162,7 +163,8 @@ export default {
       await this.$nextTick(); // wait for identifier to update
       this.isExpanded = false; // collapse the list
       await waitFor(500); // wait for animation to finish
-      this.$router.push(this.references[identifier].url);
```
mportiz08, line 165: You could also do the following instead of using `buildUrl`:

```js
this.$router.push({
  path: this.references[identifier].url,
  query: this.$route.query,
});
```

(Just noting the other option—you don't need to change it)

mportiz08 (approved):

PR #806, opened 2024-04-08
Don't emit selectedTags if there are no tags to paste

marinaaisa:
Bug/issue #125844017, if applicable: 

## Summary

Don't emit selectedTags if there are no tags to paste.
Emitting an empty selected tags array to the NavigatorCard was affecting the filter functionality.

## Dependencies

NA

## Testing

Steps:
1. Go to any documentation page with navigator
2. Copied a text
3. Paste it in the navigator's filter
4. Assert that filtering works well

## Checklist

- [x] Added tests

### DIFF
--- a/src/mixins/multipleSelection.js
+++ b/src/mixins/multipleSelection.js
@@ -537,11 +537,15 @@ export default {
       }
 
       this.setFilterInput(input.trim());
-      if (this.allSelectedTagsAreActive) {
-        this.setSelectedTags(tags);
-      } else {
-        this.updateSelectedTags(tags);
+
+      if (tags.length) {
+        if (this.allSelectedTagsAreActive) {
+          this.setSelectedTags(tags);
+        } else {
+          this.updateSelectedTags(tags);
+        }
       }
+
       this.resetActiveTags();
     },
     /**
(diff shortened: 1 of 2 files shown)


mportiz08 (approved):
I'm not 100% sure why this conditional wasn't necessary before and is now, but it seems to fix the issue with pasting and filtering and is a reasonable code change, so seems good to me.

marinaaisa:
> I'm not 100% sure why this conditional wasn't necessary before and is now, but it seems to fix the issue with pasting and filtering and is a reasonable code change, so seems good to me.

Yes, sorry, I didn't explain why it affects. It's a bit difficult and it took me time to debug.

On pasting, if we don't prevent from emitting "selectTags" when there are no tags available, we create a new empty array that triggers `nodeChangeDeps` computed property. It triggers its watcher `trackOpenNodes` one first time, it calls `generateNodesToRender` and then we persist the state calling `persistState`. The problem comes when the input changes and the watcher `trackOpenNodes` is trigger for a second time, it doesn't reach the `generateNodesToRender()` function because we have a return to skip in case we already have saved the filter in the storage.

marinaaisa:
@swift-ci test

PR #807, opened 2024-04-08
Improve grouping of elements for a better description of AX tags in videos

marinaaisa:
Bug/issue #126079234, if applicable: 

## Summary

Improve grouping of elements for a better description of AX tags in videos

## Dependencies

NA

## Testing

[videoTutorial.doccarchive.zip](https://github.com/apple/swift-docc-render/files/14448921/videoTutorial.doccarchive.zip)

Steps:
1. Download the attached .doccarchive above
2. Run: `VUE_APP_DEV_SERVER_PROXY=[path-to-doccarchive] npm run serve` 
3. Go to http://localhost:8080/tutorials/slothcreator/page-01
4. Open VoiceOver and assert that it tell you:
- The video element
- Its controls
- Description of the content

## Checklist

- [x] Added tests


src/components/ReplayableVideoAsset.vue
```diff
@@ -9,7 +9,18 @@
 -->
 
 <template>
-  <div class="video-replay-container">
+  <div
+    class="video-replay-container"
+    role="group"
+    :aria-roledescription="$t('video.title')"
```
mportiz08, line 15: I think this `aria-roledescription` attribute will end up getting duplicated in the DOM for one video because of the way that the replayable video component uses the video asset component.

We should probably make sure that there is only one of these present, regardless whether the video is replayable or not, right?

src/components/ReplayableVideoAsset.vue
```diff
@@ -9,7 +9,18 @@
...
+    :aria-labelledby="!showsDefaultControls ? ariaLabelledByContainer : null"
+  >
+    <span
+      id="custom-controls"
```
mportiz08, line 19: Is this `id` necessary? It will end getting duplicated when there are multiple replayable videos on a page, which we should try to avoid if possible.

src/components/ReplayableVideoAsset.vue
```diff
@@ -92,6 +103,7 @@ export default {
       if (this.videoEnded) return this.$t('video.replay');
       return this.isPlaying ? this.$t('video.pause') : this.$t('video.play');
     },
+    ariaLabelledByContainer: ({ id }) => `custom-controls ${id}-alt`,
```
mportiz08, line 106: Which element would this be referring to? I don't see it being set as the ID of any element, and I don't think this would even be a valid ID value since it has whitespace.

src/components/ReplayableVideoAsset.vue
marinaaisa: This should be fine because `aria-roledescription` is only read by VoiceOver if there is an `aria-label` or `aria-labelledby` in the same element.

In `.video-replay-container` and the `video` elements both have `aria-labelledby` conditionally rendered by `showsDefaultControls` so only one of them will be read by VO

src/components/ReplayableVideoAsset.vue
marinaaisa: The `aria-labelledby` property takes as value a space-separated id reference list, which means you can combine more than one element into a single accessible name. It will read the content of `custom-controls` id first and then the id alt text.

Source: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

src/components/ReplayableVideoAsset.vue
marinaaisa: True. I created unique ids for these elements: https://github.com/apple/swift-docc-render/pull/807/commits/813aa3fd78c51b8488b33f1b5fda90fe80f3d38d

marinaaisa: I also deleted delete dots from video identifiers. Because dots may cause issues in the DOM https://github.com/apple/swift-docc-render/pull/807/commits/b2b5d3ee8db097f88d5ba8d8d98ceb727a54da73

src/components/Asset.vue
```diff
@@ -111,7 +111,7 @@ export default {
         posterVariants: this.videoPoster ? this.videoPoster.variants : [],
         deviceFrame: this.deviceFrame,
         alt: this.asset.alt,
-        id: this.identifier,
+        id: this.identifier.replace('.', ''),
```
mportiz08, line 114: I don't think this is strictly necessary.

src/components/ReplayableVideoAsset.vue
mportiz08: Thanks, that makes sense.

We may need to make the 2nd part conditional since the element with the ID for the alt text will only exist if there was alt text provided.

src/components/ReplayableVideoAsset.vue
```diff
@@ -92,6 +103,7 @@ export default {
       if (this.videoEnded) return this.$t('video.replay');
       return this.isPlaying ? this.$t('video.pause') : this.$t('video.play');
     },
+    ariaLabelledByContainer: ({ id }) => `${id}-custom-controls ${id}-alt`,
```
mportiz08, line 106: The `${id}-alt` part should probably only be added if alt text was provided since the element it refers to won't be rendered for videos without alt text.

mportiz08 (approved):
Found some very minor things, but I think this is just about ready to go.

src/components/Asset.vue
marinaaisa: It's not, but it was giving me issues [using Jest](https://github.com/vuejs/vue-test-utils/issues/667), so I decided to add it just in case!

src/components/ReplayableVideoAsset.vue
marinaaisa: Sure, I fixed it https://github.com/apple/swift-docc-render/pull/807/commits/0bcacf6169211a954cef55b2997ca28fc48c0a88

src/components/Asset.vue
mportiz08: You likely ran into a CSS selector in a Jest test thinking that the dot is a class name. You can fix this by updating the selector to escape the dot using a backslash so that it treats it as a normal dot character and not the start of a classname.

https://stackoverflow.com/questions/12310090/css-selector-with-period-in-id

mportiz08: Escaping it in the test would be necessary and not in the template (just to clarify)

marinaaisa: Makes sense, thank you! https://github.com/apple/swift-docc-render/pull/807/commits/9cfb09966d7320a961bb4e3c78dd36aace0553b0

marinaaisa:
@swift-ci test

PR #811, opened 2024-04-09
Omit Mentioned In section on module pages

bitjammer:
rdar://125512559

Does not show a Mentioned In section on module pages, although it's possible in the future as the logic for collection mentions changes, along with its placement, so making this change in the renderer for now.

Testing: Manual testing with swift-markdown and unit test added.

### DIFF
--- a/tests/unit/components/DocumentationTopic.spec.js
+++ b/tests/unit/components/DocumentationTopic.spec.js
@@ -13,6 +13,7 @@ import DocumentationTopic from 'docc-render/components/DocumentationTopic.vue';
 import Language from 'docc-render/constants/Language';
 import InlinePlusCircleIcon from 'docc-render/components/Icons/InlinePlusCircleIcon.vue';
 import { TopicTypes } from '@/constants/TopicTypes';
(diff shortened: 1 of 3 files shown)


bitjammer:
@swift-ci Please test

tests/unit/components/DocumentationTopic.spec.js
```diff
@@ -368,6 +369,37 @@ describe('DocumentationTopic', () => {
...
+      references: hierarchyItemsReferences,
+      role: TopicTypes.symbol,
+      symbolKind: SymbolKind.protocol,
+      primaryContentSections: [
+        mentionSection,
+        foo,
+      ],
+    });
+
+    expect(wrapper.find(PrimaryContent).props()).toHaveProperty('sections', [
```
mportiz08, line 391: (Not a change you need to make, just pointing out that you can request specific props from the `wrapper.props` function, which can come in handy sometimes.)

```suggestion
    expect(wrapper.find(PrimaryContent).props('sections')).toEqual([
```

mportiz08 (approved):
Nice work!

PR #818, opened 2024-05-01
Use correct background color for "show other declarations button"

hqhhuang:
Bug/issue #, if applicable: rdar://126922233

## Summary
Background color of the show other declarations button should match the one used for the background of the content. 

## Dependencies
NA

## Testing
Manual testing

## Checklist

- [ ] Added tests


src/styles/core/colors/_light.scss
```diff
@@ -176,6 +176,7 @@
   --color-tabnav-item-border-color: var(--color-fill-gray);
   --color-text: var(--color-figure-gray);
   --color-text-background: var(--color-fill);
+  --color-other-decl-button: var(--color-text-background);
```
mportiz08, line 179: Super tiny thing, but most* of these are sorted alphabetically, and it would be nice to keep that pattern going so it's easier to scan through.

src/components/DocumentationTopic.vue
```diff
@@ -734,7 +734,8 @@ $space-size: 15px;
     top: 50%;
     left: 50%;
     transform: translate(-50%, -50%);
-    background-color: var(--color-fill); // cover border line
+    // cover border line
+    background-color: var(--colors-text-background, var(--color-other-decl-button));
```
mportiz08, line 738:
```suggestion
    background-color: var(--color-other-decl-button);
```

It's never not defined, so I don't think a fallback is needed.

mportiz08: I don't believe so. Theme-settings can override _any_ color var, and the [spec][1] uses the singular var name that is defined where the app defines all the shared colors.

There is an IDE-specific feature for synchronizing the IDE code syntax colors in `src/components/Tutorial/CodeTheme.vue`, but it is only used for code listings and not generic app colors—it shouldn't apply here.

Either way, the more specific color should be tried first with the more generic one being a fallback. Since this new color is already derived from an existing one, users could still override either just the text background to also apply it here, or they could override both to have a specific color for each, although that would be unlikely since this only has a specific color meant to match the background and override the border here, which is a very specific use case.

[1]: https://github.com/apple/swift-docc/blob/e6b81527f9f46347ededad3464b1f3585c81369d/Sources/SwiftDocC/SwiftDocC.docc/Resources/ThemeSettings.spec.json#L78

src/styles/core/colors/_light.scss
mportiz08: Makes sense, although I don't think that's the case since the value gets computed at runtime, so the source order shouldn't matter—test it out and double check me though!

marinaaisa (approved):

PR #826, opened 2024-05-20
Prevent issues when storing the top offset in AdjustableSidebarWidth

marinaaisa:
Bug/issue #128197719, if applicable: 

## Summary

Make sure the page is fully loaded before storing top offset in AdjustableSidebarWidth

## Dependencies

NA

## Testing

Steps:
1. Run a doccarchive with a navigator
2. Make sure that the navigator works as expected

## Checklist

- [ ] Added tests

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -15,7 +15,6 @@
   >
     <div :id="AppTopID" />
     <a href="#main" id="skip-nav" v-if="!isTargetIDE">{{ $t('accessibility.skip-navigation') }}</a>
-    <InitialLoadingPlaceholder />
     <slot name="header" :isTargetIDE="isTargetIDE">
       <SuggestLang v-if="enablei18n" />
       <!-- Render the custom header by default, if there is no content in the `header` slot -->
@@ -23,6 +22,7 @@
     </slot>
     <!-- The nav sticky anchor has to always be between the Header and the Content -->
     <div :id="baseNavStickyAnchorId" />
+    <InitialLoadingPlaceholder />
     <slot :isTargetIDE="isTargetIDE">
       <router-view class="router-content" />
       <custom-footer v-if="hasCustomFooter" :data-color-scheme="preferredColorScheme" />
--- a/src/components/AdjustableSidebarWidth.vue
+++ b/src/components/AdjustableSidebarWidth.vue
@@ -217,7 +217,6 @@ export default {
       if (this.focusTrapInstance) this.focusTrapInstance.destroy();
     });
 
-    await this.$nextTick();
     this.focusTrapInstance = new FocusTrap(this.$refs.aside);
   },
   watch: {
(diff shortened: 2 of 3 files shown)


mportiz08:
@marinaaisa are there testing instructions you can add to this?

marinaaisa:
> @marinaaisa are there testing instructions you can add to this?

Ops, sure, I just added them.

mportiz08:
> Thank you Marina for investigating this issue and explaining it. I think I may (or may not) have found a better way to fix this.
> 
> If I understood this correctly, the issue is that the offset we fetch of `baseNavStickyAnchor` is sometimes inaccurate, causing issues in our other logic. As I was reviewing your changes, I noticed that we use a `initialLoadingPlaceHolder` component [here](https://github.com/apple/swift-docc-render/blob/c9e9ef379c6e3d7fa7716c1c5fe9135f8615c3e5/src/App.vue#L18). I think the better solution is to move the placeholder component to be under the `baseNavStickyAnchor` [here](https://github.com/apple/swift-docc-render/blob/c9e9ef379c6e3d7fa7716c1c5fe9135f8615c3e5/src/App.vue#L25). Although delaying the timing of fetching this offset solves the issue, I think the actual issue is not necessarily the timing. What's probably happening is that the `baseNavStickyAnchor` is below the `initialLoadingPlaceHolder`, so the offset we are currently fetching is sometimes too big by the height of the placeholder component. If we move the placeholder to below the anchor, then the offset will always be correctly. In that case, we don't need to delay the logic for fetching the offset.
> 
> It's a bit hard to explain as well, but let me know if you think this makes sense?

Sounds right, but I think it's still a timing problem as you describe it. If I remember correctly, this placeholder is added to avoid an awkward situation before the routing and data fetching occurs so that the page doesn't look super compact and weird with the header/footer squashed together.

I would still consider it a timing issue though, because basically you need to wait for the time required for `$router.onReady` to fire (after which the placeholder is removed).

That also explains why this bug could happen, because I assume it's adding the height of the placeholder to the calculation if the calculation happens before it is removed (which is variable depending on the height of the window)—and by moving the placeholder below this anchor, its height wouldn't impact that calculation anymore.

marinaaisa:
> Sounds right, but I think it's still a timing problem as you describe it. If I remember correctly, this placeholder is added to avoid an awkward situation before the routing and data fetching occurs so that the page doesn't look super compact and weird with the header/footer squashed together.
> I would still consider it a timing issue though, because basically you need to wait for the time required for $router.onReady to fire (after which the placeholder is removed).
> That also explains why this bug could happen, because I assume it's adding the height of the placeholder to the calculation if the calculation happens before it is removed (which is variable depending on the height of the window)—and by moving the placeholder below this anchor, its height wouldn't impact that calculation anymore.

It's a really good point. I'm going to leave it as it is, then.

mportiz08:
I think the placeholder approach might be better—just wanted to highlight that it is still a problem caused by the timing of the placeholder getting removed. Sorry if I was unclear!

marinaaisa:
> Ops I was working on fixing tests. In case we still want to go with the placeholder approach I have that here: [hqhhuang@78edb51](https://github.com/hqhhuang/swift-docc-render/commit/78edb517560020932ee84c454ee3e93f1fe13223)

Okay, @hqhhuang , could you cherry pick [your commit](https://github.com/hqhhuang/swift-docc-render/commit/78edb517560020932ee84c454ee3e93f1fe13223) into this branch then?

mportiz08 (approved):

PR #828, opened 2024-05-22
Update new i18n strings

marinaaisa:
Bug/issue #125252837, if applicable: 

## Summary

Update new i18n strings

## Dependencies

NA

## Testing

Steps:
1. Make sure that everything continues working as expected

## Checklist

- [ ] Added tests


src/lang/locales/ko-KR.json
```diff
@@ -88,6 +95,7 @@
     "close": "닫기"
   },
   "sections": {
+    "attributes": "Attributes",
```
marinaaisa, line 98: Checking why this one didn't got translated

src/lang/locales/ja-JP.json
```diff
@@ -88,6 +95,7 @@
     "close": "閉じる"
   },
   "sections": {
+    "attributes": "Attributes",
```
mportiz08, line 98: Should we just remove this if we don't have a translation yet? The behavior is probably the same, just checking.

src/lang/locales/zh-CN.json
```diff
@@ -123,7 +131,7 @@
   "api-reference": "API 参考",
   "filter": {
     "title": "过滤",
-    "search": "搜索",
+    "search": "Search",
```
mportiz08, line 134: Was changing this one done on purpose?

src/lang/locales/ja-JP.json
marinaaisa: It seems that this one didn't got into the translation process (it may be newer). So I deleted this key.

src/lang/locales/zh-CN.json
marinaaisa: No, this is a mistake, reverting it.

marinaaisa:
@swift-ci test

mportiz08 (approved):

marinaaisa:
@swift-ci test

PR #829, opened 2024-05-23
Only display declaration modified diff when the other declaration list is collapsed

hqhhuang:
Bug/issue #, if applicable: rdar://127126248 ([Overloads][API Changes] Can't show other overload symbols when showing decl diff)

## Summary
Only display declaration modified diff when the other declaration list is collapsed

Also in this PR: Changed the name of a component from `DeclarationGroup` to `DeclarationList` in preparation of another bug fix. Splitting it off here with this bug since this fix is relatively small and less confusing to review.

## Dependencies
NA

## Testing
Steps:
1. Test with an overloaded page with modified changes
2. Assert that clicking on "Show other declaration" button shows other declarations

## Checklist

- [x] Added tests

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -15,7 +15,6 @@
   >
     <div :id="AppTopID" />
     <a href="#main" id="skip-nav" v-if="!isTargetIDE">{{ $t('accessibility.skip-navigation') }}</a>
-    <InitialLoadingPlaceholder />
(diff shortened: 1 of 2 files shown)


marinaaisa (approved):
Tested and it looks great. Thanks Hanqing!

PR #832, opened 2024-05-23
Other declaration pills should be clickable when there are changes data

hqhhuang:
Bug/issue #, if applicable: rdar://127125451 

## Summary
Other declaration pills should be clickable when there are changes data, and when the list of other declarations are expanded.

## Testing
Steps:
1. Test with pages with other declarations and with changes data to ensure that the pills are properly rendered and clickable 

Currently working on writing tests.

## Checklist

- [ ] Added tests

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -14,7 +14,9 @@
     :class="{ fromkeyboard: fromKeyboard, hascustomheader: hasCustomHeader }"
   >
     <div :id="AppTopID" />
-    <a href="#main" id="skip-nav" v-if="!isTargetIDE">{{ $t('accessibility.skip-navigation') }}</a>
+    <a href="#app-main" id="skip-nav" v-if="!isTargetIDE">
+      {{ $t('accessibility.skip-navigation') }}
+    </a>
     <slot name="header" :isTargetIDE="isTargetIDE">
       <SuggestLang v-if="enablei18n" />
       <!-- Render the custom header by default, if there is no content in the `header` slot -->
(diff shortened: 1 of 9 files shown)


src/components/DocumentationTopic/PrimaryContent/DeclarationDiff.vue
```diff
@@ -17,7 +17,6 @@
         :key="i"
         :declaration="declaration"
         :should-caption="currentDeclarations.length > 1"
-        :changeType="changeType"
```
mportiz08, line 20: Should we remove `changeType` as a prop for this component? It doesn't seem like it's being used anymore.

src/components/DocumentationTopic/PrimaryContent/DeclarationGroup.vue
```diff
@@ -0,0 +1,165 @@
...
+      multipleLinesClass,
+      displaysMultipleLinesAfterAPIChanges,
+    }) => ({
+      [`declaration-group--changed declaration-group--${changeType}`]: changeType,
+      [multipleLinesClass]: displaysMultipleLinesAfterAPIChanges,
+    }),
+    caption() {
+      return this.declaration.platforms.join(', ');
+    },
+    isSwift: ({ interfaceLanguage }) => interfaceLanguage === Language.swift.key.api,
```
mportiz08, line 86: Is this maybe old code? I see that it was moved, but I'm not seeing where it's being used.

mportiz08 (approved):

PR #247, opened 2022-05-03
Scroll to items as you navigate to them

dobromir-hristov:
Bug/issue #, if applicable: 90601877

## Summary

Items should be scrolled into little by little, when focused, instead of jumping to the middle of the navigator scrollarea.

This is an alternative approach to #225 

## Dependencies

NA

## Testing

Steps:
1. Using keyboard arrows or VO, navigate up/down across the navigator.
2. Assert items are scrolled as you reach the bottom

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran `npm test`, and it succeeded
- [x] Updated documentation if necessary

PR #835, opened 2024-05-24
Prevent programmatic focus change from bubbling in handleFocusIn

marinaaisa:
Bug/issue #128559835, if applicable: 

## Summary

We have a function called `getChildPositionInScroller` that was meant to scroll items in the navigator into little by little, when focused, instead of jumping to the middle of the navigator scrollarea. More info: https://github.com/apple/swift-docc-render/pull/247

This function was being called in programmatic focus changes, instead of the initial keyboard navigation purpose, so I'm preventing it from happening.

## Dependencies

NA

## Testing

Steps:
1. Run a doccarchive with a navigator
2. Assert that you can open and close nested symbols using the arrow keys without the navigator changing your focus
3. Assert that you can scroll long lists of symbols using keyboard navigation (arrow keys)

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -815,8 +815,10 @@ export default {
       if (!this.$refs.scroller) return false;
       return this.$refs.scroller.$el.contains(element);
     },
-    handleFocusIn({ target }) {
+    handleFocusIn({ target, relatedTarget }) {
       this.lastFocusTarget = target;
+      // prevent scroll when focus is programmatic
+      if (!relatedTarget) return;
       const positionIndex = this.getChildPositionInScroller(target);
       // if multiplier is 0, the item is inside the scrollarea, no need to scroll
       if (positionIndex === 0) return;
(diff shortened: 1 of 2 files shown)


mportiz08:
Things are looking pretty good so far from my testing :)

I have a question:

> This function was being called in programmatic focus changes, instead of the initial keyboard navigation purpose, so I'm preventing it from happening.

Can you explain what some example scenarios of the programmatic focus changes are? From my reading on the focus events, it seems that this could be from tabbing and things like that—does that sound right or am I misunderstanding the problematic scenario? I just want to make sure I understand.

marinaaisa:
> Things are looking pretty good so far from my testing :)
> 
> I have a question:
> 
> > This function was being called in programmatic focus changes, instead of the initial keyboard navigation purpose, so I'm preventing it from happening.
> 
> Can you explain what some example scenarios of the programmatic focus changes are? From my reading on the focus events, it seems that this could be from tabbing and things like that—does that sound right or am I misunderstanding the problematic scenario? I just want to make sure I understand.

Sure, this is a great question:
We have a function called `handleScrollerUpdate` that re-focuses the `lastFocusTarget` when we open and close groups of symbols. In this re-focus, the `focusin` event listener in the DynamicScroller is fired too and with it `handleFocusIn`.

mportiz08 (approved):

eango (approved):

mportiz08:
@swift-ci test

PR #837, opened 2024-05-26
Make Card's link position to be relative instead of absolute

marinaaisa:
Bug/issue #128642639, if applicable: 

## Summary

Make Card's link position to be relative instead of absolute. This is because when details in cards were long, the UI would get broken because the link was in a absolute position.

## Dependencies

NA

## Testing

Steps:
1. Run a doccarchive with normal and floating cards
2. Assert that cards look correct, regardless of how long the content is

## Checklist

- [x] Added tests


mportiz08 (approved):

mportiz08:
@swift-ci test

PR #846, opened 2024-05-31
Add navigator-title slot to NavigatorCard + fix small UI spacing details

marinaaisa:
Bug/issue #122909785, if applicable: 

## Summary

- Add navigator-title slot to NavigatorCard
- Add negative margin to bring the filter up while keeping the height and padding as it is

## Dependencies

NA

## Testing

Steps:
1. Assert that everything works well in desktop viewports
2. Assert that the filter is higher up in the small viewports

## Checklist

- [ ] Added tests

### DIFF
--- a/src/components/Navigator/NavigatorCard.vue
+++ b/src/components/Navigator/NavigatorCard.vue
@@ -1056,6 +1056,9 @@ $navigator-card-vertical-spacing: 8px !default;
   align-items: flex-end;
 
   .filter-on-top & {
+    // negative margin to bring the filter up
+    // while keeping the height and padding as it is
+    margin-top: calc(#{var(--nav-filter-vertical-padding)} * -1);
     border-top: none;
     align-items: flex-start;
   }
@@ -1065,9 +1068,10 @@ $navigator-card-vertical-spacing: 8px !default;
 
   @include breakpoint(medium, nav) {
     --nav-filter-horizontal-padding: 20px;
+    --nav-filter-vertical-padding: 10px;
     border: none;
-    padding-top: 10px;
-    padding-bottom: 10px;
+    padding-top: var(--nav-filter-vertical-padding);
+    padding-bottom: var(--nav-filter-vertical-padding);
     height: $filter-height-small;
   }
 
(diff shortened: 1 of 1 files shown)


src/components/Navigator/NavigatorCard.vue
```diff
@@ -32,18 +32,23 @@
...
-          :url="technologyPath"
-          :class="['technology-title', { 'router-link-exact-active': isTechnologyRoute }]"
-          @click.alt.native.prevent="toggleAllNodes"
-        >
-          <h2 class="card-link">
-            {{ technology }}
-          </h2>
-          <Badge v-if="isTechnologyBeta" variant="beta" />
-        </Reference>
+        <slot name="navigator-title" className="navigator-title">
```
mportiz08, line 35:
```suggestion
        <slot name="navigator-title">
```

I'm not sure you can add classes to the slot element directly, right? And it looks like you're already adding the class to the reference component below anyway. Also isn't `className` a react/jsx thing? 😁

src/components/Navigator/NavigatorCard.vue
```diff
@@ -1013,26 +1018,31 @@ $navigator-card-vertical-spacing: 8px !default;
   }
 }
 
-.technology-title {
+:deep(.navigator-title) {
```
mportiz08, line 1021: Does this really need to be a deep rule? Just checking.

mportiz08 (approved):

src/components/Navigator/NavigatorCard.vue
marinaaisa: Yes, to make sure the parent component also can use it.

marinaaisa:
@swift-ci test

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -184,6 +184,10 @@ export default {
       type: Object,
       default: () => ({}),
     },
+    customTag: {
+      type: String,
+      require: false,
```
mportiz08, line 189:
```suggestion
```

Or `required: false`

src/components/Navigator/NavigatorCardItem.vue
```diff
@@ -13,7 +13,7 @@
     :class="{ expanded, active: isActive, 'is-group': isGroupMarker }"
     :style="{ '--nesting-index': item.depth }"
     :data-nesting-index="item.depth"
-    :id="`container-${item.uid}`"
+    :id="item.uid ? `container-${item.uid}` : null"
```
mportiz08, line 16: I think we should keep the `uid` as a requirement since this id value is important for the underlying dynamic scrolling library that is used for nav items.

PR #841, opened 2024-05-29
Highlight unique components of overloaded declarations

mportiz08:
Bug/issue #, if applicable: 117503463

## Summary

Adds support for highlighting the unique components of overloaded declarations through a new, recursive declaration token component. These highlighted tokens will only be visually rendered when a group of overloaded symbols have been expanded by the user.

#### Example

Here, the highlights emphasize that this function is overloaded by the type of its only parameter.

Screenshot 2024-05-28 at 5 03 28 PM

## Dependencies

`swift-docc`: [Highlight declaration differences in overloaded symbol groups #928](https://github.com/apple/swift-docc/pull/928)

## Testing

Steps:
1. Checkout this branch and run `npm run build`
2. Checkout the branch of `swift-docc` from [this PR](https://github.com/apple/swift-docc/pull/928) and run `swift build`
3. Run `DOCC_HTML_DIR=/path/to/swift-docc-render/dist swift run docc preview --enable-experimental-overloaded-symbol-presentation Tests/SwiftDocCTests/Test\ Bundles/OverloadedSymbols.docc`
4. Verify that the unique parts of overloaded functions like the one in the above example are visually highlighted, only after disclosing the full list of the overloaded variants

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [X] Added tests
- [X] Ran `npm test`, and it succeeded
- [X] Updated documentation if necessary

\cc @QuietMisdreavus

PR #847, opened 2024-06-02
Highlight unique components of overloaded declarations [alternate approach]

mportiz08:
This implements the same rendering of highlighted tokens as #841 but using an alternate Render JSON schema where each token may possibly have a `highlightDiff` boolean flag instead of introducing a new token kind to represent highlights.

I'm not sure yet which approach the DocC team is going to decide on yet, so I'm opening a draft PR for both options and will close one depending on what they decide on.

## Testing

Same instructions as #841 except make sure to use the older commit `33388c3c7987904e0aec58e4035fea14bcd19965` from the swift-docc branch when building DocC instead of the latest commit from the mentioned branch.


src/components/DocumentationTopic/PrimaryContent/DeclarationList.vue
```diff
@@ -233,6 +233,11 @@ export default {
       margin: 0;
     }
   }
+
+  :deep(.highlighted) {
+    background: var(--color-syntax-highlighted, mark);
+    font-weight: $font-weight-semibold;
```
marinaaisa, line 239: As a design suggestion: I would add a transition to these properties.

marinaaisa (approved):
The implementation here looks cleaner than in https://github.com/apple/swift-docc-render/pull/841 in my opinion, but both are great.
I tested this one too and it works as expected.

I left a minor suggestion.
Thanks!

src/components/DocumentationTopic/PrimaryContent/DeclarationSource.vue
```diff
@@ -203,6 +204,10 @@ export default {
     handleWindowResize() {
       this.displaysMultipleLines = displaysMultipleLines(this.$refs.declarationGroup);
     },
+    extraClassesFor(token) {
+      const { highlightDiff: highlighted = false } = token;
```
QuietMisdreavus, line 208: I'm not familiar with this syntax. Does this mean that it will load the `highlightDiff` field from `token` into `highlighted`, setting it to `false` if it's undefined?

marinaaisa: Yes, this is a [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), taking the `highlightDiff` key and assigning it to a new const called `highlighted` with a default value of `false`, in case `highlightDiff` is undefined

QuietMisdreavus: Awesome, thanks for the info and the link!

src/components/DocumentationTopic/PrimaryContent/DeclarationList.vue
mportiz08: 55e6fd9baec9f09aabe67afc84d9825fbfcc3b6b

src/components/DocumentationTopic/PrimaryContent/DeclarationToken/RawText.vue
```diff
@@ -7,21 +7,13 @@
   See https://swift.org/LICENSE.txt for license information
   See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
+<template>
+  <span>{{text}}</span>
+</template>
 
 <script>
-// This component is an optimization that allows us to emit raw text DOM nodes
```
mportiz08, line 12: RIP 🥲

mportiz08:
@swift-ci test

QuietMisdreavus:
@mportiz08 I've merged https://github.com/apple/swift-docc/pull/928 with the `"highlight": "changed"` field, so this PR can land whenever you're ready. Looks like there's a merge conflict now.

mportiz08:
> @mportiz08 I've merged [apple/swift-docc#928](https://github.com/apple/swift-docc/pull/928) with the `"highlight": "changed"` field, so this PR can land whenever you're ready. Looks like there's a merge conflict now.

Sorry yes. We are anticipating needing to update this PR due to major refactoring of declaration components as part of #832 . I was hoping we could sync this change across both repos at the same time, but we can prioritize updating this PR now to try and avoid issues now that it has been merged in DocC.

mportiz08:
Replacing this PR with #875

PR #850, opened 2024-06-04
Fix Nav title's cursor and tag

marinaaisa:
Bug/issue #128955252, if applicable: 

## Summary

- Delete cursor default
- Change nav-title to be h2

## Dependencies

NA

## Testing

Steps:
1. Run a doccarchive
2. Go to any documentation page
3. Assert that Nav's title has a cursor when hovering
4. Assert that Nav's title is a h2

## Checklist

- [x] Added tests

### DIFF
--- a/src/views/DocumentationTopic.vue
+++ b/src/views/DocumentationTopic.vue
@@ -14,11 +14,11 @@
       v-if="topicData"
       v-bind="documentationLayoutProps"
     >
-      <template #nav-title>
+      <template #nav-title="{ className }">
         <component
           :is="rootLink ? 'router-link' : 'h2'"
           :to="rootLink"
-          class="nav-title"
+          :class="className"
         >
           {{ $t('documentation.title') }}
         </component>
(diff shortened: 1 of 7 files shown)


marinaaisa:
@swift-ci test

src/views/DocumentationTopic.vue
```diff
@@ -332,3 +332,12 @@ export default {
   },
 };
 </script>
+<style lang="scss" scoped>
+@import 'docc-render/styles/_core.scss';
+
+.nav-title {
+  font-size: inherit;
```
mportiz08, line 339: It seems like the `font-size` and `font-weight` you're adding here just get overwritten—do we really need these rules?

mportiz08 (approved):

src/views/DocumentationTopic.vue
marinaaisa: We need them for the h2 in the navigator, I moved them there.
This is because if we don't set the `h2` tag to be `inherit` it gets the `h2` styles by default.

The overwriting of the styles only happens in the `NavBase` which it's a different component

marinaaisa:
@swift-ci test

PR #868, opened 2024-06-24
Include Relationship section in minimized mode

hqhhuang:
Bug/issue #, if applicable: rdar://81022077

## Summary
Include Relationship section in minimized mode

## Dependencies
NA

## Testing
Steps:
1. Manually test in Quick Nav to ensure there are no regressions and the relationship info is displayed

## Checklist

- [x] Added tests


src/components/DocumentationTopic.vue
```diff
@@ -129,7 +129,6 @@
               :source="remoteSource"
               :sections="primaryContentSectionsSanitized"
             />
-            <ViewMore v-if="shouldShowViewMoreLink" :url="viewMoreLink" />
```
mportiz08, line 132: I may be misunderstanding things, but I'm wondering if you could simplify a lot of these changes by just conditionally rendering the relationships section above this and continuing to keep things in the "minimized-container".

As long as you make sure that you have one `<Relationships v-if="enableMinimized"...` here and one `<Relationships v-if="!enableMinimized..."` below, would that work and be a cleaner way of including this section in that mode?

I have a feeling you may have already looked into that approach, but I'm just curious. As much as possible, it would be nice to have the parent component dictate the layout for its children as opposed to having to pass the `enableMinimized` flags to lots of child components, where it doesn't really make sense in isolation for them.

src/components/DocumentationTopic/Relationships.vue
```diff
@@ -44,6 +45,10 @@ export default {
       type: Array,
       required: true,
     },
+    enableMinimized: {
```
mportiz08, line 48: I don't think you need to change this, but in the future it might be worth considering provide/inject for situations like this where you're adding props to a component solely for the purpose of passing it down to further nested children.

mportiz08 (approved):
Seems to be working well on a functional level.

As we discussed in person, the code for the "minimized" layout of documentation pages is starting to get pretty spread out and messy, but we may have to refactor things entirely using a new top-level component if we want to clean things up at this point. We may want to consider that in the future if we need to enhance this layout more.

PR #889, opened 2024-08-19
Creating an `IndexStore` for Quick Navigation

hqhhuang:
Bug/issue #, if applicable:  rdar://134272215

## Summary
1. This PR creates a new `IndexStore`. The eventual goal is to share this store between the navigator and Quick Navigation.
But for this PR, only Quick Navigation will be getting data from the new store.

2. Additionally, the index data is fetched from the page view instead.

3. Moves `IncludedArchiveIdentifiers` data from `AppStore` to the new `IndexStore`.

## Testing
Manual testing, verify that Quick Nav still behaves as expected.
I've added some tests. I'm working on adding more. 

## Checklist

- [ ] Added tests

### DIFF
--- a/tests/unit/mixins/indexProvider.spec.js
+++ b/tests/unit/mixins/indexProvider.spec.js
@@ -328,7 +328,7 @@ describe('indexProvider', () => {
     expect(IndexStore.state.flatChildren).toMatchSnapshot();
   });
 
-  it('flattens deeply nested children and provides them to the NavigatorCard', async () => {
+  it('flattens deeply nested children and sets it to `IndexStore`', async () => {
     fetchData.mockResolvedValue({
       interfaceLanguages: {
         [Language.swift.key.url]: [{
--- a/tests/unit/stores/IndexStore.spec.js
+++ b/tests/unit/stores/IndexStore.spec.js
@@ -78,17 +78,17 @@ describe('IndexStore', () => {
 
   it('sets `flatChildren`', () => {
     IndexStore.setFlatChildren(flatChildren);
-    expect(IndexStore.state.flatChildren).toBe(flatChildren);
+    expect(IndexStore.state.flatChildren).toEqual(flatChildren);
   });
 
   it('sets `references`', () => {
     IndexStore.setReferences(references);
-    expect(IndexStore.state.references).toBe(references);
+    expect(IndexStore.state.references).toEqual(references);
   });
 
   it('sets `apiChanges`', () => {
     IndexStore.setApiChanges(apiChanges);
-    expect(IndexStore.state.apiChanges).toBe(apiChanges);
+    expect(IndexStore.state.apiChanges).toEqual(apiChanges);
   });
 
   it('sets `includedArchiveIdentifiers`', () => {


marinaaisa:
I've been checking the code and testing Quick Navigation and it's looking good to me!

The only thing that I don't like in this current state is that we are duplicating the `NavigatorDataProvider` computed properties into the `indexProvider`. I understand that in the next PR we will get rid of the `NavigatorDataProvider` because `indexProvider` will replace it also for the Navigator, right?

src/components/DocumentationLayout.vue
```diff
@@ -150,6 +151,7 @@ export default {
       sidenavHiddenOnLarge: storage.get(NAVIGATOR_HIDDEN_ON_LARGE_KEY, false),
       showQuickNavigationModal: false,
       BreakpointName,
+      state: IndexStore.state,
```
mportiz08, line 154: Minor thing, but I might suggest naming this `indexState` or something since `state` is kind of generic on its own—especially since all the fields within `data` are technically state that may change over time.

src/components/DocumentationTopic.vue
```diff
@@ -615,7 +615,7 @@ export default {
         identifier: {
           interfaceLanguage,
           url: identifier,
-        },
+        } = {},
```
mportiz08, line 618: Was there a specific need for this change? Off the top of my head, I can't think of a scenario where we ever try to render a page without an identifier.

src/components/DocumentationTopic.vue
```diff
@@ -643,7 +643,7 @@ export default {
         seeAlsoSections,
         variantOverrides,
         variants = [],
-      } = json;
+      } = json || {};
```
mportiz08, line 646: Same question as above: is there ever a realistic scenario where `extractProps` is called with a null/undefined value?

src/components/Navigator/NavigatorDataProvider.vue
```diff
@@ -99,7 +99,7 @@ export default {
         );
         this.navigationIndex = Object.freeze(interfaceLanguages);
         this.navigationReferences = Object.freeze(references);
-        AppStore.setIncludedArchiveIdentifiers(includedArchiveIdentifiers);
+        IndexStore.setIncludedArchiveIdentifiers(includedArchiveIdentifiers);
```
mportiz08, line 102: Is this line still needed since this state is now being set in the view component? I know this component will likely be removed in the future, but I think the code using these identifiers are not tied to this provider like the navigator stuff is.

src/mixins/indexProvider.js
```diff
@@ -0,0 +1,98 @@
...
+        isTechnologyBeta: technologyWithChildren.beta,
+      }
+    ),
+  },
+  methods: {
+    async fetchIndexPathsData() {
+      const path = getAbsoluteUrl(this.indexDataPath);
+      return fetchData(path);
+    },
+    async fetchIndexData() {
```
mportiz08, line 74: I know the goal was to only use this new store for the quick navigation first, but this presents a temporary issue where we are always duplicating this request and the data itself, which can be very large and thus have a noticeable performance impact.

Would you have a concern with increasing the scope of this PR to use this new store in both places so we don't have that temporary performance regression?

src/views/DocumentationTopic.vue
```diff
@@ -86,6 +87,7 @@ export default {
       topicDataDefault: null,
       topicDataObjc: null,
       store: DocumentationTopicStore,
+      navigationIndex: {},
```
mportiz08, line 90: I may have just missed it, but is this used somewhere?

src/components/DocumentationLayout.vue
```diff
@@ -150,12 +152,21 @@ export default {
...
       showQuickNavigationModal: false,
       BreakpointName,
+      indexState: IndexStore.state,
     };
   },
   computed: {
     enableQuickNavigation: ({ isTargetIDE }) => (
       !isTargetIDE && getSetting(['features', 'docs', 'quickNavigation', 'enable'], true)
     ),
+    quickNavData({ indexState, interfaceLanguage }) {
```
mportiz08, line 162: Minor thing, but I think the "...Data" name is a little generic as to the shape of data it actually holds.

Maybe something like `indexNodes` or `navigatorNodes` would more clearly indicate that this prop will always resolve to the flat array of nodes?

src/components/DocumentationLayout.vue
```diff
@@ -150,12 +152,21 @@ export default {
...
+      let currentLangIndexData = indexState.flatChildren[interfaceLanguage] || [];
```
mportiz08, line 163: It seems like this is basically just looking up the nodes for the current language and falling back to the Swift ones as a default/backup, so maybe this could be cleaned up a bit? I'm thinking something like:

```javascript
indexNodes: ({ indexState: { flatChildren }, interfaceLanguage }) => (
  flatChildren[interfaceLanguage] ?? flatChildren[Language.swift.key.url]
),
```

What do you think?

mportiz08 (commented):
This seems cleaner to me overall as a first step of moving this data to a shareable store. 👍

Thanks for working through all these complex layers!

tests/unit/mixins/__snapshots__/indexProvider.spec.js.snap
mportiz08, line 1: Did you update these or is the main nav changes branch just not in sync with main? (Just checking)

tests/unit/mixins/indexProvider.spec.js
```diff
@@ -0,0 +1,521 @@
...
+describe('indexProvider', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  afterEach(() => {
+    IndexStore.reset();
+  });
+
+  it('fetches data when parent component imports `indexProvider`', async () => {
```
mportiz08, line 190: It fetches data when the path for the index data changes and not when the mixin is used, right?

tests/unit/mixins/indexProvider.spec.js
```diff
@@ -0,0 +1,521 @@
...
+          technologyClone,
+        ],
+      },
+    });
+    createWrapper();
+    await flushPromises();
+    expect(IndexStore.state.flatChildren).toMatchSnapshot();
+  });
+
+  it('flattens deeply nested children and provides them to the NavigatorCard', async () => {
```
mportiz08, line 331: Maybe "to a component"? I don't believe the navigator is using this mixin (just yet).

tests/unit/stores/IndexStore.spec.js
```diff
@@ -0,0 +1,108 @@
...
+      expect(IndexStore.state).not.toEqual(defaultState);
+      IndexStore.reset();
+      // assert all the state is reset
+      expect(IndexStore.state).toEqual(defaultState);
+    });
+  });
+
+  it('sets `flatChildren`', () => {
+    IndexStore.setFlatChildren(flatChildren);
+    expect(IndexStore.state.flatChildren).toBe(flatChildren);
```
mportiz08, line 81: I think you should use `toEqual` instead of `toBe` for any non-primitive JS types, like arrays of objects, so that it's checking for deep equality instead of reference equality.

mportiz08 (approved):

tests/unit/components/DocumentationLayout.spec.js
```diff
@@ -273,6 +287,64 @@ describe('DocumentationLayout', () => {
...
+      },
+    });
+    wrapper.setProps({
+      enableNavigator: true,
+    });
+    await wrapper.vm.$nextTick();
+    expect(wrapper.find(QuickNavigationModal).props('children')).toEqual(swiftChildren);
+  });
+
+  it('QuickNavigation falls back to swift items, if no objc items', async () => {
```
marinaaisa, line 306: Why do we fallback to swift items? just curious about this

marinaaisa (approved):
Reviewed and tested. LGTM! thanks for working on this!

PR #890, opened 2024-08-23
Update snapshot for `NavigatorDataProvider` test

hqhhuang:
Bug/issue #, if applicable: rdar://134589698

## Summary
I don’t think we saved the intended snapshot for the `NavigatorDataProvider` test.
The current snapshot has empty navigator data, but we are using it to test if the renderer contains the correct modified navigator data.

## Testing
Unit test

## Checklist

- [x] Added tests

### DIFF
--- a/bin/build-docs
+++ b/bin/build-docs
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 #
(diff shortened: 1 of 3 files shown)


marinaaisa (approved):
I run it and it looks great, thank you!

PR #891, opened 2024-09-10
Render availability items without version range details

mportiz08:
Bug/issue #, if applicable: 135420580

## Summary

This makes the `AvailabilityRange.introducedAt` prop optional instead of required.

This adds support for indicating that a symbol is available on a certain platform, even if there are no additional details on the specific version of that platform when it was first introduced.

### Example

For an example platform of _foo_, this is how it should now render:

**text**
_foo_ (instead of _foo 2.0+_)

**title** (hover text)
_Available on foo_ (instead of _Available on foo 2.0 and later_)

**aria-label**
_foo, Available on foo_ (same pattern of combining text/title)

## Testing

Steps:
1. Test that the availability strings work the same as today without regressions with symbols that have version details
2. Test for the above string patterns after removing the `introducedAt` value from existing availability items

## Checklist

- [X] Added tests

### DIFF
--- a/src/components/DocumentationTopic/Summary/AvailabilityRange.vue
+++ b/src/components/DocumentationTopic/Summary/AvailabilityRange.vue
@@ -26,7 +26,7 @@ export default {
     },
     introducedAt: {
       type: String,
-      required: true,
+      required: false,
     },
     platformName: {
       type: String,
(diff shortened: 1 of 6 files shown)


mportiz08:
@swift-ci test

marinaaisa (approved):
LGTM!

PR #895, opened 2024-09-18
Create a mixin for the index data getting logic

hqhhuang:
Bug/issue #, if applicable: rdar://136089361

## Summary
This is the 2nd PR of the refactor effort.

This PR does 2 small things:
1. Create a mixin for the index data getting logic, to allow us to avoid repeating the logic that selects the correct language variant of the index data.
2. set default state of `flatChildren` to `null`, so we don't need to create an additional variable to store `isFetching`. Instead, this can be computed by checking whether `flatChildren` is `null` and `errorFetching` is `false`. 

## Dependencies
NA

## Testing
Steps:
1. Check that there are no regressions

## Checklist

- [] Added tests


src/mixins/indexGetter.js
```diff
@@ -0,0 +1,52 @@
...
+    indexNodes({ indexState: { flatChildren }, interfaceLanguage }) {
+      if (!flatChildren) return [];
+      return flatChildren[interfaceLanguage] ?? (flatChildren[Language.swift.key.url] || []);
+    },
+    technologyProps({ indexState: { technologyProps }, interfaceLanguage, technology }) {
+      // Select technology props from fetched index data for the current language, fallback to swift
+      // If none available, fallback to technology data of the curr page or null
+      return technologyProps[interfaceLanguage] ?? technologyProps[Language.swift.key.url]
+        ?? (technology ? {
+          technology: technology?.title,
```
mportiz08, line 24: All of these `technology?.[prop]` calls shouldn't need the `?` since you're already in the branch of the conditional where there is a `technology` value, right?

src/mixins/indexGetter.js
```diff
@@ -0,0 +1,52 @@
...
+import IndexStore from 'docc-render/stores/IndexStore';
+import Language from 'docc-render/constants/Language';
+
+export default {
+  computed: {
+    indexNodes({ indexState: { flatChildren }, interfaceLanguage }) {
+      if (!flatChildren) return [];
+      return flatChildren[interfaceLanguage] ?? (flatChildren[Language.swift.key.url] || []);
+    },
+    technologyProps({ indexState: { technologyProps }, interfaceLanguage, technology }) {
```
mportiz08, line 19: Hopefully this logic can be removed at some point in the future. It's very confusing how it's written currently, although I realize you're just copying what is already there and not changing it yet.

mportiz08 (approved):

PR #909, opened 2024-10-30
Support min/max length data in "Attributes" section

mportiz08:
Bug/issue #, if applicable: 131432834

## Summary

This adds support for distinct `minimumLength` and `maximumLength` attributes within the top-level, primary "Attributes" section that may be rendered for REST documentation pages.

## Testing

Steps:
1. Verify that the "Minimum Length" and "Maximum Length" fields show up in the "Attributes" section for a page with relevant data

## Checklist

- [X] Added tests

### DIFF
--- a/src/lang/locales/en-US.json
+++ b/src/lang/locales/en-US.json
@@ -172,9 +172,9 @@
   "parameters": {
     "default": "Default",
     "minimum": "Minimum",
-    "minimumLength": "Minimum Length",
+    "minimumLength": "Minimum length",
     "maximum": "Maximum",
-    "maximumLength": "Maximum Length",
+    "maximumLength": "Maximum length",
     "possible-types": "Type | Possible types",
     "possible-values": "Value | Possible Values"
   },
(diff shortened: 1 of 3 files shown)


src/lang/locales/en-US.json
```diff
@@ -172,7 +172,9 @@
   "parameters": {
     "default": "Default",
     "minimum": "Minimum",
+    "minimumLength": "Minimum Length",
     "maximum": "Maximum",
+    "maximumLength": "Maximum Length",
     "possible-types": "Type | Possible types",
     "possible-values": "Value | Possible Values"
```
franklinsch, line 179: Not strictly related to this PR, but the casing is inconsistent across these parameter labels, e.g., Possible **t**ypes vs. Possible **V**alues. Which casing should we adopt?

mportiz08: I would be partial to `"Minimum length"` myself, but I was trying to match how I had seen it before in another place. These are default strings for the scenario where DocC doesn't provide the titles itself in the `title` field, which should be the more common scenario.

mportiz08:
@swift-ci test

franklinsch (approved):

PR #914, opened 2024-11-15
Anchor links shouldn't overwrite query parameters

marinaaisa:
Bug/issue #137046794, if applicable: 

## Summary

Anchor links shouldn't overwrite query parameters. For the Routerlink component to preserve the query, we need to pass the
routers query as a query parameter to the "to" prop.

## Dependencies

NA

## Testing

Steps:
1. Run this project using any .doccarchive
2. Add any query in the URL, example: `?context=foo`
3. Click in a title anchor
4. Assert that the query in the URL doesn't get overwritten

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode/LinkableHeading.vue
+++ b/src/components/ContentNode/LinkableHeading.vue
@@ -15,7 +15,7 @@
   >
     <router-link
       v-if="shouldLink"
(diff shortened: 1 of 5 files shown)


mportiz08 (approved):

marinaaisa:
@swift-ci test

PR #915, opened 2024-11-19
Add hover state to Hierarchy toggle background and hover + refactor

marinaaisa:
Bug/issue #135891984, if applicable: 

## Summary

Add hover state to Hierarchy toggle background and border.

Color specs for the toggle in hover state:
- Background: same background color as the Navigator's item hover
- Border: `figure-blue`

Also, I realised that we still had color styles for the nav bar in dark mode that was no longer used since we moved the breadcrumbs into the Hero at https://github.com/swiftlang/swift-docc-render/pull/772, so I deleted those styles in a second commit.

## Dependencies

NA

## Testing

Steps:
1. Go to any documentation page with many nested paths
2. Assert that you see the toggle to open the nested routes
3. Assert that the specs described above are correct

## Checklist

- [ ] Added tests


marinaaisa:
> It seems like the hover state color persists when the list is expanded but the mouse is off of the toggle. I would expect the hover background color to only be there when the mouse is actually hovering on the icon. What do you think?

Yes, this is because I added the styles to the active state too.

mportiz08:
It looks like this only adds the hover state to the actual toggle and none of the links within the expanded menu when you hover over them after using that toggle. Should we also add something for that?

biamx3:
> It looks like this only adds the hover state to the actual toggle and none of the links within the expanded menu when you hover over them after using that toggle. Should we also add something for that?

It looks like currently on hover, the color of the link changes. I think we don't need to change the fill color of the popover,  just adding the underline underneath the text like in the topics sections would be enough. 

Screenshot 2024-12-09 at 10 11 44

PR #928, opened 2025-02-04
Make nav-title 100% width in mobile viewport

marinaaisa:
Bug/issue #137436968, if applicable: 

## Summary

This is a small change to make the nav-title to be 100% width in mobile viewport instead of 90%.
There was no reason for it to be 90% instead.

## Dependencies

NA

## Testing

Steps:
1. Go to any documentation or tutorial page
2. Assert that the nav title looks correct in mobile viewport

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/NavBase.vue
+++ b/src/components/NavBase.vue
@@ -679,11 +679,11 @@ $content-max-width: map-deep-get($breakpoint-attributes, (nav, large, content-wi
   align-items: center;
   white-space: nowrap;
   box-sizing: border-box;
+  overflow: hidden;
 
   @include breakpoint(small, $scope: nav) {
     padding-top: 0;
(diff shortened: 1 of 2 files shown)


mportiz08 (commented):
This works well in the common case, but we have overflow issues when the title here is really long:

Screenshot 2025-02-04 at 3 43 51 PM

Is this something we could easily address in the same PR possibly?

marinaaisa:
Hi @mportiz08 I think I have solved this issue. Could you check that it works as expected?

mportiz08 (approved):
Awesome, seems to work great now. Thanks @marinaaisa!

franklinsch (approved):

PR #929, opened 2025-02-28
Rely on URL paths when choosing among multiple indices

mportiz08:
Bug/issue #, if applicable: 142337934

## Summary

This updates the logic used to choose between multiple top-level trees of data for the sidebar navigator:

Before, it would pick the first `module` in the list. (see #916)

Now, it will pick the first node with a path prefix matches the path prefix of the current URL.

This logic should only apply to rare scenarios where multiple trees of index data are included in a single Index JSON file.

## Testing

Steps:
1. Verify that there are no regressions with loading sidebar data in common, single-root scenarios.
2. Verify that the data loaded for the index generally matches what would be expected given the current URL in scenarios where there are multiple roots of index data.

## Checklist

- [X] Added tests

### DIFF
--- a/src/utils/navigatorData.js
+++ b/src/utils/navigatorData.js
@@ -182,7 +182,7 @@ function extractRootNode(data) {
   // note: this "root" path won't always necessarily come at the beginning of
   // the URL in situations where the renderer is being hosted at some path
   // prefix
-  const rootPathPattern = /^.*(\/documentation\/[^/]+).*$/;
+  const rootPathPattern = /(\/documentation\/[^/]+)/;
   const rootPath = window.location.href.match(rootPathPattern)?.[1] ?? '';
   // most of the time, it is expected that `data` always has a single item
   // that represents the top-level root node of the navigation tree
--- a/src/utils/navigatorData.js
+++ b/src/utils/navigatorData.js
@@ -179,16 +179,21 @@ export function getSiblings(uid, childrenMap, children) {
 }
 
 function extractRootNode(data) {
+  // note: this "root" path won't always necessarily come at the beginning of
+  // the URL in situations where the renderer is being hosted at some path
+  // prefix
+  const rootPathPattern = /^.*(\/documentation\/[^/]+).*$/;
+  const rootPath = window.location.href.match(rootPathPattern)?.[1] ?? '';
   // most of the time, it is expected that `data` always has a single item
   // that represents the top-level root node of the navigation tree
   //
   // there may be rare, unexpected scenarios where multiple top-level root
-  // nodes are provide for some reason—if that happens, we would prefer the
-  // first one categorized as a "module"
+  // nodes are provide for some reason—if that happens, we would prefer the one
+  // with a path that most closely resembles the current URL path
   //
   // otherwise, the first provided node will be used
   return data.length === 1 ? data[0] : (data.find(node => (
-    node.type === TopicTypes.module
+    node.path.toLowerCase() === rootPath.toLowerCase()
   )) ?? data[0]);
 }
 
--- a/tests/unit/utils/navigatorData.spec.js
+++ b/tests/unit/utils/navigatorData.spec.js
@@ -359,7 +359,11 @@ describe('when multiple top-level children are provided', () => {
   };
 
   describe('flattenNavigationIndex', () => {
-    it('prefers modules', () => {
+    it('prefers the root child with the same url path prefix', () => {
+      Object.defineProperty(window, 'location', {
+        value: { href: 'http://localhost/documentation/b/b42' },
+      });
+
       // use first root node if only one is provided
       let flattenedIndex = flattenNavigationIndex({ swift: [a] });
       expect(flattenedIndex.swift.length).toBe(1);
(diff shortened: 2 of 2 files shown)


mportiz08:
@swift-ci test

src/utils/navigatorData.js
```diff
@@ -179,16 +179,21 @@ export function getSiblings(uid, childrenMap, children) {
 }
 
 function extractRootNode(data) {
+  // note: this "root" path won't always necessarily come at the beginning of
+  // the URL in situations where the renderer is being hosted at some path
+  // prefix
+  const rootPathPattern = /^.*(\/documentation\/[^/]+).*$/;
```
tomstuart, line 185: Anchoring the start & end of the pattern with `^` and `$` is cancelled out by matching `.*` inside the anchor, so we can probably remove both.

tomstuart: Is there any way we can do this without assuming that the value of `node.path` a) definitely starts with `/documentation` and b) definitely contains only two path segments? It feels like it’s more likely to be robust against any future changes if we make as few assumptions as possible. (But maybe it’s not possible to avoid either of those!)

src/utils/navigatorData.js
```diff
@@ -179,16 +179,21 @@ export function getSiblings(uid, childrenMap, children) {
 }
 
 function extractRootNode(data) {
+  // note: this "root" path won't always necessarily come at the beginning of
+  // the URL in situations where the renderer is being hosted at some path
+  // prefix
```
tomstuart, line 184: Since we know the value of `baseUrl`, is it possible to match the `node.path` at the exactly correct position in the current page’s URL, i.e. directly following `baseUrl`, instead of allowing it to match anywhere?

src/utils/navigatorData.js
```diff
@@ -179,16 +179,21 @@ export function getSiblings(uid, childrenMap, children) {
...
   //
   // there may be rare, unexpected scenarios where multiple top-level root
-  // nodes are provide for some reason—if that happens, we would prefer the
-  // first one categorized as a "module"
+  // nodes are provide for some reason—if that happens, we would prefer the one
+  // with a path that most closely resembles the current URL path
   //
   // otherwise, the first provided node will be used
   return data.length === 1 ? data[0] : (data.find(node => (
-    node.type === TopicTypes.module
```
tomstuart, line 191: Is there any risk of a regression from removing the “always prefer the node whose `type` is `module`” behaviour, or do we now believe that it was a special case of the more generally correct behaviour (“always prefer the node whose `path` best matches the current page’s URL”) all along?

src/utils/navigatorData.js
```diff
@@ -179,16 +179,21 @@ export function getSiblings(uid, childrenMap, children) {
...
+    node.path.toLowerCase() === rootPath.toLowerCase()
```
tomstuart, line 196: Are we confident we won’t encounter a case where one node’s `path` is `/documentation/foo` and another’s is `/documentation/foo/bar`, and we have to pick the most specific (i.e. longest) one?

src/utils/navigatorData.js
mportiz08: > Is there any way we can do this without assuming that the value of node.path a) definitely starts with /documentation

Due to my note above in the comment about this path not always necessarily starting at the start of the URL path, I'm not sure there's an alternative I'm aware of.

> b) definitely contains only two path segments?

There isn't any other indicator we could use that I'm aware of. This code is already only executing for the top level items in the outermost (most-top level) array and only when there are multiple of these root nodes.

I'm up for any suggestions if you have any though. I agree that the less assumptions the better, but I'm not sure how this could be made any more generic at the moment.

src/utils/navigatorData.js
mportiz08: Good idea. I'll explore that.

src/utils/navigatorData.js
mportiz08: The module check was done only recently (#916), and it has turned out to not be sufficient, resulting in the need for this PR. In the past, a similar check was done before some major refactoring was done to the sidebar components, so I'm not worried about regressions from removing the module check specifically.

src/utils/navigatorData.js
mportiz08: I might be misunderstanding your concern here, but I can't think of a scenario where a nested item like `/documentation/foo/bar` would be provided as a root level node since the nesting is already visible in the path itself, so that particular node should never be a part of this logic which only runs for the top-level items (and only when there are multiple).

src/utils/navigatorData.js
mportiz08: > Anchoring the start & end of the pattern with `^` and `$` is cancelled out by matching `.*` inside the anchor, so we can probably remove both.

Fixed: 8de15f66edc94afd5daba5cf6b28e57ba46375f6

mportiz08:
@swift-ci test

marinaaisa (approved):
I'm not very happy about relying on `window.location` but I haven't found a better alternative.

I tried a different approach where I imported the Vue's router into the `navigatorData.js` file and used [currentRoute](https://v3.router.vuejs.org/api/#router-currentroute).fullPath to retrieve the path without the `baseUrl` prefix.

It wasn't cleaner than using `window.location` and it had some potentially risky changes in the `main.js` file so I don't think it's a good alternative.

Let's go with this approach then! Thank you @mportiz08

Issue #930, opened 2025-03-11
Update Tree-Sitter packages


roulpriya:
### Description

During the migration to the latest Node.js and VS Code versions, build failures that are specifically related to `swift-docc-render` occur. 

This currently blocks: https://github.com/swiftlang/vscode-swift/issues/1173#issuecomment-2710955549

To resolve this blocker, dependencies within the `swift-docc-render` package require updates, particularly `tree-sitter` related packages. One challenge is that `tree-sitter-vue` appears unmaintained, complicating the migration path. 

The compilation errors are due to old version of tree-sitter that use nan instead of node-addon-api

Updating to node-addon-api version of tree-sitter will make the package compatible with future versions of NodeJS as well

### Checklist

- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc-render/issues).

### Expected Behavior

- Successfully build `swift-docc-render` with latest Node.js and VS Code versions.

### Actual behavior

- Build fails when using the latest Node.js and VS Code versions.

### Steps To Reproduce

- Run `npm install` in the `swift-docc-render` project using Node 20

mportiz08:
Fixed with #931

PR #931, opened 2025-03-11
Update tree-sitter dependencies

roulpriya:
Fixes Bug #930

## Summary
Updating tree-sitter dependencies.

## Dependencies
This fix is the blocker to https://github.com/swiftlang/vscode-swift/issues/1173

## Testing
Existing test cases should pass, as the PR contains no code changes

Steps:
1. Run `npm install`

## Checklist

- [ ] Added tests (No code change involved, hence didn't add any tests)

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,8 +50,8 @@
     "jest": "^27.0.5",
     "sass": "^1.55.0",
     "sass-loader": "^12.0.0",
-    "tree-sitter": "^0.20.5",
-    "tree-sitter-javascript": "^0.20.1",
+    "tree-sitter": "^0.20.6",
+    "tree-sitter-javascript": "^0.20.4",
     "tree-sitter-jsdoc": "^0.19.0",
     "tree-sitter-vue": "^0.2.1",
     "vue": "^2.7.10",
@@ -63,7 +63,7 @@
     "vue-router": "^3.5.2"
   },
   "engines": {
-    "node": "18.16.1",
-    "npm": "9.5.1"
+    "node": ">=18.16.1 <21",
+    "npm": ">=9.5.1"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,8 +35,8 @@
         "jest": "^27.0.5",
         "sass": "^1.55.0",
         "sass-loader": "^12.0.0",
-        "tree-sitter": "^0.20.5",
-        "tree-sitter-javascript": "^0.20.1",
+        "tree-sitter": "^0.20.6",
+        "tree-sitter-javascript": "^0.20.4",
         "tree-sitter-jsdoc": "^0.19.0",
         "tree-sitter-vue": "^0.2.1",
         "vue": "^2.7.10",
@@ -44,8 +44,8 @@
         "vue-template-compiler": "^2.7.10"
       },
       "engines": {
-        "node": "18.16.1",
-        "npm": "9.5.1"
+        "node": ">=18.16.1 <21",
+        "npm": ">=9.5.1"
       },
       "peerDependencies": {
         "vue": "^2.7.10",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -50,9 +50,9 @@
     "jest": "^27.0.5",
     "sass": "^1.55.0",
     "sass-loader": "^12.0.0",
-    "tree-sitter": "^0.20.5",
-    "tree-sitter-javascript": "^0.20.1",
-    "tree-sitter-jsdoc": "^0.19.0",
+    "tree-sitter": "^0.21.1",
```
mportiz08, line 53: Unfortunately this is breaking the `npm run docs:build` and related scripts, which are the tasks that actually depend on the tree-sitter dev dependencies.

With that in mind, I see a few different viable options here. We can either:

1. See if vscode-swift can use the built version of the renderer: [swift-docc-render-artifact](https://github.com/swiftlang/swift-docc-render-artifact)

This contains the prebuilt assets for swift-docc-render that are generated when you run `npm run build`, and it is automatically updated anytime the main branch is updated here. I don't know enough about the setup of vscode-swift yet, but if it is able to use the prebuilt version, then it wouldn't even need node or any related npm dependencies.

2. See if there's a less new version of tree-sitter we can update to here without breaking things.

Maybe `v0.20.6` since `v0.21.0` mentions breaking API changes? Although I don't know if this will unblock the issue it is causing in vscode-swift, so maybe this isn't an option.

3. If none of the above options are viable, I think we could potentially disable the scripts for building docs temporarily to unblock vscode-swift if this is an urgent issue.

We could maybe add an early exit with a friendly message to `bin/generate-symbol-graph` to temporarily disable this functionality maybe until we are able to revive it since building docs for this project aren't super critical.

roulpriya: @mportiz08 Is it breaking due to tree-sitter-vue?

plemarquand: Unfortunately option 1 doesn't work for vscode because it applies a few patches to swift-docc-render before building it: https://github.com/swiftlang/vscode-swift/tree/main/scripts/patches/swift-docc-render. This means that it needs to build it from source. This is the root issue keeping the extension from migrating node versions.

mportiz08: > @mportiz08 Is it breaking due to tree-sitter-vue?

Seems to be, yes. It is crashing on the first line where that library is used—looking at the tree-sitter version change, it looks like there were breaking changes that the tree-sitter-vue library likely isn't compatible with yet.



> Unfortunately option 1 doesn't work for vscode because it applies a few patches to swift-docc-render before building it: https://github.com/swiftlang/vscode-swift/tree/main/scripts/patches/swift-docc-render. This means that it needs to build it from source. This is the root issue keeping the extension from migrating node versions.

That makes sense. Thanks for the context. I think we could integrate the functionality from those patches into the main branch at some point, but I'm guessing it's not trivial to do at the moment.

I think that leaves us with option 2 or 3 in order of preference. I'm ok with disabling the script for now since it seems like an urgent, blocking issue, but hopefully we could re-enable it in the near future.

roulpriya: @mportiz08 I have updated tree-sitter dependencies to the minimum version compatible with node 20. This is now building docs and tests are passing. It also works with `vscode-swift`.

tree-sitter-vue appears unmaintained, with no update for the last 4 years. It is stuck with nan bindings, which will not work for future node versions, while the rest of the node modules, such as `tree-sitter`, have migrated to `n-api` (There is no API change, rather an ABI change)

mportiz08: Thanks! Sounds like option 2 worked, but we'll need to re-consider how the `generate-symbol-graph` script works in the future when we update the minimum node version it sounds like due to the unmaintained tree-sittter-vue dependency :(

mportiz08 (approved):

mportiz08:
@swift-ci test

Teapane (approved):

Issue #919, opened 2024-12-19
Left Side Panel Very Slow


bhelyer:
### Description

(Apologies for using the Apple hosted documentation rather than reproducing locally. I'm not much of web developer.)

If I look at a documentation page for Swift's libraries, let's take https://developer.apple.com/documentation/swift/string/subsequence as our example, the time taken to scroll the left frame is longer than I would expect. On my slower machine (AMD A6, 8 gigs of RAM) scrolling a single new line of content takes 4-6 seconds. Scrolling one line up takes a further 4-6 seconds. Then scrolling one line down takes yet another delay.

On a faster machine (M2, 16 gigs of RAM) if I scroll multiple lines there's a delay of two seconds or so before content is drawn. If I then scroll up (to the place it was before) there's another delay, and then scroll down there's another delay.

### Checklist

- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc-render/issues).

### Expected Behavior

The frame should draw almost instantly if all the user is doing is scrolling, as it's static text. If there are technical reasons as to why it takes so much processing time, what I would say is that I would expect content that had previously been rendered to not take the same full time to display every time.

### Actual behavior

Content takes a long time to render (visible pause with no content while scrolling) and subsequent scrolls around the area doesn't change the render time of the content.

### Steps To Reproduce

1. Visit https://developer.apple.com/documentation/swift/string/subsequence (I was using Firefox 133) (using an older PC and not a top-end development machine ideally, but it's probably noticable on anything)
2. Scroll the left panel up and down.
3. Observe the redraw behaviour -- how long it takes, and if it gets any faster if rendering content it's already rendered.
4. Move your mouse over the labels, noting how long it takes for the blue highlight to catch up with your mouse cursor.
5. Enter text into the filter area. There should be a noticable lag. (On my slow machine, this takes four seconds for each keystroke!)

mportiz08:
Interesting. We do utilize the `vue-virtual-scroller` library that uses a windowing technique for displaying/scrolling the items in the sidebar because these navigation trees can be huge (like in the given example), and we don't want to create a huge amount of static DOM nodes that may introduce performance issues. It sounds like even with the library, we are still experiencing some performance issues here though.

[...]

PR #934, opened 2025-03-18
[Regression] [Performance] Fix delayed sidebar hover animation for large projects

mportiz08:
Bug/issue #, if applicable: #919 (rdar://147239963)

## Summary

Fixes a regression with the performance of the sidebar hover animation that can be very slow for large projects with many sidebar items.

This regression was introduced with #904.

The cause of the poor performance turned out to be a big jump in memory usage, which noticeably impacted the performance of hovering over sidebar items. By utilizing `Object.freeze` again, we prevent Vue from creating a large number of proxy items that it would normally use to track changes to state data and thereby keep the memory usage down for projects with many index nodes. Since we don't modify this data after fetching/deserializing it, this should be an acceptable tradeoff.

In the future if we need to handle even larger amounts of data or need to track changes to these nodes, we will likely need to come up with a way of streaming this data instead of storing it all in memory.

## Testing

Steps:
1. Use this branch to render a DocC catalog with a large number of navigation items.
2. Compare the performance of the hover effect for sidebar items and/or compare JS heap snapshots between this branch and the main branch to verify that the memory usage and performance better matches the behavior before #904 was merged.
3. Ensure there are no behavioral regressions with the function of the sidebar.

## Checklist

- [ ] ~~Added tests~~ (no straightforward way to unit test this change)

### DIFF
--- a/src/mixins/indexDataFetcher.js
+++ b/src/mixins/indexDataFetcher.js
@@ -28,7 +28,8 @@ export default {
           interfaceLanguages,
           references = {},
         } = await fetchData(this.indexDataPath);
-        IndexStore.setFlatChildren(flattenNavigationIndex(interfaceLanguages));
+        const flatChildren = Object.freeze(flattenNavigationIndex(interfaceLanguages));
+        IndexStore.setFlatChildren(flatChildren);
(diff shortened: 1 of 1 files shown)


marinaaisa (approved):
LGTM!

eango (approved):

PR #936, opened 2025-03-25
Compare page and node URLs less strictly when choosing between multiple indices

mportiz08:
Bug/issue #, if applicable: 147829785

## Summary

Fixes an edge case for #929 where the added logic doesn't apply cleanly when the node path in some Index JSON data includes some prefix before the `/documentation` segment.

By using the `endsWith` comparison instead of `===`, this logic can properly apply to that edge case without impacting the normal scenario with no base URLs.

## Testing

Steps:
1. Ensure #929 testing steps still work as expected
2. Test some content where the Index JSON has multiple top-level trees of data and the node paths include a path prefix before /documentation

## Checklist

- [X] Added tests

### DIFF
--- a/src/utils/navigatorData.js
+++ b/src/utils/navigatorData.js
@@ -193,7 +193,7 @@ function extractRootNode(data) {
   //
   // otherwise, the first provided node will be used
   return data.length === 1 ? data[0] : (data.find(node => (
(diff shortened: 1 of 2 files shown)


marinaaisa (approved):
LGTM!

eango (approved):

mportiz08:
@swift-ci test

PR #943, opened 2025-05-22
[TutorialsOverview] Ensure body always has dark background

mportiz08:
Bug/issue #, if applicable: 151701111

## Summary

Adds a small CSS rule to ensure that the `body` always has a dark background for the "tutorials overview" page, since it always has a dark theme, regardless of the user/system preferences.

## Testing

Steps:
1. Open a tutorials-overview page.
2. Verify that the `body` background is always black, no matter what the macOS system appearance setting and/or color scheme toggle is set to (may need to inspect it since it is usually covered up by other elements with a dark background)

## Checklist

- [ ] ~~Added tests~~ (CSS-only change)

### DIFF
--- a/src/components/TutorialsOverview.vue
+++ b/src/components/TutorialsOverview.vue
@@ -117,6 +117,15 @@ export default {
 };
 </script>
 
+<style lang="scss">
+@import 'docc-render/styles/_core.scss';
+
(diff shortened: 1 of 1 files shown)


marinaaisa (approved):
LGTM!

PR #957, opened 2025-07-28
update build-script-helper.py to use python 3 explicitly

heckj:
Updating the build script helper to explicitly use python3 to unblock using later versions of the official `node` Docker image, which doesn't include `python` as an executable - only `python3`.


heckj:
@swift-ci please test

PR #961, opened 2025-08-06
Add copy-to-clipboard support to code blocks

DebugSteven:
## Summary

This PR adds support for a new copy-to-clipboard button in code blocks. When this feature is enabled through the `enable-experimental-code-block` flag, a copy-to-clipboard button is rendered in the top-right corner of all code blocks, allowing users to easily copy its contents. When an author does not want the contents of a code block to be copyable, a `nocopy` option can be used to disable the copy-to-clipboard button.

### User Experience

When the `enable-experimental-code-block` flag is used, a copy button will appear in the top-right corner of all code blocks. Clicking the button copies the full contents of the code block to the clipboard and displays a checkmark to confirm success. If a code block includes the `nocopy` keyword in the language line, the copy button will not appear on that code block.

### Implementation Overview

- In `swift-docc-render`, this adds `copyToClipboard` to the properties on `BlockType.codeListing`.
- Adds a copy-to-clipboard button in `CodeListing.vue`, similar to swift.org’s copy-to-clipboard button on its code blocks, when `copyToClipboard` is set to true from the feature flag `enable-experimental-code-block`. Includes a function, `copyCodeToClipboard`, to copy the contents of a code block to clipboard. 
- Adding ` ```nocopy ` disables the copy button on that code block.
- This flag is forwarded from `swift-docc`. The accompanying branch/PR is here: https://github.com/swiftlang/swift-docc/pull/1273

## Dependencies

This PR depends on associated changes in `swift-docc` to pass the parameter for the copy button.

### Setup
1. Use the codeblock-copy branches for `swift-docc` and `swift-docc-render` with the copy-to-clipboard changes.
2. Rebuild documentation using `swift-docc` with the feature flag `enable-experimental-code-block` and serve it using a local build of `swift-docc-render`.

### How to Test
1. In all code listings with the `enable-experimental-code-block` flag, a copy button will appear in the top-right corner.
2. Click the copy icon. The code should be copied to your clipboard and the icon should update to a checkmark briefly. Paste to verify the copy functionality works.

To disable the copy icon with the feature flag enabled:
1. In any code listing using ` ``` ` or by adding a code listing, add the `nocopy` option like this:

``````
﻿```swift, nocopy
﻿print(“Hello, world!”)
﻿```
``````
﻿or like this:
``````
﻿```nocopy
﻿print(“Hello, world!”)
﻿```
``````
﻿
2. Verify the copy button does not appear on this code block.
Screenshot 2025-08-11 at 5 47 54 PM

## Checklist

- [x] Added tests
- [x] Updated documentation if necessary – On my DocC PR, I updated `RenderNode.spec.json` to include `copyToClipboard` as a property on `CodeListing`. Please let me know if there's any other documentation I should update here.

show on hover
always show copy button mobile

### DIFF
--- a/src/components/ContentNode/CodeListing.vue
+++ b/src/components/ContentNode/CodeListing.vue
@@ -1,7 +1,7 @@
 <!--
   This source file is part of the Swift.org open source project
 
-  Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
+  Copyright (c) 2021-2025 Apple Inc. and the Swift project authors
   Licensed under Apache License v2.0 with Runtime Library Exception
 
   See https://swift.org/LICENSE.txt for license information
@@ -22,6 +22,19 @@
     >{{ fileName }}
     </Filename>
     <div class="container-general">
+      <button
+        v-if="copyToClipboard"
+        class="copy-button"
+        :class="copyState"
+        @click="copyCodeToClipboard"
+        :aria-label="$t('icons.copy')"
+        :title="$t('icons.copy')"
+      >
+        <CopyIcon v-if="copyState === CopyState.idle" class="copy-icon"/>
+        <CheckmarkIcon v-else-if="copyState === CopyState.success" class="checkmark-icon"/>
+        <CrossIcon v-else-if="copyState === CopyState.failure" class="cross-icon"/>
+
+      </button>
       <!-- Do not add newlines in <pre>, as they'll appear in the rendered HTML. -->
       <pre><CodeBlock><template
         v-for="(line, index) in syntaxHighlightedLines"
@@ -45,16 +58,33 @@
 import { escapeHtml } from 'docc-render/utils/strings';
 import Language from 'docc-render/constants/Language';
 import CodeBlock from 'docc-render/components/CodeBlock.vue';
+import CopyIcon from 'theme/components/Icons/CopyIcon.vue';
+import CheckmarkIcon from 'theme/components/Icons/CheckmarkIcon.vue';
+import CrossIcon from 'theme/components/Icons/CrossIcon.vue';
 import { highlightContent, registerHighlightLanguage } from 'docc-render/utils/syntax-highlight';
 
 import CodeListingFilename from './CodeListingFilename.vue';
 
+const CopyState = {
+  idle: 'idle',
+  success: 'success',
+  failure: 'failure',
+};
+
 export default {
   name: 'CodeListing',
-  components: { Filename: CodeListingFilename, CodeBlock },
+  components: {
+    Filename: CodeListingFilename,
+    CodeBlock,
+    CopyIcon,
+    CheckmarkIcon,
+    CrossIcon,
+  },
   data() {
     return {
       syntaxHighlightedLines: [],
+      copyState: CopyState.idle,
+      CopyState,
     };
   },
   props: {
@@ -68,6 +98,10 @@ export default {
     content: {
       type: Array,
       required: true,
+    },
+    copyToClipboard: {
+      type: Boolean,
+      default: () => false,
     },
     startLineNumber: {
       type: Number,
@@ -91,6 +125,9 @@ export default {
       // `occ` is a legacy naming convention
       const fallbackMap = { occ: Language.objectiveC.key.url };
       return fallbackMap[this.syntax] || this.syntax;
+    },
+    copyableText() {
+      return this.content.join('\n');
     },
   },
   watch: {
@@ -121,6 +158,21 @@ export default {
       this.syntaxHighlightedLines = lines.map(line => (
         line === '' ? '\n' : line
       ));
+    },
+    copyCodeToClipboard() {
+      navigator.clipboard.writeText(this.copyableText)
+        .then(() => {
+          this.copyState = CopyState.success;
+        })
+        .catch((err) => {
+          console.error('Failed to copy text: ', err);
+          this.copyState = CopyState.failure;
+        })
+        .finally(() => {
+          setTimeout(() => {
+            this.copyState = CopyState.idle;
+          }, 1000);
+        });
     },
   },
 };
@@ -187,6 +239,7 @@ code {
   flex-direction: column;
   border-radius: var(--code-border-radius, $border-radius);
   overflow: hidden;
+  position: relative;
   // we need to establish a new stacking context to resolve a Safari bug where
   // the scrollbar is not clipped by this element depending on its border-radius
   @include new-stacking-context;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/components/ContentNode/CodeListing.vue [hunk 1]
```diff
@@ -122,6 +170,43 @@ export default {
         line === '' ? '\n' : line
       ));
     },
+    updateCopyButtonPosition() {
```
mportiz08, line 173: I'm thinking there might be a simpler way to position this element using pure CSS that would eliminate the need for refs and any custom JS listening to events, etc

As a very basic example, this is some minimal CSS that could be used to position a small red box with the letter "A" in the top-right corner of a given code listing. Maybe you could check this out and explore a CSS based solution to the layout and position for the icon/button to avoid the need for scroll event listeners?

```diff
diff --git a/src/components/ContentNode/CodeListing.vue b/src/components/ContentNode/CodeListing.vue
index 8f1524fc..ba7ceebc 100644
--- a/src/components/ContentNode/CodeListing.vue
+++ b/src/components/ContentNode/CodeListing.vue
@@ -129,6 +129,19 @@ export default {
 <style scoped lang="scss">
 @import 'docc-render/styles/_core.scss';
 
+.code-listing::after {
+  align-items: center;
+  background: red;
+  color: white;
+  content: 'A';
+  display: flex;
+  height: 25px;
+  justify-content: center;
+  position: absolute;
+  right: 0;
+  width: 25px;
+}
+
 .code-line-container {
   display: inline-block;
   width: 100%;
```

(The important pieces there being the `position: absolute` and `right: 0`)

mportiz08 (commented):
Hi @DebugSteven :) I'm really excited to see you bringing this functionality to DocC and DocC-Render, and I especially love the minimal/additive syntax proposed to make this a progressive, opt-in enhancement for DocC code listings.

Unfortunately, I'm about to be offline for a week, so I won't have time to properly test/review this PR during that time.

However, I took a brief look at the implementation and have a suggestion for a simpler way to position the icon/button that maybe you could explore—I promise to take a more complete look at this PR when I'm back online either way—sorry again for the bad timing on my end!

src/components/ContentNode/CodeListing.vue
DebugSteven: Thank you for this suggestion! I'll try to get the button positioning working with pure CSS and I'm glad it's feasible. I was concerned about the performance implications with the scroll event listeners.

src/components/ContentNode/CodeListing.vue
```diff
@@ -22,6 +22,33 @@
     >{{ fileName }}
     </Filename>
     <div class="container-general">
+      <button
```
mportiz08, line 25: Design-wise, I think the button can be a little distracting for code listings with a long first-line of code since the button always shows directly over the text (when enabled).

If we were to enable this button by default for code listings (which I personally think would be great), we may want to consider only showing it when hovering over the listing to try and eliminate some of that distraction maybe? That's just my personal opinion though. We could also consider a different layout where the button doesn't appear directly over the text as an alternative, although the positioning could get tricky in that case.

As a concrete example, 2 of the 3 code listings on this screenshot from the PR description are obscured by the button, regardless of where the user focus is:
476812913-bbc26127-bcb7-43b0-84f6-c52a95985040

src/components/ContentNode/CodeListing.vue
```diff
@@ -22,6 +22,33 @@
...
+        v-if="copyToClipboard"
+        class="copy-button"
+        :class="{ copied: isCopied }"
+        @click="copyCodeToClipboard"
+        aria-label="Copy code to clipboard"
+      >
+        <svg
```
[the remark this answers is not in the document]
DebugSteven: I separated the icons into separate components like you asked. However, I couldn’t get them to work with the `SVGIcon` component without messing up the CSS. Is there a trick I'm missing?

src/components/ContentNode/CodeListing.vue
```diff
@@ -22,6 +22,33 @@
     >{{ fileName }}
     </Filename>
     <div class="container-general">
+      <button
+        v-if="copyToClipboard"
+        class="copy-button"
+        :class="{ copied: isCopied }"
+        @click="copyCodeToClipboard"
+        aria-label="Copy code to clipboard"
```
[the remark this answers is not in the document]
DebugSteven: I added the string to en-US.json, but I’m not entirely sure it’s added correctly. Could you double check and adjust if needed or let me know the correct approach?

src/components/ContentNode/CodeListing.vue
DebugSteven: I agree that it’s nicer to show the copy button only when hovering over code blocks. Originally I had the button only show when hovering over code blocks. However, Joe and I got some feedback on the forums that it’d be better for mobile use to have the button always visible. 

I think I’ve got a decent solution here in my latest commit. I’m using `@media (hover: hover)` for devices that support hover and I added `@media (hover: none)`, when hover isn’t supported, to set the button to be always visible. Let me know what you think.

src/components/ContentNode/CodeListing.vue
```diff
@@ -205,4 +252,35 @@ pre {
   flex-grow: 1;
 }
 
+.copy-button {
+  position: absolute;
+  top: 0.2em;
+  right: 0.2em;
+  width: 24px;
```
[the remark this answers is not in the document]
DebugSteven: I changed the width/height to be 1.5em because I thought that would be equivalent to 24px. Let me know if you were thinking of doing this differently.

src/components/ContentNode/CodeListing.vue
DebugSteven: Just to share the same pictures from the forums, here's what those changes look like on desktop and mobile.
show on hover
always show button mobile

src/components/ContentNode/CodeListing.vue
```diff
@@ -122,6 +157,21 @@ export default {
...
+        .catch((err) => {
+          console.error('Failed to copy text: ', err);
+          this.copyState = CopyState.failure;
+        })
+        .finally(() => {
+          setTimeout(() => {
+            this.copyState = CopyState.idle;
+          }, 1000);
+        });
+    },
```
mportiz08, line 174:
```suggestion
    },
    CopyState,
```

Then, you can also use this constant in the template instead of the literal strings like this:

```
<CopyIcon v-if="copyState === CopyState.idle" ...
```
(etc)

mportiz08: That was my bad—thought I was pointing to a line at the same level of `methods` and not inside it. The way you did it works fine, thanks.

mportiz08 (approved):
📋✅

Nice work @DebugSteven!

heckj:
@swift-ci please test

PR #966, opened 2025-09-03
[Regression] border-width-right is hardcoded to 0 for asides

mportiz08:
Bug/issue #, if applicable: 159716327

## Summary

Fixes a regression introduced with #954 where the right border-width for asides is always hardcoded to 0 now.

While this doesn't have a major impact on the visuals since there is no right-border for asides normally, the variable that was previous used allowed for this border-width to be customized and should be re-introduced.

## Testing

Steps:
1. Load any page with an aside and verify that the borders still look the same
2. Modify the `$aside-border-width` variable and verify that the value is utilized on the right-hand-side of aside elements

## Checklist

- [X] Added tests


marinaaisa (approved):
thanks for looking into this, it looks good to me!

PR #979, opened 2025-11-24
Normalize Sample Code URLs

marinaaisa:
Bug/issue #165347857, if applicable: 

## Summary

We want to normalize Sample Code URLs, by adding the `baseUrl` to them if the path is relative and the baseUrl is provided. If the path is absolute or the baseUrl is not provided, it does not prefixes the URL.

Because the component `CallToActionButton.vue` is not only being used for Sample Code links, we created a new prop called `linksToAsset` that will normalize the relative URLs, in the cases that they are assets, like Sample Code.

## Dependencies

NA

## Testing

Steps:
1. Provide a `baseUrl`
2. Go to a page for Sample Codes that points to a relative path
3. Assert that the link gets prefixed with the baseUrl

## Checklist

- [x] Added tests

### DIFF
--- a/tests/unit/components/CallToActionButton.spec.js
+++ b/tests/unit/components/CallToActionButton.spec.js
@@ -26,7 +26,8 @@ describe('CallToActionButton', () => {
     linksToAsset: true,
   };
 
-  const relativePath = '/foo/bar';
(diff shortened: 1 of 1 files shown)


marinaaisa:
@swift-ci test

PR #980, opened 2025-11-25
Get the right data when the navigator has multiple top-level roots and multiple languages

marinaaisa:
Bug/issue #158778571, if applicable: 

## Summary

In the use cases of navigators with multiple top-level root and multiple languages, we want to make sure that the URL queries that are added in different languages don't interfere with the election of the top-level root node.

Before, we were reading the `window.location.href`, which includes queries. Now, by using the `window.location.pathname` property, we ignore the queries, which makes possible to compare with the `node.path` value to find which it's the correct top-level root.

Example: URL: `http://localhost:8080/documentation/foo?language=objc`
Navigator path: `/documentation/foo`

By using `window.location.pathname` we ignore `?language=objc` and we can compare `/documentation/foo` with `/documentation/foo`

## Dependencies

NA

## Testing

Steps:
1. Run DocC Render with a `.doccarchive` that has a navigator with multiple top-level roots and multiple languages (Swift, Obj-C)
2. Assert that the navigators are rendering correctly

## Checklist

- [x] Added tests

### DIFF
--- a/src/utils/navigatorData.js
+++ b/src/utils/navigatorData.js
@@ -183,7 +183,7 @@ function extractRootNode(data) {
   // the URL in situations where the renderer is being hosted at some path
   // prefix
   const rootPathPattern = /(\/documentation\/[^/]+)/;
-  const rootPath = window.location.href.match(rootPathPattern)?.[1] ?? '';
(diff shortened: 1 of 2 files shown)


mportiz08 (approved):

heckj:
@swift-ci please test

Issue #985, opened 2025-12-15
Scroll animation when clicking in header anchor


federicobucchi:
### Feature Name

Anchor Animation

### Description

- Users should be able to disable it (e.g.: prefers-reduced-motion)
- The focus should move with the scroll

### Motivation

It would be nice to have a scroll animation when clicking on an header.
Currently we have an animation in the open / close sidebar, so having a somewhat similar animation for the click on an header anchor would be nice.

### Importance

I don't think this is an important change, but it adds a smooth visual transition between sections.

### Alternatives Considered

The alternative is to leave it as it is, no scroll animation

GerardSerra18:
Hey @federicobucchi, since this is open, I am exploring the codebase to try implementing smooth scroll. Let me know if you have updates!

PR #986, opened 2025-12-15
Add smooth scroll

federicobucchi:
Bug/issue #985 

## Summary
Added CSS in `html` to enable the smooth scroll.  Disable the smooth scroll if the user has a setting that enables `prefers-reduced-motion`

## Dependencies
None

## Testing
All done

Steps:
- Navigate to a page with multiple headings / anchors
- Click on an anchor
- User should see a smooth scroll to the anchor

## Checklist

- [ ] Added tests

Tests are not needed for this and documentation shouldn't be updated for this.

### DIFF
--- a/src/styles/_base.scss
+++ b/src/styles/_base.scss
@@ -31,6 +31,11 @@
 
 html {
   --scale-inline: 1;
+  scroll-behavior: smooth;
+
+  @media (prefers-reduced-motion: reduce) {
+    scroll-behavior: auto;
+  }
 }
 
 html[dir="rtl"] {


mportiz08 (changes requested):
Sorry for not checking this PR out until now—I think I just missed it!

I don't see any issue with this PR in terms of the goal, but I think there are issues with the existing application that are preventing this from working as smoothly as it should 😅

When testing this locally with the PR changes, I'm sometimes seeing a regression where clicking a subsection further down on the page only partially scrolls to the right place whereas it always scrolls to the right place without these changes.

https://github.com/user-attachments/assets/b81f9224-d923-4e8f-835e-0b59d9c80e11

If I had to guess, this is likely some sort of race condition coming from logic we have related to vue-router and scroll events and not any issue with the CSS here. However, we would want to address that before taking these changes.

PR #989, opened 2025-12-17
Render ReferenceExternal when linking to an asset

marinaaisa:
Bug/issue #166301939, if applicable: 

## Summary

We want to make sure that when we render a link that links to an asset instead of a router path, we render the ReferenceExternal component, which it's meant to render links outside the router. 

## Dependencies

NA

## Testing

Steps:
1. Run a `doccarchive` with a link to an asset, for example, a sample code.
2. Go to the page where the download button of the sample code is.
3. Assert that it uses the ReferenceExternal component, instead of the Reference component. 

## Checklist

- [x] Added tests


marinaaisa:
@swift-ci test

PR #991, opened 2025-12-18
[6.3] Identify the root module of the navigator out of nested modules

marinaaisa:
**Explanation**: In some cases, our navigator can have nested modules within the top level modules. We want to extract all of them to be able to identify which it's the root module that the navigator should use.
This change introduces a recursively function that flatten a nested module structure into a single array.

**Scope**: Navigator
**Issue**: rdar://163125305.
**Risk**: Low
**Testing**: added new unit test.
**Reviewer**: @mportiz08 
**Original PR**: https://github.com/swiftlang/swift-docc-render/pull/981

### DIFF
--- a/src/utils/navigatorData.js
+++ b/src/utils/navigatorData.js
@@ -178,7 +178,19 @@ export function getSiblings(uid, childrenMap, children) {
   return getChildren(item.parent, childrenMap, children);
 }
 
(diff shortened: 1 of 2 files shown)


marinaaisa:
@swift-ci test

PR #997, opened 2026-02-22
URL 'q' parameter to automatically display the quick navigation UI

icanzilb:
Bug/issue #, if applicable: 170552244

## Summary

An enhancement that allows DocC Render to bring up the quick navigation UI automatically with a pre-filled search term, if a query is passed via the URL, i.e. `https://localhost..../argumentparser/?q=customize`:

image

Clients that send users to an online documentation are, this way, able to include a filter query in the URL that will display the filter UI as soon as the page opens. I.e. a blog post can refer a reader to all APIs in a framework that have to do with "compute" `https://......./myTechnology/?q=compute` or an error log message from a compiled framework might send developers debugging an issue directly to the matches that have to do with the specific error: `https://......./myTechnology/?q=Error 3964`

Finally, this allows fixed clients (in binary form) fast track users to documentation pages without the prior knowledge of their existence. E.g. like in the example above every error might be logged to a log with a link to the online documentation of the framework or app and pages with specific information for each specific error number might be added later on as these docs are created and published.

#### Implementation

When the layout is mounted there is an extra check in `handleInitialQueryFilter()` - if `q` is present in the URL query, it's consumed, the quick navigation UI shows up and the value of `q` is pre-filled. The window location is replaced with a new location without the `q` parameter to prevent other side-effects.

## Dependencies

No new or changed dependencies.

## Testing

The new behavior is covered with unit tests.

To test manually:

Steps:

1. Run dev server — npm run serve with a .doccarchive proxy set
2. Navigate to a documentation page with ?q=sometext — e.g.
  http://localhost:8080/documentation/myframework?q=Array
3. Verify: Quick Navigation modal opens automatically with "Array" pre-filled in
   the filter input
4. Verify: The ?q= param is stripped from the URL (browser address bar shows no
  ?q=)

## Checklist

- [X] Added tests

### DIFF
--- a/tests/unit/components/DocumentationLayout.spec.js
+++ b/tests/unit/components/DocumentationLayout.spec.js
@@ -595,7 +595,7 @@ describe('DocumentationLayout', () => {
         },
       });
       await w.setProps({
-        enableNavigator: true 
+        enableNavigator: true,
       });
       await w.vm.$nextTick();
       const modal = w.findComponent(QuickNavigationModal);
@@ -616,7 +616,7 @@ describe('DocumentationLayout', () => {
         },
       });
       await w.setProps({
-        enableNavigator: true 
+        enableNavigator: true,
       });
       await w.vm.$nextTick();
       const modal = w.findComponent(QuickNavigationModal);


d-ronnqvist:
I don't care particularly how this new query parameter is spelled, as long as it's a deliberate decision that has considered the alternatives and their respective trade offs.

src/constants/QuickNavigationQueryParams.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2026 Apple Inc. and the Swift project authors
```
mportiz08: Small thing, but can you make the following change in `bin/check-source` so that it doesn't trip over the year 2026 here? 😅 haha

```diff
diff --git a/bin/check-source b/bin/check-source
index a844b5d2..a5a14e52 100755
--- a/bin/check-source
+++ b/bin/check-source
@@ -18,7 +18,7 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 function replace_acceptable_years() {
     # this needs to replace all acceptable forms with 'YEARS'
-    sed -e 's/20[12][789012345]-20[12][89012345]/YEARS/' -e 's/20[12][89012345]/YEARS/'
+    sed -e 's/20[12][789012345]-20[12][890123456]/YEARS/' -e 's/20[12][890123456]/YEARS/'
 }
 
 printf "=> Checking license headers… "
```

tests/unit/components/DocumentationLayout.spec.js
```diff
@@ -539,4 +539,75 @@ describe('DocumentationLayout', () => {
...
+      getSetting.mockReturnValueOnce(true);
+      const w = createWrapper({
+        mocks: {
+          ...mocks,
+          $route: { path: '/documentation/somepath', query: { q: 'foo' } },
+          $router: { replace: jest.fn() },
+        },
+      });
+      await w.vm.$nextTick();
+      expect(w.vm.quickNavigationInitialFilter).toBe('foo');
```
mportiz08: Same note about using `findComponent` here and in the other new tests in this file

icanzilb: Yup, I changed all new tests. For the last one (navigator disabled) I didn't set stubs, etc., it wouldn't make sense to do the setup only to codify the default value right?

```
...
await w.setProps({
  enableNavigator: false
});
```

If you think there's merit in that, let me know. Thanks

mportiz08: Thanks!

There is already a checklist [item][1] for _Ran `npm test`, and it succeeded_ and this [script][2] does run both the linter and the script that checks for the license comment headers in source files.

[1]: https://github.com/swiftlang/swift-docc-render/blob/320eb485e4a8f8bf36e43c3b41a84da7c595599c/.github/pull_request_template.md?plain=1#L26
[2]: https://github.com/swiftlang/swift-docc-render/blob/320eb485e4a8f8bf36e43c3b41a84da7c595599c/package.json#L9

mportiz08:
@swift-ci test

PR #1003, opened 2026-04-15
Add locale fallback with redirect for 404 pages

marinaaisa:
Bug/issue #172836734, if applicable: 

## Summary

When a localized page (e.g. /zh-CN/documentation/...) returned a 404, the app would immediately show the not-found page without attempting to check if the content existed in the default locale.

Users navigating to a localized URL for untranslated content would see a 404 instead of being redirected to the available default locale version.

The fix adds a fallback step inside the 404 handler in fetchDataForRouteEnter. When the current locale differs from the default, it attempts to fetch the equivalent page with the locale prefix stripped from the path. If that fetch succeeds, it calls next() [1] with the fallback path to trigger a Vue Router navigation to the default locale page. If the fallback also fails, the error is logged to the console for debugging and the 404 page is shown as before.

The fallback path is derived by removing the locale segment from to.path (e.g. /zh-CN/documentation/foo becomes /documentation/foo), and the data path is verified via fetchData(createDataPath(fallbackPath)) before redirecting.

[1] https://v3.router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards

## Dependencies

NA

## Testing

Steps:
1. Download [SlothCreator.doccarchive.zip](https://github.com/user-attachments/files/26756075/SlothCreator.doccarchive.zip)
2. Run `VUE_APP_DEV_SERVER_PROXY=[path-to-doccarchive-file] npm run serve`
3. Go to http://localhost:8080/zh-CN/documentation/links and assert that it redirects to http://localhost:8080/documentation/links, since the JSON file only exists in English.
5. Go to http://localhost:8080/zh-CN/documentation/slothcreator and asserts that it shows "SlothCreator in Chinese" in the title, since the JSON file exists in English and Chinese.
6. Go to http://localhost:8080/zh-CN/documentation/nonexistingpage and asserts that it goes to the 404 page, since none of the JSON files exist.

## Checklist

- [x] Added tests

### DIFF
--- a/src/utils/data.js
+++ b/src/utils/data.js
@@ -13,6 +13,7 @@ import {
   queryStringForParams, areEquivalentLocations, getAbsoluteUrl,
 } from 'docc-render/utils/url-helper';
 import emitWarningForSchemaVersionMismatch from 'docc-render/utils/schema-version-check';
+import { defaultLocale } from 'theme/lang/index';
 import RedirectError from 'docc-render/errors/RedirectError';
 import FetchError from 'docc-render/errors/FetchError';
 
@@ -95,6 +96,19 @@ export async function fetchDataForRouteEnter(to, from, next) {
     }
 
     if (error.status && error.status === 404) {
+      // If the current locale is not the default, try fetching the default
+      // locale version as a fallback and redirect to it.
+      const locale = to.params && to.params.locale;
+      if (locale && locale !== defaultLocale) {
+        const fallbackPath = to.path.replace(`/${locale}`, '');
+        try {
+          await fetchData(createDataPath(fallbackPath), to.query);
+          next(fallbackPath);
+          return null;
+        } catch (fallbackError) {
+          console.error(`Fallback fetch failed for locale "${locale}":`, fallbackError);
+        }
+      }
       // route to 404 page if missing data, but not in IDE build
       next({
         name: 'not-found',
--- a/src/utils/data.js
+++ b/src/utils/data.js
@@ -13,6 +13,8 @@ import {
   queryStringForParams, areEquivalentLocations, getAbsoluteUrl,
 } from 'docc-render/utils/url-helper';
 import emitWarningForSchemaVersionMismatch from 'docc-render/utils/schema-version-check';
+import { defaultLocale } from 'theme/lang/index';
+import { localeIsValid } from 'docc-render/utils/i18n-utils';
 import RedirectError from 'docc-render/errors/RedirectError';
 import FetchError from 'docc-render/errors/FetchError';
 
@@ -95,6 +97,14 @@ export async function fetchDataForRouteEnter(to, from, next) {
     }
 
     if (error.status && error.status === 404) {
+      // Destructure the locale out of the route params, leaving the rest.
+      const { locale, ...paramsWithoutLocale } = to.params ?? {};
+      if (locale && locale !== defaultLocale && localeIsValid(locale)) {
+        // Call `next` with the same route but without the locale param,
+        // redirecting to the non-localized version of the route.
+        next({ ...to, params: paramsWithoutLocale });
+        return null;
+      }
       // route to 404 page if missing data, but not in IDE build
       next({
         name: 'not-found',
--- a/tests/unit/utils/data.spec.js
+++ b/tests/unit/utils/data.spec.js
@@ -34,7 +34,7 @@ const goodFetchResponse = {
   ok: true,
   json: () => Promise.resolve({ foobar: 'foobar' }),
 };
-const notFoundFetchResposne = {
+const notFoundFetchResponse = {
   ok: false,
   status: 404,
 };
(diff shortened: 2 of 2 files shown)


marinaaisa:
@swift-ci test

src/utils/data.js
```diff
@@ -95,6 +96,19 @@ export async function fetchDataForRouteEnter(to, from, next) {
     }
 
     if (error.status && error.status === 404) {
+      // If the current locale is not the default, try fetching the default
+      // locale version as a fallback and redirect to it.
+      const locale = to.params && to.params.locale;
```
mportiz08:
```suggestion
      const locale = to.params?.locale;
```

src/utils/data.js
```diff
@@ -95,6 +96,19 @@ export async function fetchDataForRouteEnter(to, from, next) {
...
+      if (locale && locale !== defaultLocale) {
+        const fallbackPath = to.path.replace(`/${locale}`, '');
```
mportiz08: To avoid the string replacement logic, I think you could just create a new route object with the locale param removed.

It might look something like this:

```javascript
const { locale, ...paramsWithoutLocale } = to.params;
if (locale && locale !== defaultLocale) {
  next({ ...to, params: otherParams });
  return null;
}
```

src/utils/data.js
```diff
@@ -95,6 +96,19 @@ export async function fetchDataForRouteEnter(to, from, next) {
...
+        try {
+          await fetchData(createDataPath(fallbackPath), to.query);
```
mportiz08: This line is probably unnecessary since it fetches data that it never uses—the `next` call in the subsequent line will trigger a new call to this function which already gets the data for a given URL (and the URL will have changed by removing the locale param).

src/utils/data.js
```diff
@@ -95,6 +96,19 @@ export async function fetchDataForRouteEnter(to, from, next) {
...
+          next(fallbackPath);
+          return null;
+        } catch (fallbackError) {
+          console.error(`Fallback fetch failed for locale "${locale}":`, fallbackError);
```
mportiz08: This is probably also unnecessary, since there is already an appropriate error handler for this scenario (the 404 page)

mportiz08 (commented):
Personally I think it makes more sense to have this redirect logic reside in the server that provides the data, but I don't think there's anything technically wrong with doing it client-side here.

src/utils/data.js
marinaaisa: great way to simply it, yes! I implemented :)

src/utils/data.js
marinaaisa: You are right, it's kind of redundant. The fetch result is thrown away, and navigating with next() will re-enter fetchDataForRouteEnter for the new URL anyway. If the fallback page doesn't exist, that second call will hit 404 and show the not-found page. I removed the whole try and catch block. 👍

src/utils/data.js
```diff
@@ -95,6 +96,11 @@ export async function fetchDataForRouteEnter(to, from, next) {
     }
 
     if (error.status && error.status === 404) {
+      const { locale, ...paramsWithoutLocale } = to.params ?? {};
+      if (locale && locale !== defaultLocale) {
```
mportiz08: I think we might need an additional check to make sure that the `locale` here is one of the supported ones—otherwise URLs like `/fakelocale/documentation/example` won't 404 like they should.

src/utils/data.js
```diff
@@ -95,6 +96,11 @@ export async function fetchDataForRouteEnter(to, from, next) {
     }
 
     if (error.status && error.status === 404) {
+      const { locale, ...paramsWithoutLocale } = to.params ?? {};
```
mportiz08: Can you add a brief comment explaining what this logic is doing here? The conventions for function names like `next` here are likely a little unclear to someone who isn't already familiar with the vue-router API

src/utils/data.js
marinaaisa: Yep, you are right. I used `localeIsValid` from `docc-render/utils/i18n-utils` to only redirect when the locale is valid.

src/utils/data.js
marinaaisa: done!

marinaaisa:
@swift-ci test

marinaaisa:
@swift-ci test

src/utils/data.js
mportiz08: Looks like this was changed back by mistake?

src/utils/data.js
mportiz08: I'm not seeing that. Did something go wrong in the rebase maybe?

mportiz08 (commented):
I think something didn't go as expected with a rebase/force-push. It looks like the code from the original PR is here now

marinaaisa:
> I think something didn't go as expected with a rebase/force-push. It looks like the code from the original PR is here now

Yes, you are right, something went wrong with the previous rebase, sorry. I've already fixed it and now it's ready to be reviewed. Thanks for the patience 🙏

src/utils/data.js
marinaaisa: it's back! ready to be reviewed

mportiz08 (approved):

marinaaisa:
@swift-ci test

PR #1004, opened 2026-04-16
New `@Card` directive

mportiz08:
Bug/issue #, if applicable: 173272119

## Summary

This PR adds support for a new `@Card` block directive, which groups and highlights blocks of content
to be rendered within a single "card" block.

See the associated Swift Forums pitch/discussion [here][1].

### Examples

```markdown
@Card {
	Hello

	World
}
```

example-01

```markdown
@Card {
	### This is a heading

	More heading content

	---

	This content is the main body of the card.

	Here is some more body content.
}
```

example-02

## Dependencies

https://github.com/swiftlang/swift-docc/pull/1508

## Testing

Steps:
1. Verify unit tests pass with `npm test`.
2. Build this branch with `npm run build` and build the swift-docc branch listed above with `swift build`.
3. Test example content with `DOCC_HTML_DIR=/path/to/swift-docc-render/dist swift run preview --enable-experimental-card-directive /path/to/Example.docc`

## Checklist

- [X] Added tests

[1]: https://forums.swift.org/t/card-directive/86076


src/components/ContentNode.vue
```diff
@@ -428,6 +430,14 @@ function renderNode(createElement, references) {
...
     case BlockType.thematicBreak:
       return createElement(ThematicBreak);
+    case BlockType.card:
+      return createElement(Card, {}, ([
+        ...renderChildren(node.head ?? []).map(vnode => ({
+          ...vnode,
+          data: { ...vnode.data, slot: 'head' },
+        })),
+        ...renderChildren(node.content),
+      ]));
```
marinaaisa, line 440: I'm guessing you have already tried this approach but would `scopedSlots` work, like in `tabNavigator` and `unorderedList` cases?

Something like:
```js
return createElement(Card, {
  scopedSlots: {
    head: () => renderChildren(node.head ?? []),
  },
}, renderChildren(node.content));
```

mportiz08: That doesn't work since these are just simple named slots and not scoped slots.

I do think that manually re-constructing the vnode using the data attribute like this is a _little_ hacky though. I basically need to do `createElement(..., { slot: 'head' }, ...)` for each of the children in the head section—but it's tricky to do without changing the way that the recursive `renderNode`/`renderChildren` functions work.

It's a little harder to read, but I think this is a way I could do it purely using the `createElement` API instead of manually re-assigning the data to the vnodes it returns:

```diff
diff --git a/src/components/ContentNode.vue b/src/components/ContentNode.vue
index c3af60a3..46250bbc 100644
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -121,6 +121,14 @@ const TabNavigatorVerticalThreshold = 7;
 //
 // Note: A plain string of text is returned for nodes with `type="text"`
 function renderNode(createElement, references) {
+  const createElementInSlot = slot => (element, data, children) => createElement(
+    element,
+    { ...data, slot },
+    children,
+  );
+  const renderChildrenInSlot = (children, slot) => children.map(
+    renderNode(createElementInSlot(slot), references),
+  );
   const renderChildren = children => children.map(
     renderNode(createElement, references),
   );
@@ -432,10 +440,7 @@ function renderNode(createElement, references) {
       return createElement(ThematicBreak);
     case BlockType.card:
       return createElement(Card, {}, ([
-        ...renderChildren(node.head ?? []).map(vnode => ({
-          ...vnode,
-          data: { ...vnode.data, slot: 'head' },
-        })),
+        ...renderChildrenInSlot(node.head ?? [], 'head'),
         ...renderChildren(node.content),
       ]));
     case InlineType.codeVoice:
```

That would be a less hacky version of assigning the slots normally (although it might be harder to read). Which version do you prefer?

marinaaisa: Oh true, this is a normal slot, not a scoped one. Good call!

Your proposal is very interesting but `createElementInSlot` seems to inject `slot` into every element in the subtree, not just the top-level children. Vue only checks slot on direct children of Card, so nested elements silently get unnecessary data.

Maybe we could do something in between like:

```diff
diff --git a/src/components/ContentNode.vue b/src/components/ContentNode.vue
index 72fa9517..7bf16598 100644
--- a/src/components/ContentNode.vue
+++ b/src/components/ContentNode.vue
@@ -121,18 +121,12 @@ const TabNavigatorVerticalThreshold = 7;
 //
 // Note: A plain string of text is returned for nodes with `type="text"`
 function renderNode(createElement, references) {
-  const createElementInSlot = slot => (element, data, children) => createElement(
-    element,
-    { ...data, slot },
-    children,
-  );
-  const renderChildrenInSlot = (children, slot) => children.map(
-    renderNode(createElementInSlot(slot), references),
-  );
-
   const renderChildren = children => children.map(
     renderNode(createElement, references),
   );
+  const renderChildrenInSlot = (slot, children) => (
+    renderChildren(children).map(vnode => ({ ...vnode, data: { ...vnode.data, slot } }))
+  );

   const renderListItems = items => items.map(item => (
     createElement('li', {}, (
@@ -441,7 +435,7 @@ function renderNode(createElement, references) {
       return createElement(ThematicBreak);
     case BlockType.card:
       return createElement(Card, {}, [
-        ...renderChildrenInSlot(node.head ?? [], 'head'),
+        ...renderChildrenInSlot('head', node.head ?? []),
         ...renderChildren(node.content),
       ]);
     case InlineType.codeVoice:
```

It's very similar to the original code you wrote but we can re-use `renderChildrenInSlot` in the future.

src/components/ContentNode/Card.vue
```diff
@@ -0,0 +1,55 @@
...
+  --card-color-head: var(--color-fill-tertiary);
+
+  background: var(--card-color-content);
+  border: var(--card-border-width) var(--card-border-style) var(--card-color-border);
+  border-radius: var(--card-border-radius);
+
+  @include space-out-between-siblings(var(--spacing-stacked-margin-xlarge));
+
+  & > * {
+    padding: 20px;
```
marinaaisa, line 46: For better control over spacing, I’d use Flex’s gap property in the `.card` element instead of `padding`. This avoids the potential for unexpected double spacing between elements.

Also, `$article-stacked-margin-small` seems to be used in rows and columns as the default spacing property.

```scss
.card {
  display: flex;
  flex-direction: column;
  gap: $article-stacked-margin-small;
}
```

src/components/ContentNode.vue
mportiz08: You're right about the recursive issue applying slot to deeply nested children when it shouldn't, although it surprisingly doesn't seem to cause any issues in practice.

In that case, I would prefer to just leave it as-is for now since your most recent suggestion is basically the same as the original code but just extracted to a function, and we're only using it in one place for now.

src/components/ContentNode/Card.vue
mportiz08: I'm not sure I entirely follow. Although I could rewrite this spacing using flex box and a gap between the header/body sections, I think it makes logical sense the way I've done it here:

* card is split into 1-2 boxes (2 if there is a header section)
  + each box gets 20px padding for its content

And just to clarify, the `> *` only applies to the 2 distinct boxes (head/content)—it doesn't affect any children within those boxes as well. There would need to be padding even if using a flex box container since we want the visual spacing to include the background colors.

(I could replace the hardcoded `20px` to the `$article-stacked-margin-small` variable though, although it's a little odd to read since it's not really used purely for stacking items here.)

src/components/ContentNode.vue
marinaaisa: leave it as-is for now is a good option too 👍

src/components/ContentNode/Card.vue
marinaaisa: > And just to clarify, the > * only applies to the 2 distinct boxes (head/content)—it doesn't affect any children within those boxes as well. There would need to be padding even if using a flex box container since we want the visual spacing to include the background colors.

Thanks for the clarification. I think it makes sense now if we want to include the background colours of the header to follow your original approach.

Also, $article-stacked-margin-small is a margin, not a padding, as you intend to add, so I’m happy to stick with the original approach.

marinaaisa (approved):
I tested it and it looks great. Thank you @mportiz08 ! 👏

As a minor detail: the script for testing with the `docc` branch https://github.com/swiftlang/swift-docc/pull/1508 is:

```bash
DOCC_HTML_DIR=/path/to/swift-docc-render/dist swift run docc preview --enable-card-directive /path/to/Example.docc
```
instead of:
```bash
DOCC_HTML_DIR=/path/to/swift-docc-render/dist swift run --enable-card-directive preview /path/to/Example.docc
```

mportiz08:
> I tested it and it looks great. Thank you @mportiz08 ! 👏
> 
> As a minor detail: the script for testing with the `docc` branch [swiftlang/swift-docc#1508](https://github.com/swiftlang/swift-docc/pull/1508) is:

I fixed this and also updated it with the updated flag name: `--enable-experimental-card-directive`

mportiz08:
@swift-ci test

PR #1005, opened 2026-04-17
[AX] Announce code block changes to screen readers in tutorial steps

### LINKED ISSUE #949: [AX] Code blocks are hard to reach and follow while reading a tutorial
### Description

Issues found:
- We don’t notify the user when the code block is going to be updated and when has been updated.
- We don’t provide an alternative and manual way to update the code block.
- The code block is too far away in the HTML tree of each step description, so the user cannot find it.

### Checklist

- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc-render/issues).

### Expected Behavior

- Users must be notified if the page will be updated as a result of interaction
- Users must be informed when a content change has occurred
- Users need an alternative method to manually trigger content updates

### Actual behavior

- We don’t notify the user when the code block is going to be updated and when has been updated.
- We don’t provide an alternative and manual way to update the code block.
- The code block is too far away in the HTML tree of each step description, so the user cannot find it.

### Steps To Reproduce

1. Turn on "VoiceOver"
2. Go to a tutorials page
3. Navigate through the steps
4. Assert that updates in the code blocks aren't being notified 
5. Assert that code blocks are only accessible after tabbing all the steps

_No response_

akinshaywai:
## Summary

Closes #949

When a tutorial step with a code block scrolls into view, screen readers were not notified of the change. This made it difficult or impossible for assistive technology users to follow tutorial code along as they scrolled through steps.

### Changes

- **`src/components/Tutorial/SectionSteps.vue`**: Adds a visually-hidden `aria-live="polite" aria-atomic="true"` region. When `onFocus(key)` is called for a step that has a code block, the region is updated with an announcement string ("Code updated for step N of M"), which is immediately read aloud by screen readers.
- **`src/lang/locales/en-US.json`** (and ja-JP, ko-KR, zh-CN): Adds the `tutorials.steps.code-updated` i18n key used by the announcement.
- **`tests/unit/components/Tutorial/SectionSteps.spec.js`**: Adds two tests — one verifying the announcement fires for code steps, one verifying it does not fire for media-only steps.

## Test plan

- [ ] Run `npm run test:unit` — all 20 `SectionSteps` tests pass
- [ ] Navigate a tutorial page using VoiceOver/NVDA — confirm "Code updated for step N of M" is announced when scrolling to a step with a code block
- [ ] Confirm media-only steps produce no announcement

### DIFF
--- a/src/components/Tutorial/SectionSteps.vue
+++ b/src/components/Tutorial/SectionSteps.vue
@@ -10,6 +10,11 @@
 
 <template>
   <div class="steps">
+    <div
+      aria-live="polite"
+      aria-atomic="true"
+      class="visuallyhidden"
+    >{{ stepAnnouncement }}</div>
     <div class="content-container">
       <component
         v-for="(node, index) in contentNodes"
@@ -107,6 +112,7 @@ export default {
         runtimePreview,
       },
       activeStep: firstStepIndex,
+      stepAnnouncement: '',
     };
   },
   computed: {
@@ -196,6 +202,17 @@ export default {
         media,
         runtimePreview,
       };
+
+      if (code) {
+        const stepNode = this.stepNodes.find(({ props }) => props.index === key);
+        if (stepNode) {
+          const { stepNumber } = stepNode.props;
+          this.stepAnnouncement = this.$t('tutorials.steps.code-updated', {
+            number: stepNumber,
+            total: this.numberOfSteps,
+          });
+        }
+      }
     },
     onRuntimePreviewToggle(value) {
       this.$emit('runtime-preview-toggle', value);
--- a/src/components/Tutorial/SectionSteps.vue
+++ b/src/components/Tutorial/SectionSteps.vue
@@ -157,6 +157,9 @@ export default {
       return this.tutorialState.breakpoint === BreakpointName.small;
     },
     stepNodes: ({ contentNodes, isStepNode }) => contentNodes.filter(isStepNode),
+    activeStepNode({ stepNodes, activeStep }) {
+      return stepNodes.find(({ props }) => props.index === activeStep);
+    },
     intersectionRootMargin: () => IntersectionMargins,
   },
   async mounted() {
@@ -203,15 +206,11 @@ export default {
         runtimePreview,
       };
 
-      if (code) {
-        const stepNode = this.stepNodes.find(({ props }) => props.index === key);
-        if (stepNode) {
-          const { stepNumber } = stepNode.props;
-          this.stepAnnouncement = this.$t('tutorials.steps.code-updated', {
-            number: stepNumber,
-            total: this.numberOfSteps,
-          });
-        }
+      if (code && this.activeStepNode) {
+        this.stepAnnouncement = this.$t('tutorials.code-updated', {
+          number: this.activeStepNode.props.stepNumber,
+          total: this.numberOfSteps,
+        });
       }
     },
     onRuntimePreviewToggle(value) {
--- a/tests/unit/components/Tutorial/SectionSteps.spec.js
+++ b/tests/unit/components/Tutorial/SectionSteps.spec.js
@@ -336,10 +336,8 @@ describe('SectionSteps', () => {
       wrapper.vm.onIntersect({ target: target(2), isIntersecting: true });
       await wrapper.vm.$nextTick();
 
-      const liveRegion = wrapper.find('[aria-live="polite"]');
-      expect(liveRegion.exists()).toBe(true);
       // $t mock returns 'key number total'
-      expect(wrapper.vm.stepAnnouncement).toBe('tutorials.steps.code-updated 2 5');
+      expect(wrapper.find('[aria-live="polite"]').text()).toBe('tutorials.code-updated 2 5');
     });
 
     it('does not announce when a step with only media becomes active', async () => {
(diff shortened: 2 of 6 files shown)


src/components/Tutorial/SectionSteps.vue
```diff
@@ -196,6 +202,17 @@ export default {
         media,
         runtimePreview,
       };
+
+      if (code) {
+        const stepNode = this.stepNodes.find(({ props }) => props.index === key);
+        if (stepNode) {
+          const { stepNumber } = stepNode.props;
+          this.stepAnnouncement = this.$t('tutorials.steps.code-updated', {
```
marinaaisa: Very minor comment but I don't think we need to create a subcategory `steps` if we only have a single string in there.

```suggestion
          this.stepAnnouncement = this.$t('tutorials.code-updated', {
```

in case we update this, we will also need to update the JSON files with the translated content

src/components/Tutorial/SectionSteps.vue
```diff
@@ -10,6 +10,11 @@
 
 <template>
   <div class="steps">
+    <div
+      aria-live="polite"
+      aria-atomic="true"
```
marinaaisa, line 15: Nice one!

src/components/Tutorial/SectionSteps.vue
```diff
@@ -196,6 +202,17 @@ export default {
         media,
         runtimePreview,
       };
+
+      if (code) {
+        const stepNode = this.stepNodes.find(({ props }) => props.index === key);
```
marinaaisa: I see how `stepNode` or `activeStepNode` could be valuable in the future.

I propose to create its own computed property. Something like:

```diff
diff --git a/src/components/Tutorial/SectionSteps.vue b/src/components/Tutorial/SectionSteps.vue
index bd58d131..1d00b3ee 100644
--- a/src/components/Tutorial/SectionSteps.vue
+++ b/src/components/Tutorial/SectionSteps.vue
@@ -157,6 +157,9 @@ export default {
       return this.tutorialState.breakpoint === BreakpointName.small;
     },
     stepNodes: ({ contentNodes, isStepNode }) => contentNodes.filter(isStepNode),
+    activeStepNode({ stepNodes, activeStep }) {
+      return stepNodes.find(({ props }) => props.index === activeStep);
+    },
     intersectionRootMargin: () => IntersectionMargins,
   },
   async mounted() {
@@ -203,15 +206,11 @@ export default {
         runtimePreview,
       };

-      if (code) {
-        const stepNode = this.stepNodes.find(({ props }) => props.index === key);
-        if (stepNode) {
-          const { stepNumber } = stepNode.props;
-          this.stepAnnouncement = this.$t('tutorials.steps.code-updated', {
-            number: stepNumber,
-            total: this.numberOfSteps,
-          });
-        }
+      if (code && this.activeStepNode) {
+        this.stepAnnouncement = this.$t('tutorials.steps.code-updated', {
+          number: this.activeStepNode.props.stepNumber,
+          total: this.numberOfSteps,
+        });
       }
     },
     onRuntimePreviewToggle(value) {
```

so we can reuse `activeStepNode` later on. What do you think?

tests/unit/components/Tutorial/SectionSteps.spec.js
```diff
@@ -330,6 +330,27 @@ describe('SectionSteps', () => {
...
+  describe('aria-live announcements', () => {
+    it('announces code block updates when a step with code becomes active', async () => {
+      // step at index 2 has code (exampleStepWithCode, stepNumber 2 of 5)
+      wrapper.vm.onIntersect({ target: target(2), isIntersecting: true });
+      await wrapper.vm.$nextTick();
+
+      const liveRegion = wrapper.find('[aria-live="polite"]');
+      expect(liveRegion.exists()).toBe(true);
+      // $t mock returns 'key number total'
+      expect(wrapper.vm.stepAnnouncement).toBe('tutorials.steps.code-updated 2 5');
```
marinaaisa:
```suggestion
      expect(liveRegion.text()).toBe('tutorials.steps.code-updated 2 5');
```
Instead of checking if the data in the component exists (`wrapper.vm.stepAnnouncement`), it’s better to check if the data has been rendered as expected. This approach aligns more closely with the user behaviour we want to achieve.

tests/unit/components/Tutorial/SectionSteps.spec.js
```diff
@@ -330,6 +330,27 @@ describe('SectionSteps', () => {
...
+    });
+
+    it('does not announce when a step with only media becomes active', async () => {
+      // step at index 1 has media only (exampleStepWithMedia)
+      wrapper.vm.onIntersect({ target: target(1), isIntersecting: true });
+      await wrapper.vm.$nextTick();
+
+      expect(wrapper.vm.stepAnnouncement).toBe('');
```
marinaaisa:
```suggestion
      const liveRegion = wrapper.find('[aria-live="polite"]');
      expect(liveRegion.text()).toBe('');
```
same feedback as above

marinaaisa (commented):
Hi @akinshaywai!

Thank you very much for your contribution. This is a really good solution for the first issue described on https://github.com/swiftlang/swift-docc-render/issues/949.

I’ve left some minor comments to help you improve the existing code. Let me know if you have any questions!

Thanks again 😊

marinaaisa:
@swift-ci test

marinaaisa (approved):
Thanks for working on this PR and addressing the feedback @akinshaywai ! looks great!

franklinsch:
@marinaaisa is this ready to be merged? If so feel free to go ahead and merge 🙂

marinaaisa:
@swift-ci test

PR #1017, opened 2026-05-27
Add below-abstract slot to DocumentationTopic for content injection

marinaaisa:
Bug/issue #176906827, if applicable: 

## Summary

Without a slot between the Abstract and the elements that follow it, consumers of DocumentationTopic have no way to inject custom content immediately after the page description. The only option would be to override or fork the entire component, which is fragile and breaks on upstream updates.

A new named slot `below-abstract` [1] is added directly after the `Abstract` component and before the `v-if="sampleCodeDownload"` div. This placement means content projected into the slot appears below the page description but above the download button, availability badges, and declarations, keeping it logically grouped within the introductory part of the hero without interfering with any of the surrounding conditionally rendered elements.

A test is added to verify that content projected into the slot is rendered inside the DocumentationHero component at the expected location.

[1] https://vuejs.org/guide/components/slots#named-slots

## Dependencies

NA

## Testing

Steps:
1. Extend the DocumentationTopic.vue component adding a slot with name `below-abstract`
2. Assert that it renders as expected

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/DocumentationTopic.vue
+++ b/src/components/DocumentationTopic.vue
@@ -61,6 +61,7 @@
           :class="{ 'minimized-abstract': enableMinimized }"
           :content="abstract"
         />
+        <slot name="below-abstract" />
         <div v-if="sampleCodeDownload">
           <DownloadButton
             class="sample-download"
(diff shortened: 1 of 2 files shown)


marinaaisa:
@swift-ci test

mportiz08 (approved):

marinaaisa:
@swift-ci test

PR #1019, opened 2026-06-01
Make Aside label customisable via a named slot

marinaaisa:
Bug/issue #176906827, if applicable: 

## Summary

The label rendered inside the Aside component was hardcoded as a plain `<p class="label">` element with no extension point. Any consuming component that needed a different label presentation, or no label at all, had no way to achieve that without adding a new special-cased prop or a new kind value just to toggle the element off.

Vue's named slots with fallback content [1] are the best solution here: the slot renders its default content when the parent provides nothing, preserving all existing behaviour exactly as before, while still giving any consumer the freedom to supply its own label markup or suppress the label entirely by providing a non-rendering element (e.g. `<span />`) in the `#label` slot.

The `<p class="label">` is now wrapped in `<slot name="label">…</slot>` so that it continues to act as the default for every current usage of `<Aside>` (ContentNode, DocumentationTopic, etc.), and a new test documents and verifies the override path.

## Dependencies

NA

## Testing

Steps:
1. Assert that the Aside component renders as expected.
2. Create a new component on top of this Aside component with a `label` slot and assert that it overwrites the slot.

## Checklist

- [x] Added tests

### DIFF
--- a/src/components/ContentNode/Aside.vue
+++ b/src/components/ContentNode/Aside.vue
@@ -10,7 +10,9 @@
 
 <template>
   <aside :class="kind" :aria-label="kind">
-    <p class="label">{{ name || $t(label) }}</p>
+    <slot name="label">
+      <p class="label">{{ name || $t(label) }}</p>
+    </slot>
     <slot />
   </aside>
 </template>
--- a/tests/unit/components/ContentNode/Aside.spec.js
+++ b/tests/unit/components/ContentNode/Aside.spec.js
@@ -59,6 +59,20 @@ describe('Aside', () => {
     expect(label.text()).toBe('Custom Name');
   });
 
+  it('renders a default label slot that can be overridden with custom content', () => {
+    const wrapper = shallowMount(Aside, {
+      propsData: {
+        kind: 'note',
+      },
+      slots: {
+        label: '<span class="custom-label">My Label</span>',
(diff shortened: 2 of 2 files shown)


marinaaisa:
@swift-ci test

src/components/ContentNode/Aside.vue
```diff
@@ -10,7 +10,7 @@
 
 <template>
   <aside :class="kind" :aria-label="kind">
-    <p class="label">{{ name || $t(label) }}</p>
+    <p v-if="!isNoLabel" class="label">{{ name || $t(label) }}</p>
```
mportiz08: It feels a little odd to me that we are creating a new special `kind` value but not actually using it anywhere since there is no Render JSON change emitting this special new kind.

Could we instead just make this label a named slot that renders this existing `p.label` as the default? Then other components using `Aside` can still overwrite this to omit it entirely or provide their own label UI implementation.

marinaaisa: This is great feedback, thanks a lot! It does make sense to simply add a label slot so that other components can overwrite the label when building on top of `Aside`. I changed the implementation 👍

mportiz08 (approved):

marinaaisa:
@swift-ci test

PR #1023, opened 2026-06-30
style: enable smooth scrolling for anchor links

GerardSerra18:
Bug/issue #, if applicable: #985

## Summary

This PR addresses the abrupt page jump that occurs when navigating via anchor links across documentation pages, such as clicking on specific section headers, table of contents items, or direct sub-page indexing links.

### Expected User Experience

Instead of an instantaneous and jarring snap down the viewport, the page now smoothly glides to the target element.

This provides better visual continuity and a more polished user experience, aligned with modern Apple design and web standards.

### Implementation Overview

Added `scroll-behavior: smooth` to the global `html` element within:

```scss
src/styles/base/_reset.scss
```

To ensure accessibility compliance, this animation is explicitly wrapped inside a `@media (prefers-reduced-motion: no-preference)` media query.

This guarantees that users with system-level reduced motion preferences enabled will keep the default immediate jump behavior instead of receiving an animated scroll.

## Dependencies

None.

## Testing

This presentational change has been verified by compiling the production asset pipeline and validating the compiled stylesheet behavior on a localized static production server.

### Steps

1. Run the production build script to ensure the SCSS asset pipeline compiles cleanly without syntax regressions:

   ```bash
   npm run build
   ```

2. Serve the generated static artifacts locally to bypass environment-specific Node/proxy constraints:

   ```bash
   npx serve dist
   ```

3. Open the local server URL provided in the terminal.

   Typically:

   ```text
   http://localhost:3000
   ```

4. Since the application mounts on the root directory without a default documentation bundle loaded, open the browser Developer Tools.

   You can do this with:

   ```text
   F12 / Inspect
   ```

5. Paste the following snippet into the Console to simulate an anchor-linked document structure:

   ```javascript
   document.body.innerHTML = `
     <div style="padding: 40px; font-family: sans-serif; min-height: 3000px; background: white;">
       <a href="#test-scroll" style="position: fixed; top: 20px; left: 20px; background: #0066cc; color: white; padding: 12px 24px; border-radius: 8px; font-weight: bold; text-decoration: none; box-shadow: 0 4px 12px rgba(0,0,0,0.15);">
         👉 Click to Verify Smooth Scroll
       </a>

       <div style="height: 1500px; margin: 40px 0; background: #f5f5f7; border: 2px dashed #ccc; display: flex; align-items: center; justify-content: center; color: #86868b;">
         (Simulated Documentation Content Viewport)
       </div>

       <h2 id="test-scroll" style="font-size: 32px; color: #1d1d1f; margin-top: 50px; padding-top: 20px; border-top: 2px solid #eee;">
         📍 Target Destination Reached Successfully
       </h2>
     </div>
   `;
   ```

6. Click the rendered blue action button:

   ```text
   👉 Click to Verify Smooth Scroll
   ```

7. Verify that the viewport smoothly transitions down to the target header element.

### Optional Accessibility Verification

1. Enable **Reduce Motion** in macOS System Settings.

2. Click the rendered blue action button again.

3. Verify that the behavior cleanly falls back to an instantaneous viewport jump.

## Checklist

- [x] Added tests
- **Reason:** Not applicable. The implementation utilizes native web standards and does not alter configuration files, public configuration flags, or custom design tokens.

### DIFF
--- a/src/styles/base/_reset.scss
+++ b/src/styles/base/_reset.scss
@@ -18,6 +18,10 @@ html {
   // enough that it does not actually exceed the breakpoint height.
   background-color: var(--colors-text-background, var(--color-text-background));
   height: 100%;
+  // Enable smooth scrolling for anchor links unless reduced motion is preferred.
+  @media (prefers-reduced-motion: no-preference) {
+    scroll-behavior: smooth;
+  }
 }
 
 //============================================================
--- a/src/styles/base/_reset.scss
+++ b/src/styles/base/_reset.scss
@@ -18,10 +18,6 @@ html {
   // enough that it does not actually exceed the breakpoint height.
   background-color: var(--colors-text-background, var(--color-text-background));
   height: 100%;
-  // Enable smooth scrolling for anchor links unless reduced motion is preferred.
-  @media (prefers-reduced-motion: no-preference) {
-    scroll-behavior: smooth;
-  }
 }
 
 //============================================================
--- a/src/utils/router-utils.js
+++ b/src/utils/router-utils.js
@@ -63,7 +63,17 @@ export async function scrollBehavior(to, from, savedPosition) {
     const offset = baseNavOffset + apiChangesNavHeight + getExtraScrollOffset(to);
 
     const y = process.env.VUE_APP_TARGET === 'ide' ? 0 : offset;
-    return { selector: cssEscapeTopicIdHash(hash), offset: { x: 0, y } };
+    const position = {
+      selector: cssEscapeTopicIdHash(hash),
+      offset: { x: 0, y },
+    };
+    if (typeof window !== 'undefined' && window.matchMedia && window.matchMedia('(prefers-reduced-motion: no-preference)').matches) {
+      return {
+        ...position,
+        behavior: 'smooth',
+      };
+    }
+    return position;
   }
   if (areEquivalentLocations(to, from)) {
     // Do not change the scroll position if the location hasn't changed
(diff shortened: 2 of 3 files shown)


mportiz08:
Hi @GerardSerra18. The contents of this PR look very similar to that of #986. Can you test these changes in a similar manner to see if it has the same issue mentioned [here][1]? I expect that it does, and we would need further iteration on this enhancement before we can merge it to avoid degrading the overall ux.

[1]: https://github.com/swiftlang/swift-docc-render/pull/986#pullrequestreview-4385426772

GerardSerra18:
Thanks for that point, @mportiz08.

I updated the implementation to avoid applying `scroll-behavior: smooth` globally on the `html` element. After looking at the existing scroll flow, I agree that changing the global browser/programmatic scroll behavior could interfere with the current Vue Router anchor navigation and offset handling.

The new approach scopes smooth scrolling to the existing router anchor scroll path instead:

* The current selector and offset logic are preserved.
* Smooth scrolling is only added when handling anchor navigation through `scrollBehavior`.
* Users with reduced motion preferences keep the default instant scroll behavior.
* The global CSS rule has been removed to avoid affecting unrelated scroll operations.

I also updated the router scroll behavior unit tests to cover both the smooth-scroll path and the reduced-motion fallback.

Validation completed locally:

```text
Test Suites: 271 passed, 271 total
Tests:       2301 passed, 2301 total
Snapshots:   31 passed, 31 total
```

I also tried to reproduce the issue using `npm run docs`, but local docs generation currently fails before the server starts due to `bin/generate-symbol-graph` throwing a `tree-sitter` `Invalid language object` error. I therefore focused this iteration on reducing the scope of the change and covering the behavior through the existing router scroll tests.

PR #1036, opened 2026-09-17
Fix locale switch to a translation that has no data

marinaaisa:
Bug/issue #187696128, if applicable: 

## Summary

A page can advertise a translation whose render JSON was never produced. Selecting that language fetched the localized JSON, got a 404, and fell back to the same route without the locale param. Since the router only registers locale prefixed route records, that fallback resolves to the location the reader is already on, which Vue Router rejects as duplicated, so the page was left without data while the picker had already switched the interface locale.

This change passes the router to the route data helper, which resolves the fallback location and aborts with `next(false)` when it matches the current route, keeping the reader on the intact page. The documentation view opts in and ignores the empty result the helper returns once it has resolved the navigation itself, and the locale selector applies the chosen locale only when the navigation completes.
Switching to a translation that exists, and falling back from another page or locale, behave as before.

A text is being rendered in the UI saying "Translation for “{language}” is not available yet".
The message is a status live region, so assistive technology announces it without moving focus.

## Dependencies

NA

## Testing

Steps:
1. Use a doccarchive that contains `availableLocales` but doesn't contain the RenderJSON for those locales.

## Checklist

- [x] Added tests


marinaaisa:
Closing this PR in favor of https://github.com/swiftlang/swift-docc-render/pull/1037

